PR #5, opened 2024-11-13
FTN-73: Add a database container to the compose.yaml file

gareth-allan-bjss:
This PR adds a database container and a "database setup" container to the `compose.yaml` file used to run the application locally, allowing us to start up the whole application and make a request from the frontend through the API to the DB on a local machine.

Changes:
- Added a `db` service to the `compose.yaml` file. This uses an MS SQL Server image to run SQL Server, which is what the Azure SQL Database a real deployment will be using is based on.
- Added a `db-setup` service to the `compose.yaml` file, and the necessary Docker file and script (to `database/fingertips-db`)to build it. This container builds the .dacpac file and then applies it to the DB using the microsoft.sqlpackage package tool. The intent is that this could be used to apply the .dacpac to the DB in a real deployment too
- Updated the API's DB connection code to allow a `TRUST_CERT` environment variable to be set in order to allow us to disable validation of the SQL server certificate. This is required because the SQL Server container uses a self-signed certificate, which does not appear to be straightforward to disable. If `TRUST_CERT` is set, the API will log a message at warning level. The same pattern is used in the DB setup container when publishing the .dacpac file.


database/fingertips-db/PostDeployment/PopulateWeatherForecast.sql
```diff
@@ -1,5 +1,5 @@
 -- This file contains SQL statements that will be executed after the build script.
-
+TRUNCATE TABLE WeatherForecast;
```
gareth-allan-bjss, line 2: I added this as otherwise we were ending up with multiple sets of entries in the `WeatherForecast` table when running the .dacpac file against a DB that had already been populated.

compose.yaml
```diff
@@ -6,14 +6,55 @@ services:
...
       - 3000:3000
 
   api:
     profiles:
       - all
       - api
+
+    depends_on:
+      db-setup:
+        condition: service_completed_successfully
```
gareth-allan-bjss, line 20: This condition means that the API doesn't start until the `db-setup` container has run successfully. It means that the API doesn't start until the DB is ready.

database/fingertips-db/PostDeployment/PopulateWeatherForecast.sql
steves-bjss: OK with this for now but in the longer term, for reference data we should probably put a guard in to figure out if we need to execute it or not.

steves-bjss (approved):

database/fingertips-db/PostDeployment/PopulateWeatherForecast.sql
gareth-allan-bjss: Yes, definitely!

PR #7, opened 2024-11-15
Update Frontend Packages

gareth-allan-bjss:
Make some dependency updates to the frontend app.

Changes:
- Update to Next 15.0.3
- Update to Node 22 (now in Active LTS status, as of 2024-10-29: https://nodejs.org/en/about/previous-releases)
- Update other NPM dependencies


.github/workflows/ci.yml
```diff
@@ -50,7 +50,7 @@ jobs:
     strategy:
       matrix:
         # See supported Node.js release schedule at https://nodejs.org/en/about/releases/
-        node-version: ["20.x", "22.x"]
+        node-version: ["22.x"]
```
gareth-allan-bjss, line 53: I don't think there's much point keeping the build against Node 20 any more, but I think keeping the matrix will allow us to add Node 24 when it becomes available.

steves-bjss (approved):

PR #10, opened 2024-11-28
FTN-14 Tech Setup

steves-bjss:


.github/workflows/ci.yml
```diff
@@ -91,6 +91,42 @@ jobs:
...
+                run: dotnet --version
+
+              - name: Restore Packages
+                run: dotnet restore
+
+              - name: Build API
+                run: dotnet build
+
+              - name: Run Unit Tests
+                run: dotnet test
```
gareth-allan-bjss, line 128: This indentation looks off. I don't think the contents of the `api-build` job need to be indented so far, do they?

.github/workflows/ci.yml
```diff
@@ -91,6 +91,42 @@ jobs:
...
+            defaults:
+              run:
+                working-directory: ${{ env.api-directory }}
+
+            steps:
+              - name: Checkout source
+                uses: actions/checkout@v4
+
+              - name: Setup dotnet ${{ matrix.dotnet-version }}
+                uses: actions/setup-dotnet@v3
```
gareth-allan-bjss, line 114: Looks like there's a v4 of this action. Any reason not to use that?

.github/workflows/ci.yml
```diff
@@ -91,6 +91,42 @@ jobs:
...
+                with:
+                  dotnet-version: ${{ matrix.dotnet-version }}
+
+              - name: Display dotnet version
+                run: dotnet --version
```
gareth-allan-bjss, line 119: Looks like the "Setup dotnet" step already prints the dotnet version being used, so is this redundant?
Screenshot 2024-12-05 at 16 39 21

api/DHSC.FingertipsNext.Modules.Search.ModuleInterfaces/ISearchController.cs
gareth-allan-bjss, line 1: Looks like this has been added without removing the `api/DHSC.FingertipsNext.Modules.Search.Interfaces/ISearchController.cs` file, so you've now got two copies of the `ISearchController.cs`.

gareth-allan-bjss (commented):
A few comments, but nothing major.

.github/workflows/ci.yml
steves-bjss: No they don't - reformatted as suggested.

.github/workflows/ci.yml
steves-bjss: No - updated to V4.

.github/workflows/ci.yml
steves-bjss: Removed.

api/DHSC.FingertipsNext.Modules.Search.ModuleInterfaces/ISearchController.cs
steves-bjss: Removed

gareth-allan-bjss (approved):

PR #15, opened 2024-12-04
Fix docker compose down command in README

adawari-thomas-bjss:
<h1>Description</h1>
This PR updates the docker compose down command in the README file, as the `profile` argument was placed in the wrong position, causing the command to fail. 

<h1>Changes</h1>

- Updated the `docker compose down` command and placed the `profile` argument in the right position.


gareth-allan-bjss (approved):
Thanks for fixing this, that was my mistake in modifying a command without testing it!

PR #20, opened 2024-12-10
FTN-72: search results page

adawari-thomas-bjss:
# Description

Jira ticket: [FTN-72](https://example.com)

## Changes

- Created components to display search results
- Created dummy results to display on the results page
- Created unit and snapshot tests for the results page
- Updated the Playwright tests to ensure results are displayed

## Validation

image


frontend/fingertips-frontend/components/pages/search/results/index.module.css
```diff
@@ -0,0 +1,18 @@
...
+}
+
+.result li:first-child {
+  border-top: 1px solid
+}
+
+.result li p {
+  font-size: 15px;
+  line-height: 10%;
+}
```
alibombaye, line 18: Can we try and not use any css styles. The govuk-react component library for Paragraph contains modifiers that we can use to set the size. 

If there is any spacing we need to do. Try and used the styled-components way. An example is on 
`frontend/fingertips-frontend/components/forms/SearchForm/index.tsx`
```
const StyledInputField = styled(InputField)(
  spacing.withWhiteSpace({ marginBottom: 6 })
);
```

There is a component for the line known as section break https://govuk-react.github.io/govuk-react/?path=/docs/section-break--docs

Also an unordered list which we can use https://govuk-react.github.io/govuk-react/?path=/docs/unordered-list--docs so we don't need to style it. 

Hopefully using these components should make the css changes we need to make non-existent or minimal.

Happy to discuss

frontend/fingertips-frontend/components/pages/search/results/index.tsx
```diff
@@ -11,6 +23,14 @@ export function SearchResults({ indicator }: Readonly<{ indicator: string }>) {
       />
       <H1>Search results</H1>
       <Paragraph>{`You searched for indicator "**${indicator}**"`}</Paragraph>
+      {searchResults.length ? (
+        <UnorderedList className={classes['result']}>
```
alibombaye, line 27: can use prop `listStyleType="none"` here and don't need the css

frontend/fingertips-frontend/components/pages/search/results/index.tsx
```diff
@@ -11,6 +23,14 @@ export function SearchResults({ indicator }: Readonly<{ indicator: string }>) {
...
+          {searchResults.map((result) => (
+            <SearchResult key={result.id} result={result} />
+          ))}
+        </UnorderedList>
+      ) : null}
+      {!searchResults.length && <Paragraph>No results found</Paragraph>}
```
alibombaye, line 33: This can go where you have the `null`?

frontend/fingertips-frontend/lib/registry.test.tsx
```diff
@@ -32,7 +32,8 @@ describe('StyledComponentsRegistry', () => {
     expect(container.asFragment()).toMatchSnapshot();
   });
 
-  it('should not return inline styles when rendering the wrapped component server side', () => {
+  it.skip('should not return inline styles when rendering the wrapped component server side', () => {
```
alibombaye, line 35: Yeah this test is an issue. I'm happy for you to delete this test file. Discussed with @williamwillis11 and we think this test will be better placed as an e2e playwright test

frontend/fingertips-frontend/components/pages/search/results/index.module.css
adawari-thomas-bjss: updated

frontend/fingertips-frontend/components/pages/search/results/index.tsx
adawari-thomas-bjss: done

frontend/fingertips-frontend/components/pages/search/results/index.tsx
adawari-thomas-bjss: done

frontend/fingertips-frontend/lib/registry.test.tsx
adawari-thomas-bjss: deleted the test files

frontend/fingertips-frontend/playwright/page-objects/pages/resultsPage.ts
```diff
@@ -4,6 +4,7 @@ import { expect } from '../pageFactory';
 export default class ResultsPage extends BasePage {
   readonly resultsText = 'You searched for indicator';
   readonly backLink = 'search-results-back-link';
+  readonly searchResult = 'search-result';
 
   async checkSearchResults(searchTerm: string) {
```
gareth-allan-bjss, line 9: Can the `checkSearchResults` method now be removed, given you've added `checkResultsAreDisplayed`?

frontend/fingertips-frontend/components/pages/search/results/Results.test.tsx
```diff
@@ -0,0 +1,42 @@
...
+    render(<SearchResults indicator="" searchResults={[]} />);
+
+    expect(screen.queryByText(/no results found/i)).toBeInTheDocument();
+    expect(screen.queryByRole('list')).not.toBeInTheDocument();
+    expect(screen.queryAllByTestId('search-result')).toHaveLength(0);
+  });
+
+  test('snapshot test', () => {
+    const container = render(
+      <SearchResults indicator="" searchResults={MOCK_DATA} />
```
gareth-allan-bjss, line 37: Probably worth passing this an indicator value, as I'm not sure we would ever support not having an indicator specified.

frontend/fingertips-frontend/components/pages/search/results/index.tsx
```diff
@@ -1,8 +1,25 @@
...
-export function SearchResults({ indicator }: Readonly<{ indicator: string }>) {
+import { SearchResultInterface } from '@/app/search/results/search-result-data';
+
+type SearchResultsProps = {
+  indicator: string;
+  searchResults: SearchResultInterface[];
+};
+
+export default function SearchResults({
+  indicator,
```
gareth-allan-bjss, line 20: Should this handle the case where an empty indicator is specified?

frontend/fingertips-frontend/components/pages/search/results/Results.test.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { expect } from '@jest/globals';
+import { MOCK_DATA } from '@/app/search/results/search-result-data';
+import SearchResults from '.';
+
+describe('Search Results Suite', () => {
+  test('should render elements', async () => {
+    const indicator = 'test';
+    render(<SearchResults indicator={indicator} searchResults={[]} />);
+
+    expect(screen.getByRole('link')).toBeInTheDocument();
```
gareth-allan-bjss, line 11: Should we be using test IDs for all elements we look for in this page, rather than looking for text, etc.?

@williamwillis11, @alibombaye?

frontend/fingertips-frontend/components/pages/search/results/index.tsx
```diff
@@ -11,6 +28,16 @@ export function SearchResults({ indicator }: Readonly<{ indicator: string }>) {
       />
       <H1>Search results</H1>
       <Paragraph>{`You searched for indicator "**${indicator}**"`}</Paragraph>
+      {searchResults.length ? (
+        <UnorderedList listStyleType="none">
+          {searchResults.map((result) => (
+            <SearchResult key={result.id} result={result} />
+          ))}
+          <SectionBreak visible="true" />
```
gareth-allan-bjss, line 36: Looks like you have a `SectionBreak` in the `SearchResult` component itself too. Why do we need both? It seems odd to have one before in that component and one after the component here.

frontend/fingertips-frontend/components/molecules/Search/result/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+    padding: [
+      { size: 3, direction: 'top' },
+      { size: 0, direction: 'bottom' },
+      { size: 7, direction: 'left' },
+      { size: 7, direction: 'right' },
+    ],
+  })
+);
+
+export default function SearchResult({ result }: Readonly<SearchResultProps>) {
```
alibombaye, line 26: Sorry minor but can we remove the default please. A named export I find makes it easier for the ide to find the import and we will always the name of the import to be the same.

frontend/fingertips-frontend/components/molecules/Search/result/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+const StyledListItem = styled(ListItem)(
+  spacing.withWhiteSpace({
+    padding: [
+      { size: 3, direction: 'top' },
+      { size: 0, direction: 'bottom' },
+      { size: 7, direction: 'left' },
+      { size: 7, direction: 'right' },
+    ],
+  })
+);
```
alibombaye, line 24: Nice 👍

frontend/fingertips-frontend/playwright/tests/search.spec.ts
```diff
@@ -22,4 +22,29 @@ test('Search via indicator and assert results', async ({
...
+  // Assert
+  await resultsPage.checkURLIsCorrect(indicator);
+  await resultsPage.checkSearchResults(indicator);
+
+  // Act
+  await resultsPage.clickBackLink();
+
+  // Assert
+  await searchPage.checkURLIsCorrect(`?indicator=${indicator}`);
+  expect((await axeBuilder.analyze()).violations).toEqual([]);
```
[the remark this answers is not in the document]
williamwillis11: yes agreed - if we don't expect anything different to be displayed on the page then lets not check a11y again

frontend/fingertips-frontend/playwright/tests/search.spec.ts
```diff
@@ -22,4 +22,29 @@ test('Search via indicator and assert results', async ({
...
+test('Should return to search page upon results page back link click', async ({
+  searchPage,
+  resultsPage,
+  axeBuilder,
+}) => {
+  // Arrange
+  await searchPage.navigateToSearch();
+
+  // Act
+  await searchPage.typeIndicator(indicator);
```
williamwillis11, line 37: can we add an // Assert section to call

  await searchPage.checkURLIsCorrect();

or is this overkill?

frontend/fingertips-frontend/app/search/results/search-result-data.ts
```diff
@@ -0,0 +1,36 @@
+export interface SearchResultInterface {
```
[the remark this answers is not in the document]
adawari-thomas-bjss: renaming the interface causes a naming conflict within `SearchResult` component. This could be solved using an import alias, so I'm fine with whichever approach is chosen.

frontend/fingertips-frontend/playwright/tests/search.spec.ts
adawari-thomas-bjss: I didn't add this as it seems the first test already does that check upon initial navigation to the search page, but I'm happy to add it in.

frontend/fingertips-frontend/app/search/results/search-result-data.ts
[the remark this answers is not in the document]
gareth-allan-bjss: Ah. I think if there's already a component called `SearchResult`, then having this type named `SearchResultInterface` might be even more confusing!

How about, since it's the result of searching for an indicator `IndicatorSearchResult`? It's looking likely that we will be searching for things other than indicators in future, so it's probably worth being clear what kind of result it is.

frontend/fingertips-frontend/components/pages/search/results/index.tsx
```diff
@@ -1,16 +1,52 @@
...
 
-export function SearchResults({ indicator }: Readonly<{ indicator: string }>) {
+import { IndicatorSearchResult } from '@/app/search/results/search-result-data';
+
+type SearchResultsProps = {
+  indicator: string;
+  searchResults: IndicatorSearchResult[];
+};
+
+export default function SearchResults({
```
alibombaye, line 20: Can we remove the default here too please

adawari-thomas-bjss: done

alibombaye (approved):

frontend/fingertips-frontend/components/pages/search/results/index.tsx
gareth-allan-bjss: I can see you've not got one `SectionBreak` at the start of the list and then each result adds its own after its content, so I'm happy to resolve this.

gareth-allan-bjss (approved):

PR #21, opened 2024-12-11
Feature/ftn 93 spike playwright

williamwillis11:
# Description

Jira ticket: [FTN93](https://dh-ft.atlassian.net/browse/FTN-93)

This PR is to switch to Playwright (from Cypress) for e2e tests and add accessibility testing (to a WCAG 2.2 AA standard) while also adding Safari (Webkit) browser coverage in CI.

## Changes

- Removed Cypress and added Playwright
- Rewrote Cypress tests into Playwright tests using async + await + POM
- Added 2 libraries to support automated a11y testing to WCAG2.2 AA
- Implemented a11y checks in existing e2e test for each page landed on
- Split out current build frontend and e2e test job into 2 new self explanatorily named jobs that run concurrently: 
1. frontend-unit-test-and-static-analysis
2. e2e-tests
- Added uploading into CI artefacts of Playwright reports generated when a test fails to help with debugging failures
- Using the dockerfile from Microsoft for faster pipelines as this way we don't need to install the browsers we need manually. Currently saves around 30 seconds but will get larger once more browsers start getting covered for full GDS on main project. - https://hub.docker.com/r/microsoft/playwright
- Updated frontend Readme to reflect these changes

## Validation

Ran e2e tests locally on both Chrome and Safari, both headed and headless and got passes. Ran successful pipelines as you can see in Actions.

For testing out a11y i added a second button **without** a name to components/forms/SearchForm/index.tsx and this correctly triggered a lot of a11y violations locally and in CI.


frontend/fingertips-frontend/playwright/page-objects/basePage.ts
```diff
@@ -0,0 +1,33 @@
...
+
+    let pageNavigation;
+    //Retry logic for page navigation
+    while (retryAttempt < maxRetries) {
+      try {
+        pageNavigation = await this.page.goto(page);
+        break;
+      } catch (error) {
+        retryAttempt++;
+        //Throw error if max retries reached
```
alibombaye, line 24: Don't think we need these comments. Code seems to self explain this anyway.

frontend/fingertips-frontend/.gitignore
```diff
@@ -12,6 +12,7 @@
 
 # testing
 /coverage
+*.png
```
alibombaye, line 15: do we want to gitignore this?

frontend/fingertips-frontend/components/forms/SearchForm/__snapshots__/SearchForm.test.tsx.snap
```diff
@@ -2,6 +2,114 @@
...
+
+}
+
+@media only screen and (min-width: 641px) {
+
+}
+
+@media only screen and (min-width: 641px) {
+
+}
```
alibombaye, line 76: Not sure why we are getting these additional empty css classes added to our snapshot when all that was added was a H1 tag to the search form - Was something else changed?

frontend/fingertips-frontend/eslint.config.mjs
```diff
@@ -10,11 +10,10 @@ const compat = new FlatCompat({
   recommendedConfig: js.configs.recommended,
   allConfig: js.configs.all,
 });
-export default [
+export default js.configs = [
```
alibombaye, line 13: just out of interest why was has this been named and why was the rule removed?

williamwillis11: i was getting an eslint warning for unnamed default export 

the rule was removed as it was for a cypress file which has been removed - so no longer need the rule

frontend/fingertips-frontend/.gitignore
williamwillis11: ah meant to remove this on last PR but then closed it and forgot - will remove it ta

frontend/fingertips-frontend/playwright/page-objects/pageFactory.ts
```diff
@@ -0,0 +1,74 @@
...
+        ]);
+
+        await use(axeBuilder);
+
+        //Dedicated accessibility tests do not need a re-scan - this can't be triggered sooner in the fixture as the tests will hang
+        if (testInfo.project.testMatch.toString().includes('accessibility'))
+          return;
+
+        // //Skip the accessibility scan following each UI test if `CONFIG.test.accessibility.autoscanEnabled` is set to 'true'.
+        // if (CONFIG.test.accessibility.autoscanEnabled === false) return;
```
alibombaye: Do we want to keep this commented out code?

frontend/fingertips-frontend/components/forms/SearchForm/__snapshots__/SearchForm.test.tsx.snap
alibombaye: Yeah just adding the H1 tag caused all this 😢  - It's a shame but I think we just have to live with it for now and keep an eye on it. I think we need to use the snapshot tests very sparingly going forward.

alibombaye:
Ran the e2e tests locally - both headless and headed and everything passes. 

Also tested making the test fail by changing a data-testid and the tests fails as expected.

All good for me 👍

alibombaye (approved):

PR #22, opened 2024-12-11
FTN-73: hardcoded indicator data endpoint

MonkeyDate:
# Description

Jira ticket: [FTN-73](https://example.com)

Add a new module that provides a single REST endpoint for accessing mock indicator data. Agreed deviations from jira ticket include
- Hard coded test data does not have to match FTN-68.

Not complete
- Scalar documentation. I have not been able to get good documentation to appear using scalar.
- Error handling has not been customised from the project default.
- For test purposes only negative indicator IDs will generate an exception when fetched.

## Changes

- Add `indicators` module with a single working controller.
- Unit tests in separate project.
- HTTP scripts to exercise endpoint.

Test Coverage only missing on module registration of types, and the hard coded data provider
image

## Validation

Endpoint works as expected using a browser

image


steves-bjss (approved):
Approving to unblock work over the holidays.  This will be revisited in the new year, along with the SonarQube issues.

PR #24, opened 2024-12-12
FTN-63: search page updates

ataylor-bjss:
# Description

Jira ticket: FTN-63

Updated content of search page and added validation of user input (not blank).

## Changes

- Added additional text as per Figma design.
- Validate user input (not blank) before submitting request.
- Display error message (using standard govuk component) if input is blank.

## Validation

Screenshot 2024-12-12 at 14 56 37

### DIFF
--- a/frontend/fingertips-frontend/app/search/page.tsx
+++ b/frontend/fingertips-frontend/app/search/page.tsx
@@ -9,6 +9,11 @@ export default async function Page(
 ) {
   const searchParams = await props.searchParams;
   const indicator = searchParams?.indicator ?? '';
+  const initialState = {
+    indicator: indicator,
+    message: null,
+    errors: {}
+  };
 
-  return <SearchForm indicator={indicator} />;
+  return <SearchForm searchFormState={initialState} />;
 }
--- a/frontend/fingertips-frontend/components/forms/SearchForm/SearchForm.test.tsx
+++ b/frontend/fingertips-frontend/components/forms/SearchForm/SearchForm.test.tsx
@@ -2,6 +2,7 @@ import { expect } from '@jest/globals';
 import { render, screen } from '@testing-library/react';
 import { SearchForm } from '@/components/forms/SearchForm';
 import { SearchFormState } from './searchActions';
+import { registryWrapper } from '@/lib/testutils';
 
 jest.mock('react', () => {
   const originalModule = jest.requireActual('react');
@@ -22,14 +23,31 @@ jest.mock('react', () => {
   };
 });
 
-test('snapshot test - renders the form', () => {
-  const container = render(<SearchForm indicator="" />);
+const initialState: SearchFormState = {
+  indicator: '',
+  message: null,
+  errors: {}
+};
 
+test('snapshot test - renders the form', () => {
+  const container = render(registryWrapper(<SearchForm searchFormState={initialState} />));
+  
   expect(container.asFragment()).toMatchSnapshot();
 });
 
 test('should have an input field to input the indicatorId', () => {
-  render(<SearchForm indicator="" />);
+  render(registryWrapper(<SearchForm searchFormState={initialState} />));
 
   expect(screen.getByTestId('search-form-input-indicator')).toBeInTheDocument();
+});
+test('should display the error summary component when there is a validation error', () => {
+  const errorState : SearchFormState = {
+    indicator: '',
+    message: 'Error message',
+    errors: {}
+  };
+  
+  render(registryWrapper(<SearchForm searchFormState={errorState} />));
+
+  expect(screen.getByTestId('error-summary')).toBeInTheDocument();  
 });
--- a/frontend/fingertips-frontend/components/forms/SearchForm/index.tsx
+++ b/frontend/fingertips-frontend/components/forms/SearchForm/index.tsx
@@ -1,7 +1,7 @@
 'use client';
 
 import { searchIndicator, SearchFormState } from './searchActions';
-import { H1, Button, InputField } from 'govuk-react';
+import { H1, ErrorSummary, LeadParagraph, InsetText, Button, InputField } from 'govuk-react';
 import { spacing } from '@govuk-react/lib';
 import { useActionState } from 'react';
 import styled from 'styled-components';
@@ -10,34 +10,41 @@ const StyledInputField = styled(InputField)(
   spacing.withWhiteSpace({ marginBottom: 6 })
 );
 
-export const SearchForm = ({ indicator }: { indicator: string }) => {
-  const initialState: SearchFormState = {
-    indicator: '',
-    message: null,
-    errors: {},
-  };
-  const [_state, formAction] = useActionState(searchIndicator, initialState);
+export const SearchForm = ({ searchFormState }: { searchFormState: SearchFormState }) => {
+  const [state, formAction] = useActionState(searchIndicator, searchFormState);
 
   return (
-    <>
-      <H1>Search</H1>
+    <main>
       <form action={formAction}>
+        {state.message &&
+          <ErrorSummary
+            description="At least one of the following fields must be populated:"
+            errors={[
+              {
+                targetName: 'indicator',
+                text: 'Indicator field'
+              }
+            ]}
+            data-testid="error-summary"
+          />
+        }
+        <H1>Find public health data</H1>
+        <LeadParagraph>Find public health data by subjects, and health or local authority areas.</LeadParagraph>
+        <InsetText>Use both search options to help you find the most accurate data available.</InsetText>
         <StyledInputField
           input={{
             id: 'indicator',
             name: 'indicator',
-            defaultValue: indicator,
+            defaultValue: searchFormState.indicator,
           }}
-          hint={<>Type in here the IndicatorId.</>}
+          hint={<>For example diabetes, public health indicator, or indicator ID241</>}
           data-testid="search-form-input-indicator"
         >
-          Indicator id or keyword
+          Search by subject
         </StyledInputField>
 
-        <Button data-testid="search-form-button-submit" type="submit">
-          Search
-        </Button>
+        <Button type="submit" data-testid="search-form-button-submit">Search</Button>
       </form>
-    </>
+    </main>
   );
 };
--- a/frontend/fingertips-frontend/components/forms/SearchForm/__snapshots__/SearchForm.test.tsx.snap
+++ b/frontend/fingertips-frontend/components/forms/SearchForm/__snapshots__/SearchForm.test.tsx.snap
@@ -2,7 +2,170 @@
 
 exports[`snapshot test - renders the form 1`] = `
 <DocumentFragment>
-  .c0 {
+  .c8 {
+  font-family: "nta",Arial,sans-serif;
+  -webkit-font-smoothing: antialiased;
+  -moz-osx-font-smoothing: grayscale;
+  font-weight: 400;
+  font-size: 16px;
+  line-height: 19px;
+  box-sizing: border-box;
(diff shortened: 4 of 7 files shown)


frontend/fingertips-frontend/app/search/page.tsx
```diff
@@ -9,6 +9,11 @@ export default async function Page(
 ) {
   const searchParams = await props.searchParams;
   const indicator = searchParams?.indicator ?? '';
+  const initialState = {
```
Joe-Hendley, line 12: is this also a `SearchFormState` as used elsewhere?

frontend/fingertips-frontend/components/forms/SearchForm/SearchForm.test.tsx
```diff
@@ -22,14 +23,33 @@ jest.mock('react', () => {
...
   expect(container.asFragment()).toMatchSnapshot();
 });
 
 test('should have an input field to input the indicatorId', () => {
-  render(<SearchForm indicator="" />);
+  render(registryWrapper(<SearchForm searchFormState={initialState} />));
 
   expect(screen.getByTestId('search-form-input-indicator')).toBeInTheDocument();
 });
+test('should display the error summary component when there is a validation error', () => {
```
alibombaye, line 45: very minor but can you add a blank line inbetween the tests please. There is a plugin we can add that enforces this which would be nice to add https://www.npmjs.com/package/eslint-plugin-jest-formatting 🥇

frontend/fingertips-frontend/components/forms/SearchForm/SearchForm.test.tsx
```diff
@@ -22,14 +23,33 @@ jest.mock('react', () => {
...
+  const errorState: SearchFormState = {
+    indicator: '',
+    message: 'Error message',
+    errors: {},
+  };
+
+  render(registryWrapper(<SearchForm searchFormState={errorState} />));
+
+  expect(screen.getByTestId('search-form-error-summary')).toBeInTheDocument();
+});
```
alibombaye, line 55: Can you add another test please that checks the input field is pre-populated with the data from the initialState

frontend/fingertips-frontend/components/forms/SearchForm/index.tsx
```diff
@@ -10,34 +17,57 @@ const StyledInputField = styled(InputField)(
...
+  searchFormState,
+}: {
+  searchFormState: SearchFormState;
+}) => {
+  const [state, formAction] = useActionState(searchIndicator, searchFormState);
 
   return (
-    <>
-      <H1>Search</H1>
+    <main>
```
alibombaye, line 28: I don't think we need this to be `main` - The container component already does this. I think this was added at some point in some other PR so it could have been complaining about needing this before.

frontend/fingertips-frontend/components/forms/SearchForm/index.tsx
```diff
@@ -10,34 +17,57 @@ const StyledInputField = styled(InputField)(
...
           input={{
             id: 'indicator',
             name: 'indicator',
-            defaultValue: indicator,
+            defaultValue: searchFormState.indicator,
           }}
-          hint={<>Type in here the IndicatorId.</>}
+          hint={
+            <>
+              For example diabetes, public health indicator, or indicator ID241
```
gareth-allan-bjss, line 59: Did you mean to include the `241` here?
```suggestion
              For example diabetes, public health indicator, or indicator ID
```

frontend/fingertips-frontend/components/forms/SearchForm/index.tsx
```diff
@@ -10,34 +17,57 @@ const StyledInputField = styled(InputField)(
...
-      <H1>Search</H1>
+    <main>
       <form action={formAction}>
+        {state.message && (
+          <ErrorSummary
+            description="At least one of the following fields must be populated:"
+            errors={[
+              {
+                targetName: 'indicator',
+                text: 'Indicator field',
```
gareth-allan-bjss, line 36: My expectation (and I think this is backed up by [the Gov.UK frontend guidance](https://design-system.service.gov.uk/components/error-summary/#linking-from-the-error-summary-to-each-answer)) is that clicking the link in the error summary should scroll the relevant field (i.e. the indicator field) into view and bring it into focus. This isn't working for me.

frontend/fingertips-frontend/components/forms/SearchForm/index.tsx
```diff
@@ -10,34 +17,57 @@ const StyledInputField = styled(InputField)(
...
+}) => {
+  const [state, formAction] = useActionState(searchIndicator, searchFormState);
 
   return (
-    <>
-      <H1>Search</H1>
+    <main>
       <form action={formAction}>
+        {state.message && (
+          <ErrorSummary
```
gareth-allan-bjss, line 31: The [Gov.UK frontend guidance for the Error Summary component](https://design-system.service.gov.uk/components/error-summary/) says:
> When a user makes an error, you must show both an error summary and an [Error message component](https://design-system.service.gov.uk/components/error-message/) next to each answer that contains an error.

So I think we need to do that too.

gareth-allan-bjss (commented):
Some initial comments.

I think if the issues with the form validation that I've flagged prove to really tricky to fix we could always consider deferring them to a later PR or ticket. We know this is an initial version of the page.

frontend/fingertips-frontend/components/forms/SearchForm/index.tsx
ataylor-bjss: I don't think it says that it sets focus to the field, I think it should just scroll the offending field into view (if it isn't).

frontend/fingertips-frontend/components/forms/SearchForm/index.tsx
ataylor-bjss: Discussed this with Fawad, he logic for not doing this is that when the page has two fields leaving either of them blank is not an error, it's only a problem when both are blank. In this case you can't really say which of the fields is in error so I don't think it makes sense to highlight one or the other.

frontend/fingertips-frontend/components/forms/SearchForm/index.tsx
ataylor-bjss: Yes, the ID241 is as it appears in the latest Figma design.

frontend/fingertips-frontend/components/forms/SearchForm/index.tsx
ataylor-bjss: Yes, think this went in from another PR. I notice the adding of landmarks however has added one (footer) to a component that already contains one (the govuk footer), should I remove it?

frontend/fingertips-frontend/components/forms/SearchForm/index.tsx
gareth-allan-bjss: I think there's a newer version of the designs (I can see a "round 3") without this. Also, I don't think it makes sense. Even if it was intended as an example it's not formatted in a way that makes it clear.

ataylor-bjss: Round 3 looks quite a bit different. Should I be updating with the additional text?

frontend/fingertips-frontend/components/forms/SearchForm/index.tsx
gareth-allan-bjss: I agree it's not clear about focusing the field (it just says to "link to" the field), but the behaviour of the examples on the page I linked is to focus the linked field, as well as scrolling to it.

Regardless of that, scrolling to the field isn't working for me on the page as it is now. I don't get taken to the error summary when it is shown, and clicking the link to the indicator field doesn't appear to do anything.

frontend/fingertips-frontend/components/forms/SearchForm/SearchForm.test.tsx
```diff
@@ -22,14 +23,48 @@ jest.mock('react', () => {
...
+    indicator: 'test value',
+    message: '',
+    errors: {},
+  };
+  render(registryWrapper(<SearchForm searchFormState={indicatorState} />));
+
+  expect(screen.getByDisplayValue('test value')).toHaveAttribute(
+    'id',
+    'indicator'
+  );
```
gareth-allan-bjss, line 57: Not a massive issue, but this seems a bit backwards. You're trying to assert the indicator input field has the expected value, but you're searching for elements with the expected display value and then checking the ID is that of the indicator field. Couldn't you just get the indicator field by ID (or the `search-form-input-indicator` `data-testid`) and then assert against its value?

frontend/fingertips-frontend/components/forms/SearchForm/SearchForm.test.tsx
```diff
@@ -42,6 +42,18 @@ test('should have an input field to input the indicatorId', () => {
...
+
+test('should set the input field with indicator value from the form state', () => {
+  const indicatorState: SearchFormState = {
+    indicator: 'test value',
+    message: '',
+    errors: {},
+  };
+  render(registryWrapper(<SearchForm searchFormState={indicatorState} />));
+
+  expect(screen.getByDisplayValue('test value')).toHaveAttribute('id', 'indicator');
```
alibombaye: Can we swap this around a little bit. Something like this perhaps;

```
expect(screen.getByRole('textbox', { name: /indicator/i })).toHaveValue(
    'test value'
  );
```

frontend/fingertips-frontend/components/forms/SearchForm/index.tsx
gareth-allan-bjss: We probably need to be careful about spending time tracking the design updates, because we know they're not final and the wording at least will likely change again. I don't think we should worry about it now.

I just think the `241` looks strange and is likely to cause questions when people see it. If you feel really strongly about keeping it I don't want to hold things up, but it just looks like a typo to me.

frontend/fingertips-frontend/components/forms/SearchForm/index.tsx
alibombaye: ah yes try removing it. The accessibility test were complaining about the footer landmark being missing but you are right it already does contain a `footer` landmark as part of the govuk-footer

frontend/fingertips-frontend/components/forms/SearchForm/SearchForm.test.tsx
alibombaye: Yes I agree with this. 

Just something for us to be vary of. The data-testid `search-form-input-indicator` is against the full component. The input is within it so finding by data-testid and checking it's value doesn't quite work. Probably another reason to prefer using the `getByRole` instead.  
 
Screenshot 2024-12-13 at 12 46 25

alibombaye (approved):

frontend/fingertips-frontend/components/forms/SearchForm/index.tsx
```diff
@@ -36,6 +36,11 @@ export const SearchForm = ({
             },
           ]}
           data-testid="search-form-error-summary"
+          onHandleErrorClick={(targetName: string) => {
```
gareth-allan-bjss, line 39: I think given we've got our own logic here it would nice to have a test that validates clicking on the error link actually results in the indicator field being focused.

gareth-allan-bjss (approved):
This page looks really good now, thanks Alan.

One nice-to-have test I've suggested, but happy to discuss deferring this if we think it might not be straightforward.

gareth-allan-bjss (approved):
Thanks for adding that test, I think it'll be a nice example of how to do it as we go forward.

PR #25, opened 2024-12-13
Feature/ftn 68 view line chart

priya-thorpe-bjss:
# Description

Jira ticket: [FTN:68]

Render a line chart using the Highcharts library.

## Changes
- Render a line chart using the highcharts library.
- Render a table using the gov-react-table when JavaScript is disabled.

## Validation

Table rendered when JavaScript is enabled.
Screenshot 2024-12-13 at 16 11 44

Table rendered when JavaScript is disabled
Screenshot 2024-12-13 at 16 09 26


frontend/fingertips-frontend/playwright/page-objects/pageFactory.ts
```diff
@@ -2,11 +2,13 @@
 import AxeBuilder from '@axe-core/playwright';
 import SearchPage from './pages/searchPage';
 import ResultsPage from './pages/resultsPage';
+import ChartPage from '@/playwright/page-objects/pages/chartPage';
```
gareth-allan-bjss, line 5: The structure of this import is different to the other pages being imported:
```suggestion
import ChartPage from ./pages/chartPage';
```

frontend/fingertips-frontend/.env.development
```diff
@@ -1 +1 @@
-FINGERTIPS_API_URL="http://localhost:5144"
+FINGERTIPS_API_URL="http://localhost:5144/weatherforecast"
```
gareth-allan-bjss, line 1: I know I suggested doing this when you were seeing issues, but thinking about it, we should really be leaving this value alone and including the `/weatherforecast` path in the page(s) that want weather forecast data. Otherwise we're going to stop ourselves from being able to use this URL as the base for any other API queries.

frontend/fingertips-frontend/app/chart/page.tsx
```diff
@@ -0,0 +1,27 @@
+import { LineChart } from '@/components/pages/chart';
+import {connection} from "next/server";
+
+async function fetchLineChartData() {
+  // We don't want to render this page statically
+  await connection();
+  
+  const response = await fetch('http://localhost:5144/weatherforecast', {
```
gareth-allan-bjss, line 8: This needs to use `process.env.FINGERTIPS_API_URL`, as is done in `frontend/fingertips-frontend/app/page.tsx`. This is probably a good point to extract a little utility function to the `lib` directory (please not in `utils.js`, I want to avoid having a single general-purpose "utilities" file!).

frontend/fingertips-frontend/app/chart/page.tsx
```diff
@@ -0,0 +1,27 @@
...
+    next: { revalidate: 60 },
+  });
+  
+  if (!response.ok) {
+    throw new Error('Failed to fetch forecast data');
+  }
+  return await response.json();
+}
+
+export default async function getLineChartData() {
```
gareth-allan-bjss, line 19: I don't think this function should be called `getLineChartData`, as it's the main function for this page. Perhaps `ChartPage`?

gareth-allan-bjss: Also, can we not use a default export here please? As [documented in our new coding standards](https://confluence.collab.test-and-trace.nhs.uk/display/FTN/Coding+Standards#CodingStandards-UseNamedExportsRatherthanDefaultExports) (I appreciate I haven't publicised them yet!).

priya-thorpe-bjss: It can be called ChartPage, I don't have access to confluence pages yet. Could you send me a pdf of the coding standards please?

frontend/fingertips-frontend/components/pages/chart/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+            {data.map((item) => (
+              <Table.Row key={`${item.date}-${item.temperatureC}`}>
+                <Table.Cell>{item.date}</Table.Cell>
+                <Table.Cell numeric>{item.temperatureC}</Table.Cell>
+                <Table.Cell numeric>{item.temperatureF}</Table.Cell>
+                <Table.Cell>{item.summary}</Table.Cell>
+              </Table.Row>
+            ))}
+          </Table>
+        </StyledComponentsRegistry>
```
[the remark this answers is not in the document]
gareth-allan-bjss: Given we know we'll need to display a table alongside a chart, even when JS is enabled, should we just take this table out of the `noscript` tag for now and avoid having to wrap it in `StyledComponentsRegistry` again?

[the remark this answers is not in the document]
alibombaye: Yeah that's a fair shout. Happy for it to be removed 👍

frontend/fingertips-frontend/lib/fetchData.ts
```diff
@@ -0,0 +1,26 @@
+import {
+  Configuration,
+  WeatherForecast,
+  WeatherForecastApi,
+} from '@/generated-sources/api-client';
+import { connection } from 'next/server';
+
+export async function fetchData(): Promise<WeatherForecast[]> {
```
gareth-allan-bjss, line 8: I don't think we're going to be able to have a generic `fetchData` function as we expand the application to call multiple API endpoints with different parameters. Perhaps what would be useful is keeping this function as something that reads the API URL environment variable and creates (and returns) a `Configuration` object, and then we can leave the part that actually uses the `Configuration` to call the specific API in the page?

frontend/fingertips-frontend/lib/fetchData.ts
```diff
@@ -0,0 +1,26 @@
...
+  // We don't want to render this page statically
+  await connection();
```
gareth-allan-bjss, line 10: I think this `connection()` call should stay in the page, as it's about the behaviour of the page, and not necessarily something we always want to call when this function is called.

frontend/fingertips-frontend/components/table/index.tsx
```diff
@@ -0,0 +1,34 @@
+import { Table } from 'govuk-react';
+import { WeatherForecast } from '@/generated-sources/api-client';
+
+interface tableProps {
+  data: WeatherForecast[];
+}
+
+export function PlainTable({ data }: Readonly<tableProps>) {
```
gareth-allan-bjss, line 8: I think we (now) know that the tables will be different for different chart types, so perhaps worth calling this component `LineChartTable`?

frontend/fingertips-frontend/package.json
```diff
@@ -72,6 +74,8 @@
     "npm": ">=10.9.0"
   },
   "overrides": {
+    "react": "$react",
+    "react-dom": "$react-dom",
```
gareth-allan-bjss, line 78: Do we need these overrides any more? We're now on the stable release of React 19 so hopefully the `highcharts-react-official` peer dependency is fulfilled by that.

frontend/fingertips-frontend/package-lock.json
gareth-allan-bjss, line 1: It looks like a `resolved` and `integrity` field have been added for all dependencies in this file, which I wouldn't expect from your changes to `package.json`. What version of Node/NPM are you running? It might be worth reverting your changes to the `package-lock.json` and running `npm install` again.

frontend/fingertips-frontend/components/linechartsOptions/index.tsx
gareth-allan-bjss, line 1: I don't think this component is in the right place in the directory structure. I think it should be under the `molecules` directory (do you agree it's a molecule rather than an organism @alibombaye?) and have a name that matches the component name (i.e. `LineChart`). So this file should be moved to `frontend/fingertips-frontend/components/molecules/LineChart/index.tsx`.

frontend/fingertips-frontend/components/table/index.tsx
gareth-allan-bjss, line 1: Again, I think this file should be in `frontend/fingertips-frontend/components/molecules/LineChartTable/index.tsx` (assuming we're renaming the component to `LineChartTable`).

frontend/fingertips-frontend/components/pages/chart/index.tsx
```diff
@@ -0,0 +1,29 @@
+'use client';
```
gareth-allan-bjss, line 1: Can we move this `use client` directive down into the `LineChart` component? As I think it's the `HighChartsReact` component that requires us to use it, isn't it?

frontend/fingertips-frontend/components/pages/chart/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+type ChartProps = {
+  data: WeatherForecast[];
+};
+
+export function Chart({ data }: Readonly<ChartProps>) {
+  return (
+    <>
+      <H1>Line Chart</H1>
+      <div>
```
gareth-allan-bjss, line 16: Given the `LineChart` and `PlainTable` components already wrap a `<div>` around everything within them, are the `div`s wrapped around the components at this level redundant?

frontend/fingertips-frontend/components/pages/chart/chart.test.tsx
```diff
@@ -0,0 +1,39 @@
...
+  const highcharts = screen.getByTestId('highcharts-react-component');
+  expect(highcharts).toBeInTheDocument();
+});
+
+test('should render the LineChart component with the correct props', () => {
+  render(<Chart data={mockData} />);
+
+  const lineChartTitle = screen.getByText('Line Chart');
+  expect(lineChartTitle).toBeInTheDocument();
+});
```
gareth-allan-bjss, line 32: This is just checking that the title has been rendered, so the test name probably needs updating to reflect that.

frontend/fingertips-frontend/components/pages/chart/chart.test.tsx
```diff
@@ -0,0 +1,39 @@
...
+  {
+    date: new Date('2024-11-01T00:00:00.000Z'),
+    temperatureC: 0,
+    temperatureF: 32,
+    summary: 'Bracing',
+  },
+];
+
+test('should render the Highcharts react component', () => {
+  render(<Chart data={mockData} />);
```
gareth-allan-bjss, line 22: @alibombaye: Should we be wrapping this `Chart` component in a call to `registryWrapper`?

frontend/fingertips-frontend/components/pages/chart/chart.test.tsx
```diff
@@ -0,0 +1,39 @@
...
+  const highcharts = screen.getByTestId('highcharts-react-component');
+  expect(highcharts).toBeInTheDocument();
+});
```
gareth-allan-bjss, line 25: I think this could reference the `lineChart-component` data-test-id value, so we don't have to specifically check for Highcharts presence at this level (we could porentially change the implementation of the chart to use another library, and we wouldn't want the chart page test to fail).

```suggestion
test('should render the line chart', () => {
  render(<Chart data={mockData} />);
  const lineChart = screen.getByTestId('lineChart-component');
  expect(lineChart).toBeInTheDocument();
});
```

frontend/fingertips-frontend/components/linechartsOptions/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+  title?: string;
+  xAxisTitle?: string;
+  yAxisTitle?: string;
+}
+
+export function LineChart({
+  data,
+  title = 'Weather Forecast',
+  xAxisTitle = 'Date',
+  yAxisTitle = 'Temperature (°C)',
```
gareth-allan-bjss, line 16: I don't think we need default values here. I don't think there are any values that can sensibly be applied to all uses of this component, so we should require them to always be provided.

frontend/fingertips-frontend/components/linechartsOptions/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+    series: [
+      {
+        type: 'line',
+        name: yAxisTitle,
+        data: temperatureData,
+      },
+    ],
+    accessibility: {
+      enabled: true,
+      description: 'A line chart showing temperature data over time.',
```
gareth-allan-bjss, line 42: This description should be a prop too.

frontend/fingertips-frontend/components/table/index.tsx
gareth-allan-bjss, line 1: I think we should have some tests for this component. It could presumably have a snapshot test, couldn't it?

frontend/fingertips-frontend/components/table/index.tsx
```diff
@@ -0,0 +1,34 @@
...
+  return (
+    <div data-testid="plainTable-component">
+      <Table
+        head={
+          <Table.Row>
+            <Table.CellHeader date>Date</Table.CellHeader>
+            <Table.CellHeader numeric>Temperature C</Table.CellHeader>
+            <Table.CellHeader numeric>Temperature F</Table.CellHeader>
+            <Table.CellHeader>Summary</Table.CellHeader>
```
gareth-allan-bjss, line 17: We should be passing in these heading values as props, in the same way as we do for the `LineChart` component.

frontend/fingertips-frontend/components/linechartsOptions/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+      title: { text: yAxisTitle },
+    },
+    series: [
+      {
+        type: 'line',
+        name: yAxisTitle,
+        data: temperatureData,
+      },
+    ],
+    accessibility: {
```
gareth-allan-bjss, line 40: Did you say the accessibility module wasn't working? [This GitHub issue](https://github.com/highcharts/highcharts-react/issues/141) might help to get it sorted. I think we should be trying to include it in this PR, if it doesn't take too much time (and making sure we address it later if not).

gareth-allan-bjss (commented):
The structure is looking much better now, nice work. Just a few more comments, but I don't think any are too major.

frontend/fingertips-frontend/package.json
gareth-allan-bjss: I've tried this, and I think they can be removed.

frontend/fingertips-frontend/package-lock.json
gareth-allan-bjss: I did this by running the following in the `frontend/fingertips-frontend` directory (after removing the `overrides` values for `react` and `react-dom`:
```
nvm use
git checkout origin/main package-lock.json
npm install
```

Give me a shout if this doesn't work for you.

frontend/fingertips-frontend/components/linechartsOptions/index.tsx
priya-thorpe-bjss: The fix I put in previously removed the accessibility console error, I removed the fix because when JS was disabled on the browser nothing rendered on the page, no title or table. Im happy to revert it back, but not sure if we want a blank page when JS is disabled.

gareth-allan-bjss: We definitely don't want a blank page. Perhaps not one to worry about in this PR then, as long as we make sure to look at it in a future one.

gareth-allan-bjss (approved):
Looks good to me, thanks! Just a few comments on naming, but I'm happy for this to go in once they're addressed.

gareth-allan-bjss (approved):

PR #26, opened 2024-12-16
Feature/FTN-37-Logging

steves-bjss:
# Description

Jira ticket: FTN-37

Introduction of logging / monitoring for the front-end and API.

## Changes

- Connected front-end to Application Insights for monitoring / logging
- Connected back-end to Application insights for monitoring / logging
- Example of compile time logging

## Validation

Changes were tested manually in the Azure Portal to ensure exceptions were being logged appropriately.


frontend/fingertips-frontend/.env.development
```diff
@@ -1 +1,2 @@
 FINGERTIPS_API_URL="http://localhost:5144"
+APPLICATIONINSIGHTS_CONNECTION_STRING=
```
gareth-allan-bjss, line 2: Is it worth adding something to the README about how to configure this locally (this file is for when running the Next dev server only). Is that something we'd want to support? If not, I'd suggest we don't include this here.

api/DHSC.FingertipsNext.Modules.Core/LoggingExtensions.cs
```diff
@@ -0,0 +1,7 @@
+namespace DHSC.FingertipsNext.Modules.Core;
+
+public static partial class LoggingExtensions
+{
+    [LoggerMessage(0, LogLevel.Error, "Log Example failed operation during: {Context}")]
```
gareth-allan-bjss, line 5: Should both log examples have an event ID of 0?

api/DHSC.FingertipsNext.Modules.Core.Service/LoggingExtensions.cs
```diff
@@ -0,0 +1,9 @@
+using Microsoft.Extensions.Logging;
+
+namespace DHSC.FingertipsNext.Modules.Core.Service;
+
+public static partial class LoggingExtensions
+{
+    [LoggerMessage(EventId = 0, Level = LogLevel.Error, Message = "Search Specific Log Example failed operation during: `{Context}`")]
+    public static partial void LogExampleTwo(this ILogger logger, string context, Exception exception);
```
gareth-allan-bjss, line 8: Just for my understanding, it's possible to define log messages in this way that don't take an exception parameter, right? i.e. so we can log non-error messages?

frontend/fingertips-frontend/next.config.ts
```diff
@@ -5,6 +5,16 @@ const nextConfig: NextConfig = {
...
     styledComponents: true,
   },
+  serverExternalPackages: ['@azure/monitor-opentelemetry',
+    '@opentelemetry/api',
+    '@opentelemetry/resources',
+    '@opentelemetry/sdk-metrics',
+    '@opentelemetry/sdk-node',
+    '@opentelemetry/sdk-trace-base',
+    '@opentelemetry/semantic-conventions',
+    '@opentelemetry/exporter-jaeger'],
```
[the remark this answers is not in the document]
steves-bjss: It does indeed - its a dependency of one of the other packages and gets bundled in without it being specifically excluded.

api/DHSC.FingertipsNext.Modules.Core.Service/LoggingExtensions.cs
steves-bjss: You can yes - you would just excluded the exception in the method parameters, e.g.

 public static partial void LogExampleTwo(this ILogger logger, string context)

frontend/fingertips-frontend/package.json
```diff
@@ -18,6 +18,13 @@
     "prettier-ci": "prettier . --check"
   },
   "dependencies": {
+    "@azure/monitor-opentelemetry": "^1.8.0",
+    "@opentelemetry/api": "^1.9.0",
+    "@opentelemetry/resources": "^1.29.0",
+    "@opentelemetry/sdk-metrics": "^1.29.0",
+    "@opentelemetry/sdk-node": "^0.56.0",
+    "@opentelemetry/sdk-trace-base": "^1.29.0",
+    "@opentelemetry/semantic-conventions": "^1.28.0",
```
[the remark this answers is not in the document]
steves-bjss: These are specified packages from Microsoft in-order to connect a Node application to App Insights:
https://learn.microsoft.com/en-us/azure/azure-monitor/app/opentelemetry-enable?tabs=nodejs

api/DHSC.FingertipsNext.Modules.Core/LoggingExtensions.cs
steves-bjss: Fair - I've updated one of them to show a different event ID

frontend/fingertips-frontend/instrumentation.ts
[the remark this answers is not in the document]
steves-bjss: Updated as suggested.

frontend/fingertips-frontend/.env.development
steves-bjss: You may want to test it / work on it but not as a matter of course.  I've removed it for now - people can always add it back in.

PR #28, opened 2024-12-18
FTN 65: Set up Azure AI search

adawari-thomas-bjss:
# Description

Jira ticket: [FTN-65](https://example.com)

## Changes

- Updated Terraform to create/manage Azure AI search instance
- Added a console app to create AI search instance using the Azure SDK for Typescript
- Use console app to programmatically push data to index

## Validation

Validated in Azure to confirm that the expected AI search instance was created, alongside it's index, and the index was populated with the provided sample data.


terraform/environment/ai_search.tf
```diff
@@ -0,0 +1,14 @@
+resource "azurerm_resource_group" "az_search_rg" {
+  name     = "${local.resource_prefix}-rg-search"
+  location = var.region
+  tags     = local.tags
+}
+
+resource "azurerm_search_service" "search" {
+  name                = "${local.resource_prefix}-az-search"
```
gareth-allan-bjss, line 8: I've been trying to stick to the [Abbreviation recommendations for Azure resources](https://learn.microsoft.com/en-us/azure/cloud-adoption-framework/ready/azure-best-practices/resource-abbreviations) when referencing Azure services in resource names (I appreciate this isn't necessarily stated anywhere though!). So I think this should be:
```suggestion
  name                = "${local.resource_prefix}-srch"
```

terraform/environment/ai_search.tf
```diff
@@ -0,0 +1,14 @@
+resource "azurerm_resource_group" "az_search_rg" {
+  name     = "${local.resource_prefix}-rg-search"
```
gareth-allan-bjss, line 2: Again from the "official" MS abbreviations I think this should be:
```suggestion
  name     = "${local.resource_prefix}-rg-srch"
```

terraform/environment/ai_search.tf
```diff
@@ -0,0 +1,14 @@
...
+  tags     = local.tags
+}
+
+resource "azurerm_search_service" "search" {
+  name                = "${local.resource_prefix}-az-search"
+  resource_group_name = azurerm_resource_group.az_search_rg.name
+  location            = azurerm_resource_group.az_search_rg.location
+  sku                 = var.ai_search_sku
+  replica_count       = var.ai_search_replica_count
+  partition_count     = var.ai_search_partition_count
```
gareth-allan-bjss, line 13: Can you make sure to add tags to the AI search resource please? I've been using them to make sure all resources are tagged with the environment name (which helps map resources to an environment, since the actual resource names contain a hash, to avoid issues in name lengths).

```suggestion
  partition_count     = var.ai_search_partition_count
  tags                = local.tags
```

gareth-allan-bjss (commented):
A few more minor comments on the TF.

search-setup/package.json
```diff
@@ -1,10 +1,13 @@
...
   "type": "module",
   "scripts": {
-    "dev": "tsc && node --env-file=.env main.js"
+    "create-index": "tsc && node --env-file=.env main.js"
   },
   "devDependencies": {
+    "@types/jest": "^29.5.14",
     "@types/node": "^22.10.2",
+    "jest": "^29.7.0",
+    "ts-jest": "^29.2.5",
```
gareth-allan-bjss, line 10: Why install jest-related dependencies? There's nothing that runs using Jest, is there?

search-setup/indexOperations.ts
```diff
@@ -0,0 +1,64 @@
...
+}
+
+export async function populateIndex(
+  searchClient: SearchClient<Data>,
+  indexData: Data[],
+  indexName: string
+): Promise<void> {
+  await searchClient.mergeOrUploadDocuments(indexData);
+  console.log(`Uploaded data to index with name: ${indexName}`);
+}
```
gareth-allan-bjss, line 26: I think you could get the index name from `searchClient` in order to avoid having to pass it in as an argument:
```suggestion
export async function populateIndex(
  searchClient: SearchClient<Data>,
  indexData: Data[]
): Promise<void> {
  await searchClient.mergeOrUploadDocuments(indexData);
  console.log(`Uploaded data to index with name: ${searchClient.indexName}`);
}
```

gareth-allan-bjss (approved):
Looks good to me now. Just a couple of minor comments that don't have to be blockers.

With regard to the Sonar quality gate failure, the two issues are to do with the new Terraform resource for the AI search service and the fact that we're not restricting internet access to it and not using Azure managed identities. Given we'll need to allow internet access to the AI search service until we get an app gateway implementation in place, the fact that all requests to it are required to provide an API key, and the fact that we're not actually deploying this terraform anywhere yet I'm happy that we can merge it as-is.

search-setup/package.json
adawari-thomas-bjss: I made the changes on the unit testing branch, then merged it into this one, so this must've pulled in as well. Leaving as is because the other branch will use the same dependencies when this gets merged

gareth-allan-bjss: Fair enough, I suspected this might be the case after I made the comment!

PR #29, opened 2024-12-18
FTN-58: Search by indicator id

ataylor-bjss:
# Description

Jira ticket: FTN-58

## Changes

- Etc.

## Validation


frontend/fingertips-frontend/lib/search/searchResultData.ts
```diff
@@ -0,0 +1,56 @@
...
+  const day = String(date.getDate()).padStart(2, '0');
+  const month = date.toLocaleString('en-GB', { month: 'long' });
+  const year = date.getFullYear();
+
+  return `${day} ${month} ${year}`;
+}
+
+let searchService: SearchService;
+try {
+  searchService = new SearchService();
```
alibombaye, line 42: If I'm understanding this correctly. This will be a singleton of the SearchService instance on the NextJS server to handle all requests. Rather than create a new instance per request. Is this the intended idea here and does this have any draw backs? Just thinking performance, load or concurrency.

frontend/fingertips-frontend/lib/search/searchService.ts
```diff
@@ -0,0 +1,62 @@
...
+};
+
+const getEnvironmentVariable = (name: string): string => {
+  return (
+    process.env[name] ??
+    (() => {
+      throw new Error(`Missing environment variable ${name}`);
+    })()
+  );
+};
```
gareth-allan-bjss, line 23: This might make more sense in its own file, as I expect we'll want to use it from other places.

frontend/fingertips-frontend/lib/search/searchResultData.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export const getSearchService = (): IndicatorSearch => {
+  return searchService
+    ? searchService
+    : {
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
+        searchByIndicator(indicator: string): Promise<IndicatorSearchResult[]> {
+          return Promise.resolve(MOCK_DATA);
+        },
+      };
```
gareth-allan-bjss, line 55: It might be cleaner to implement this as a separate implementation of the `IndicatorSearch` interface.

frontend/fingertips-frontend/lib/search/searchService.ts
```diff
@@ -0,0 +1,62 @@
...
+export class SearchService implements IndicatorSearch {
+  readonly searchClient: SearchClient<Indicator>;
+
+  constructor() {
+    const serviceName = getEnvironmentVariable('DHSC_AI_SEARCH_SERVICE_NAME');
+    const indexName = getEnvironmentVariable('DHSC_AI_SEARCH_INDEX_NAME');
+    const apiKey = getEnvironmentVariable('DHSC_AI_SEARCH_API_KEY');
+
+    this.searchClient = new SearchClient<Indicator>(
+      `https://${serviceName}.search.windows.net`,
```
gareth-allan-bjss, line 34: I appreciate this was how I did it in the PoC, but I think we're probably better taking the full URL to the search service as an environment variable (e.g. `https://<serviceName>.search.windows.net`), as it means we're not baking in knowledge of the AI search URL structure to our code.

frontend/fingertips-frontend/lib/search/searchService.ts
```diff
@@ -0,0 +1,62 @@
...
+      indexName,
+      new AzureKeyCredential(apiKey)
+    );
+  }
+
+  async searchByIndicator(indicator: string): Promise<IndicatorSearchResult[]> {
+    const query = `/.*${indicator}.*/`;
+
+    const searchResponse = await this.searchClient.search(query, {
+      queryType: 'full',
```
gareth-allan-bjss, line 44: Do we need `queryType: 'full'` here? I _think_ I had to add that to allow fuzzy search to be enabled (which you're not using here).

frontend/fingertips-frontend/lib/search/searchResultData.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  return `${day} ${month} ${year}`;
+}
+
+let searchService: SearchService;
+try {
+  searchService = new SearchService();
+} catch {
+  // Handle error
+}
```
gareth-allan-bjss, line 45: Do we want to fall back to hardcoded data if we can't set up our search service?

frontend/fingertips-frontend/lib/search/searchResultData.ts
ataylor-bjss: Not sure, although I did think that performing the initial set-up of the search client on each call would be a repetition of effort. My thinking was if Node is single-threaded (think it might be) then creating a new instance each time doesn't help as it's just repeated work, & if it's multi-threaded then there's nothing to stop different threads running the same function.

frontend/fingertips-frontend/lib/search/searchResultData.ts
ataylor-bjss: Not in the longterm but thought for now this made most sense; it means the application behaviour remains the same if the search service isn't configured and can still run locally for people without their having to access the AI search.

frontend/fingertips-frontend/lib/search/searchService.ts
ataylor-bjss: Needs to be `full` if you want to use Lucene query syntax, which I think includes wildcards (quick test seems to confirm this).

frontend/fingertips-frontend/lib/search/searchResultData.ts
gareth-allan-bjss: Discussed offline, and we'll do this for local development/our current set of E2E tests (that don't run against a deployed environment yet).

frontend/fingertips-frontend/lib/search/searchService.ts
gareth-allan-bjss: I'm not sure you need the full Lucene syntax for wildcards, but I'm happy to leave it using this syntax for now.

frontend/fingertips-frontend/lib/search/searchResultData.ts
```diff
@@ -0,0 +1,26 @@
+import { SearchService } from './searchService';
+import { SearchServiceMock } from './searchServiceMock';
+
+export interface BasicSearchResult {
```
gareth-allan-bjss, line 4: I think I preferred the name `IndicatorSearchResult` we were using previously, given this is specifically a type for results from the indicator search index.

frontend/fingertips-frontend/lib/search/searchService.ts
```diff
@@ -0,0 +1,74 @@
...
+      'DHSC_AI_SEARCH_SERVICE_URL'
+    ) as string;
+    this.indexName = getEnvironmentVariable(
+      'DHSC_AI_SEARCH_INDEX_NAME'
+    ) as string;
+    this.apiKey = getEnvironmentVariable('DHSC_AI_SEARCH_API_KEY') as string;
+  }
+
+  async searchWith(searchTerm: string): Promise<BasicSearchResult[]> {
+    const query = `${searchTerm} /.*${searchTerm}.*/`;
```
gareth-allan-bjss, line 38: We're adding the search term to the query twice? Once with a wildcard and once without? What's that doing?

gareth-allan-bjss: I'm happy you've added a comment to explain this for now.

frontend/fingertips-frontend/app/search/results/page.tsx
```diff
@@ -11,8 +12,24 @@ export default async function Page(
...
+    // Perform async API call using indicator prop
+    const searchResults = await getSearchService().searchWith(indicator);
+    return (
+      <SearchResults indicator={indicator} searchResults={searchResults} />
+    );
+  } catch (error) {
+    // Log error response
+    // TBC
+    console.log(
+      `Error response received from call to Search service: ${error}`
```
HowardMayBjss, line 25: There are various types of search - it would be helpful to be specific e.g. Error response received from call to Indicator Search service.

frontend/fingertips-frontend/lib/search/searchService.ts
```diff
@@ -0,0 +1,77 @@
...
+    this.apiKey = getEnvironmentVariable('DHSC_AI_SEARCH_API_KEY') as string;
+  }
+
+  async searchWith(searchTerm: string): Promise<IndicatorSearchResult[]> {
+    // Search with both full search term and wildcard for now to allow for
+    // observed behaviour where wildcard match of full term did not always return
+    // the matching document.
+    const query = `${searchTerm} /.*${searchTerm}.*/`;
+
+    const searchClient = new SearchClient<Indicator>(
```
HowardMayBjss, line 43: This is creating a new SearchClient on each search request - I would presume we want to create a single client when our SearchService is instantiated.

frontend/fingertips-frontend/lib/search/searchResultData.ts
HowardMayBjss: The code has been changed to only use the mock server if DHSC_AI_SEARCH_USE_MOCK_SERVICE is true

frontend/fingertips-frontend/lib/search/searchService.ts
```diff
@@ -0,0 +1,77 @@
...
+
+    const scoringProfile = getEnvironmentVariable(
+      'DHSC_AI_SEARCH_SCORING_PROFILE',
+      false
+    );
+    if (scoringProfile) {
+      searchOptions.scoringProfile = scoringProfile;
+    }
+
+    const searchResponse = await searchClient.search(query, searchOptions);
```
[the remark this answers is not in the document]
HowardMayBjss: Typing is being used on the SearchClient - but the structure we want to return the data in is a little different and so this mapping is still needed.

gareth-allan-bjss:
This PR has been superseded by #59.

PR #32, opened 2024-12-19
DHSCFT-25: Link results page to chart page

alibombaye:
# Description

Jira ticket: [DHSCFT-25](https://bjss-enterprise.atlassian.net/browse/DHSCFT-25)

- Fixes rendering of govuk-react components
- Adds checkboxes to the search results page
- Manages search state using url to navigate back and forth from the search pages to the chart page.

Screenshot 2024-12-20 at 17 08 45


frontend/fingertips-frontend/app/chart/page.tsx
```diff
@@ -1,15 +1,31 @@
...
   // We don't want to render this page statically
   await connection();
 
   const config = getApiConfiguration();
   const forecastApi = new WeatherForecastApi(config);
   const data = await forecastApi.getWeatherForecast();
 
-  return <Chart data={data} />;
+  return (
+    <Chart
```
williamwillis11, line 25: should the chart have a dynamic data-testid ? for test automation

frontend/fingertips-frontend/components/pages/chart/index.tsx
```diff
@@ -2,18 +2,29 @@
...
+  data,
+  indicator,
+  indicatorsSelected = [],
+}: Readonly<ChartProps>) {
+  const searchState = new SearchStateManager({ indicator, indicatorsSelected });
+  const backLinkPath = searchState.generatePath('/search/results');
+
   return (
     <>
+      <BackLink data-testid="chart-page-back-link" href={backLinkPath} />
```
williamwillis11, line 27: can you add a unit test to

frontend/fingertips-frontend/components/pages/chart/chart.test.tsx

for clicking the backlink?

frontend/fingertips-frontend/components/pages/search/results/Results.test.tsx
```diff
@@ -57,7 +76,55 @@ describe('Search Results Suite', () => {
     expect(screen.queryAllByTestId('search-result')).toHaveLength(0);
   });
 
-  test('snapshot test', () => {
+  it('should mark indicators selected as checked', () => {
```
williamwillis11, line 79: happy to go with _it_ over _test_

however can see we are now split between _it_ and _test_ language across the unit tests. Please could you do a find and replace to change all the tests all over to _it_ language?

frontend/fingertips-frontend/playwright/tests/search.spec.ts
```diff
@@ -48,3 +48,30 @@ test('Should return to search page upon results page back link click', async ({
   // Assert
   await searchPage.checkURLIsCorrect(`?indicator=${indicator}`);
 });
+
+test('Should navigate to chart page and clicking back should keep checkboxes and search term selected', async ({
```
williamwillis11, line 52: as discussed - on a follow-on ticket i will merge the 4 e2e tests we currently have into 1 to avoid duplication which is creeping in

frontend/fingertips-frontend/app/chart/page.tsx
alibombaye: Yeah that's right the data-testid for it is within the component itself

frontend/fingertips-frontend/components/molecules/Search/result/searchResult.test.tsx
```diff
@@ -3,22 +3,39 @@ import { expect } from '@jest/globals';
...
+    ...originalModule,
+    usePathname: () => mockPath,
+    useSearchParams: () => ({ indicator: 'test' }),
+    useRouter: jest.fn().mockImplementation(() => ({
+      replace: mockReplace,
+    })),
+  };
+});
 
 describe('Search Result Suite', () => {
```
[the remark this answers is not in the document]
alibombaye: yes there should be. Good shout

frontend/fingertips-frontend/components/pages/search/results/Results.test.tsx
alibombaye: I'll make that change in the follow up PR https://github.com/dhsc-govuk/FingertipsNext/pull/38 👍

frontend/fingertips-frontend/components/molecules/Search/result/searchResult.test.tsx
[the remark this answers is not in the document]
alibombaye: Added a unit test in `Results.test.tsx` to check for a back link and it's href

frontend/fingertips-frontend/components/pages/chart/index.tsx
alibombaye: Added a test. Not quite clicking but just checking the href of the link is as expected.

frontend/fingertips-frontend/components/pages/chart/chart.test.tsx
```diff
@@ -19,6 +19,17 @@ const mockData: WeatherForecast[] = [
...
 
+it('should render the backLink', () => {
+  render(
+    <Chart data={mockData} indicator="test" indicatorsSelected={['1', '2']} />
+  );
+
+  expect(screen.getByRole('link', { name: /back/i })).toBeInTheDocument();
+  expect(screen.getByRole('link', { name: /back/i }).getAttribute('href')).toBe(
+    '/search/results?indicator=test&indicatorsSelected=1%2C2'
+  );
```
williamwillis11, line 30: should we setup a mockNavigate jest.fn() and also click the backlink 

then we could do something like 

expect(mockNavigate).toHaveBeenCalledTimes(1);
expect(mockNavigate).toHaveBeenCalledWith(URLOFWHEREWEEXPECTTOGO);

frontend/fingertips-frontend/components/pages/search/results/Results.test.tsx
```diff
@@ -40,6 +40,15 @@ describe('Search Results Suite', () => {
     );
   });
 
+  it('should render the backLink', () => {
+    render(<SearchResults indicator={indicator} searchResults={[]} />);
+
+    expect(screen.getByRole('link', { name: /back/i })).toBeInTheDocument();
+    expect(
+      screen.getByRole('link', { name: /back/i }).getAttribute('href')
+    ).toBe('/search?indicator=test');
```
williamwillis11, line 49: should we setup a mockNavigate jest.fn() and also click the backlink

then we could do something like

expect(mockNavigate).toHaveBeenCalledTimes(1);
expect(mockNavigate).toHaveBeenCalledWith(URLOFWHEREWEEXPECTTOGO);

frontend/fingertips-frontend/instrumentation.ts
```diff
@@ -4,7 +4,9 @@ const startMockServer = async () => {
     process.env.MOCK_SERVER === 'true'
   ) {
     const { server } = await import('./mock/server/node');
-    server.listen();
+    server.listen({
+      onUnhandledRequest: 'bypass',
```
gareth-allan-bjss, line 8: What's this doing?

frontend/fingertips-frontend/components/pages/search/results/Results.test.tsx
alibombaye: Since it's just using a normal link and not via next/navigation we can't go about it this way.

frontend/fingertips-frontend/instrumentation.ts
alibombaye: There are some warning logs from MSW for unhandledRequests on the console. This was added to disable them and reduce that noise

williamwillis11: for context @gareth-allan-bjss, when running the e2e playwright tests i was seeing some unhandledRequests in the browser devtools console so i reported them to @alibombaye

gareth-allan-bjss: Do you know which requests weren't being handled? Was it causing issues with the tests? I'm not necessarily against turning the warning message off, but I'd assume it's defaulting to logging a message to make it easier to tell when your handlers are configured incorrectly, which might be nice to have.

alibombaye: Yeah it's a strange one this. It's request to the nextjs server and it's for any request where we are redirecting to a url with a query param.

eg? `GET http://localhost:3000/search/results?indicator=testing?indicator=testing`

For some reason it's got the query string in there twice. I have got no idea why it's doing this. All the tests pass and I've added logs to before we do the redirect to check the path and it's as expected. The tests and when running manually all work as expected. 

Investigated the network when running to see if it's making calls with this odd url and can't see anything. 

If the handler is configured incorrectly then we still get an error to indicate the interceptor couldn't find a response
Screenshot 2025-01-06 at 13 46 34

With this the best I could do was to just disable these warnings.

frontend/fingertips-frontend/components/molecules/Search/result/index.tsx
```diff
@@ -30,18 +33,49 @@ const StyledRow = styled(GridRow)(
...
+  indicatorSelected,
+}: Readonly<SearchResultProps>) {
+  const searchParams = useSearchParams();
+  const pathname = usePathname();
+  const { replace } = useRouter();
+
+  const handleClick = (indicatorId: string, checked: boolean) => {
+    const params = new URLSearchParams(searchParams);
+    const searchState = new SearchStateManager();
+    searchState.setStateFromParams(params);
```
gareth-allan-bjss, line 47: I'm wondering if we could use a React context to provide this state (the indicator searched for and the indicators selected) to child components? It would mean we could read them from query params in the page and not have to do it (again) in low-level components like this. What do you think?

frontend/fingertips-frontend/components/molecules/Search/result/searchResult.test.tsx
```diff
@@ -33,7 +50,42 @@ describe('Search Result Suite', () => {
     );
   });
 
-  test('snapshot test', () => {
+  it('should mark the checkbox as checked if indicatorSelected is true', () => {
+    render(
+      registryWrapper(
```
gareth-allan-bjss, line 55: I thought you were getting rid of `registryWrapper`?

frontend/fingertips-frontend/components/molecules/Search/result/searchResult.test.tsx
```diff
@@ -33,7 +50,42 @@ describe('Search Result Suite', () => {
...
+
+    await user.click(screen.getByRole('checkbox'));
+
+    expect(mockReplace).toBeCalledWith(
+      `${mockPath}?indicator=test&indicatorsSelected=1`,
+      {
+        scroll: false,
+      }
+    );
+  });
```
gareth-allan-bjss, line 86: I think we should have an equivalent of this that verifies unchecking an initially-checked checkbox removes the expected query parameter.

frontend/fingertips-frontend/lib/searchStateManager.ts
```diff
@@ -0,0 +1,75 @@
+export const encodedCommaSeperator = encodeURIComponent(',');
+
+export type SearchState = {
+  indicator?: string | null;
+  indicatorsSelected?: string[] | null;
+};
+
+export class SearchStateManager {
+  private searchState: SearchState | undefined | null;
```
gareth-allan-bjss, line 9: Why does this need to be of type `SearchState | undefined | null`? You always set it in the constructor, so I'm not sure when it'll ever be `undefined` or `null`.

frontend/fingertips-frontend/lib/searchStateManager.ts
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  public generatePath(path: string) {
+    const generatedPath = [];
+    generatedPath.push(path);
+
+    if (this.hasState()) {
+      generatedPath.push(this.addIndicatorToPath());
+      generatedPath.push(this.addIndicatorsSelectedToPath());
+    }
```
gareth-allan-bjss, line 71: Do you need to wrap this in the `hasState()` check? Don't the individual methods being called result in the right behaviour when their related values aren't set anyway?

frontend/fingertips-frontend/lib/searchStateManager.ts
```diff
@@ -0,0 +1,75 @@
...
+      return `?indicator=${this.searchState.indicator}`;
+    }
+  }
+
+  private addIndicatorsSelectedToPath(): string | undefined {
+    if (
+      this.searchState?.indicatorsSelected &&
+      this.searchState?.indicatorsSelected.length > 0
+    ) {
+      return `&indicatorsSelected=${this.searchState.indicatorsSelected.join(encodedCommaSeperator)}`;
```
gareth-allan-bjss, line 37: This is assuming that a query parameter has already been added to the URL, as it's using the `&` to separate them. If an instance of `SearchStateManager` has been instantiated without an `indicator` value, but with `indicatorsSelected` values it's not going to generate a valid URL.

I think the answer is probably to make the `indicator` parameter mandatory, as I'm not sure it really makes sense to not have one set. If for some reason we can't, then we should probably use one of the standard utilities to manage adding query parameters, to avoid having to implement the logic ourselves (and add a unit test for the scenario).

frontend/fingertips-frontend/app/chart/page.tsx
```diff
@@ -1,15 +1,31 @@
...
-import { getApiConfiguration } from '@/lib/getApiConfiguration';
+import { getApiConfiguration } from '@/lib/fetchData';
 import { connection } from 'next/server';
 import { WeatherForecastApi } from '@/generated-sources/api-client';
 
-export default async function ChartPage() {
+export default async function ChartPage(
+  props: Readonly<{
+    searchParams?: Promise<{
+      indicator?: string;
```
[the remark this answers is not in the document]
alibombaye: It's to keep hold of the state so when we click on the back link we know the original indicator that was searched for and show that on the search results page.

frontend/fingertips-frontend/components/molecules/Search/result/index.tsx
alibombaye: I was thinking this also but React context will hold the state client side only and then we would have to keep the state in sync which could become bit difficult to manage. I think using the url as state management would be the easier option here.

frontend/fingertips-frontend/app/chart/page.tsx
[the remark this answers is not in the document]
gareth-allan-bjss: Sorry, I wasn't particularly clear. What I was asking was, does `indicator` need to be defined as an optional parameter? Will it ever (legitimately) not be set?

frontend/fingertips-frontend/components/molecules/Search/result/searchResult.test.tsx
alibombaye: In the follow up PR https://github.com/dhsc-govuk/FingertipsNext/pull/38

frontend/fingertips-frontend/components/pages/search/results/index.tsx
```diff
@@ -2,25 +2,42 @@
...
+    : false;
+};
+
 export function SearchResults({
   indicator,
+  indicatorsSelected,
   searchResults,
 }: Readonly<SearchResultsProps>) {
+  const initialState: SearchResultState = { indicators: indicatorsSelected };
+  const [state, formAction] = useActionState(viewCharts, initialState);
```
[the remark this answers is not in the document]
alibombaye: I've not factored in the error handling as part of this PR just yet. Since the scope was just to link the pages. I agree we should either disable the button or have some error message when trying to view a chart and not selecting any indicators which I think is the more GDS compliant option. But I've left this out till we get some clear requirements on this. 

If we do show the error message then we would need to use useActionState. We need to pass the formData to the function for the purpose of the redirecting to the chart page with the selected indicators.

frontend/fingertips-frontend/lib/fetchData.ts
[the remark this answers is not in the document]
alibombaye: Ah ok I had branch off the chart PR and I think this must have changed later on. Will update 👍

frontend/fingertips-frontend/package-lock.json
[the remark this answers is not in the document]
alibombaye: I've got this down by deleting my package-lock and then merging in main and then doing the `npm i`. But it's still surprisingly big. Will continue to investigate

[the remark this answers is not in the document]
alibombaye: Managed to get this sorted. Good spot 👍

frontend/fingertips-frontend/components/molecules/Search/result/index.tsx
alibombaye: I see what you mean by this now. I'll give that a go 👍

gareth-allan-bjss: Yes, I wasn't suggesting that we don't use the URL for state (that lets us bookmark pages/share links too), just that I think we could pull the details from the URL at the page level (maybe the layout? I'm not 100% sure which would make more sense) and use a context to allow access to it from components further down the tree. Probably not one to spend loads of time on, but if we can do it relatively simply it might make things cleaner.

frontend/fingertips-frontend/components/molecules/Search/result/searchResult.test.tsx
gareth-allan-bjss: I'm happy either way, but I had noticed you didn't add it in some other new tests in this PR, which I thought made sense. Happy to resolve this and do it in the follow up PR if you prefer.

frontend/fingertips-frontend/components/molecules/Search/result/index.tsx
alibombaye: It makes sense and yeah it should be cleaner and not have to pass down props all the way down. Just need to figure out how to do the react context 😂

frontend/fingertips-frontend/components/molecules/Search/result/searchResult.test.tsx
alibombaye: I didn't want to clutter this PR with too many changes - Resolved in the follow up PR

frontend/fingertips-frontend/components/molecules/Search/result/index.tsx
alibombaye: As discussed - agreed to not do this and tweak the SearchStateManager instead

frontend/fingertips-frontend/lib/searchStateManager.ts
alibombaye: Updated

frontend/fingertips-frontend/components/molecules/Search/result/searchResult.test.tsx
alibombaye: Done

frontend/fingertips-frontend/lib/searchStateManager.ts
alibombaye: Updated to have some logic to figure out the correct symbol to use

frontend/fingertips-frontend/lib/searchStateManager.ts
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  public generatePath(path: string) {
+    this.addPathName(path);
+
+    this.addIndicatorToPath();
+    this.addIndicatorsSelectedToPath();
+
+    return this.constructPath();
+  }
```
gareth-allan-bjss, line 86: If you call this function twice, isn't it going to add all the query parameters to the path again (you never clear `generatedPath`)? I don't think it's necessarily a problem for the way this class is used now, but it's not very logical behaviour.

frontend/fingertips-frontend/components/pages/search/results/index.tsx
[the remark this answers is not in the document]
gareth-allan-bjss: I think we agreed offline that you'd add some simple validation, so we'd keep the use of `useActionState`, right?

frontend/fingertips-frontend/app/chart/page.tsx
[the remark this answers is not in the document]
gareth-allan-bjss: I think we agreed offline that this edge-case handling could be covered under a future ticket.

frontend/fingertips-frontend/lib/searchStateManager.ts
alibombaye: added clearing out the generatedPath to an empty array as part addPathName function

frontend/fingertips-frontend/components/pages/search/results/index.tsx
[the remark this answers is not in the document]
alibombaye: Error handling added

frontend/fingertips-frontend/lib/searchStateManager.ts
alibombaye: Ah yeah that would be lot nicer. I'm going to work on a follow up ticket to change param `indicator` to `searchedIndicator` - ok if I do it as part of that and we can resolve this one.

alibombaye: Added to ticket https://bjss-enterprise.atlassian.net/browse/DHSCFT-164

gareth-allan-bjss: I'm happy to resolve on that basis.

frontend/fingertips-frontend/components/molecules/Search/result/index.tsx
```diff
@@ -59,6 +59,7 @@ export function SearchResult({
       <StyledRow>
         <GridCol>
           <Checkbox
+            id={result.id.toString()}
```
gareth-allan-bjss, line 62: Perhaps worth putting a prefix on this ID (like `search-results-indicator`), as just using the indicator's ID leads to checkboxes with pretty generic numeric IDs?

frontend/fingertips-frontend/components/pages/search/results/Results.test.tsx
```diff
@@ -88,55 +133,71 @@ describe('Search Results Suite', () => {
...
+    expect(
+      screen.getByTestId('search-result-form-error-summary')
+    ).toBeInTheDocument();
+  });
 
-    await user.click(screen.getByRole('checkbox', { name: /DHSC/i }));
-    await waitFor(() => {
-      expect(screen.getByRole('checkbox', { name: /DHSC/i })).toBeChecked();
-    });
+  it('should focus to the first checkbox when there clicking on the error link in the summary', async () => {
```
gareth-allan-bjss, line 164:
```suggestion
  it('should focus to the first checkbox when clicking on the error link in the summary', async () => {
```

frontend/fingertips-frontend/instrumentation.ts
gareth-allan-bjss: OK, fair enough. Might be worth digging into a little bit more in future, but not worth holding this PR up over.

gareth-allan-bjss (approved):
A few very minor comments, but I'm happy for this to go in.

frontend/fingertips-frontend/components/pages/search/results/searchResultsActions.ts
```diff
@@ -1,24 +1,51 @@
...
-
-  const searchState = new SearchStateManager({ indicator, indicatorsSelected });
+  const validatedFields = $SearchResultFormSchema.safeParse({
+    indicator: formData.get('searchedIndicator'),
+    indicatorsSelected: formData.getAll('indicator'),
+  });
+
+  if (!validatedFields.success) {
+    return {
+      indicator: formData.get('indicator')?.toString() ?? '',
```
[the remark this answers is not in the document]
alibombaye: yeah it should. Good spot 👍

PR #33, opened 2024-12-20
DHSCFT-50: Unit testing AI search

adawari-thomas-bjss:
# Description

Jira ticket: [DHSCFT-50](https://bjss-enterprise.atlassian.net/browse/DHSCFT-50)

## Changes

- Added unit tests to test for the successful creation and population of an index within existing AI search instance.

## Validation
image


search-setup/__tests__/search.test.ts
```diff
@@ -0,0 +1,51 @@
...
+    });
+
+    const index: SearchIndexResponse = await response.json();
+
+    expect(index.name).toBe(indexName);
+    expect(index.fields.length).toBe(2);
+    expect(index.fields.at(0)?.name).toBe("IID");
+    expect(index.fields.at(0)?.type).toBe("Edm.String");
+    expect(index.fields.at(0)?.key).toBe(true);
+    expect(index.fields.at(0)?.retrievable).toBe(true);
```
gareth-allan-bjss, line 30: I think we want to be checking for the value of the `searchable` field here, as it's important to being able to actually search on certain fields.

search-setup/__tests__/search.test.ts
```diff
@@ -0,0 +1,51 @@
...
+    expect(index.fields.at(1)?.name).toBe("Descriptive");
+    expect(index.fields.at(1)?.type).toBe("Edm.ComplexType");
+    expect(index.fields.at(1)?.fields?.length).toBe(2);
+    expect(index.fields.at(1)?.fields?.at(0)?.name).toBe("Name");
+    expect(index.fields.at(1)?.fields?.at(1)?.name).toBe("Definition");
```
gareth-allan-bjss, line 35: I think we need to check all the properties of the `Name` and `Definition` fields too (e.g. their type, whether they're searchable, etc.).

I think really we want to be checking all of the properties that we set for a field when we create it.

search-setup/__tests__/search.test.ts
```diff
@@ -0,0 +1,51 @@
+import { config } from "dotenv";
+import { DocumentResponse, SearchIndexResponse } from "../types";
+
+config();
+
+const searchEndpoint = process.env["AI_SEARCH_SERVICE_ENDPOINT"] ?? "";
+const indexName = process.env["AI_SEARCH_INDEX_NAME"];
+const apiKey = process.env["AI_SEARCH_API_KEY"] ?? "";
```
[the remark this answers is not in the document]
adawari-thomas-bjss: The compiler complains when the API key passed as a header could possibly be undefined. This way the test still fails when an api key is not provided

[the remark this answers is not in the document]
gareth-allan-bjss: I think if you throw an error if any of these values are undefined Typescript should be clever enough to know they'll never be undefined when you use them later. I think that makes more sense than continuing with the tests if we're missing critical information.

search-setup/__tests__/search.test.ts
```diff
@@ -0,0 +1,64 @@
...
+    expect(index.fields.at(1)?.fields?.at(0)?.retrievable).toBe(true);
+    expect(index.fields.at(1)?.fields?.at(0)?.searchable).toBe(true);
+    expect(index.fields.at(1)?.fields?.at(0)?.sortable).toBe(true);
+    expect(index.fields.at(1)?.fields?.at(0)?.filterable).toBe(true);
+    expect(index.fields.at(1)?.fields?.at(1)?.name).toBe("Definition");
+    expect(index.fields.at(1)?.fields?.at(1)?.type).toBe("Edm.String");
+    expect(index.fields.at(1)?.fields?.at(1)?.retrievable).toBe(true);
+    expect(index.fields.at(1)?.fields?.at(1)?.searchable).toBe(true);
+    expect(index.fields.at(1)?.fields?.at(1)?.sortable).toBe(true);
+    expect(index.fields.at(1)?.fields?.at(1)?.filterable).toBe(true);
```
gareth-allan-bjss, line 48: This list of assertions is already long, and our index only has three fields. I think we need to look at whether we can compare this against an object, or if that doesn't work at least having a function to check each field is as expected.

adawari-thomas-bjss: using helper methods to fix this as the object returned by the fetch call contains plenty fields that are not being used. This way, only the necessary fields are being tested.

gareth-allan-bjss (approved):

PR #36, opened 2025-01-02
Feature/dhscft 4 line chart new designs

priya-thorpe-bjss:
# Description

Jira ticket: [DHSCFT-4](https://bjss-enterprise.atlassian.net/browse/DHSCFT-4)

## Changes
- Calling the health data endpoint with an indicator and area code.
- Rendering the data on the line chart and table.

## Validation
Screenshot 2025-01-09 at 17 34 59


frontend/fingertips-frontend/app/chart/health-data.ts
```diff
@@ -0,0 +1,10 @@
+export interface HealthCareData {
+  areaCode: string;
+  healthData: {
```
williamwillis11, line 3: could this use the HealthDataPoint exported interface in frontend/fingertips-frontend/generated-sources/ft-api-client/models/HealthDataPoint.ts ? or should we not use those files as they are auto generated @alibombaye ?

frontend/fingertips-frontend/components/organisms/LineChart/lineChart.test.tsx
```diff
@@ -1,21 +1,54 @@
...
+      {
+        year: 2023,
+        count: 222,
+        value: 506.60912,
+        lowerCi: 441.69151,
+        upperCi: 578.32766,
+      },
+    ],
   },
 ];
```
williamwillis11, line 54: is it worth adding a snapshot test into this file as well as the existing one?

frontend/fingertips-frontend/app/chart/health-data.ts
alibombaye: That is an interesting question. Eventhough the type does match what we want I think it would be better to leave this as is and decouple this from the type that the api generates. 

One to discuss when @gareth-allan-bjss is back.

frontend/fingertips-frontend/app/chart/page.tsx
```diff
@@ -1,15 +1,29 @@
...
+  const searchParams = await props.searchParams;
+  const indicator = searchParams?.indicator ?? '';
+  const areaCode = searchParams?.areaCode ?? '';
+
   const config = getApiConfiguration();
-  const forecastApi = new WeatherForecastApi(config);
-  const data = await forecastApi.getWeatherForecast();
+  const indicatorApi = new IndicatorsApi(config);
+  const data = await indicatorApi.getHealthDataForAnIndicator({
+    indicatorId: Number(indicator),
```
alibombaye, line 24: Another one to discuss when @gareth-allan-bjss and @steves-bjss are back. Why is the indicatorId of type number in the api? 

If we are not going to do any calculations with this then the general consensus is for ids to be strings.

frontend/fingertips-frontend/components/organisms/LineChart/lineChart.test.tsx
priya-thorpe-bjss: The consensus was to not add one as the snapshot test kept failing because Highcharts dynamically generates ids. So on every run of the test the ids change and the snapshots never match.

frontend/fingertips-frontend/components/organisms/LineChart/index.tsx
```diff
@@ -17,28 +17,24 @@ export function LineChart({
...
-  const temperatureData = data.map((item) => item.temperatureC ?? '');
+  const years = data.map((item) => item.healthData.map((point) => point.year));
+  const formatYear = years.flat().map((year) => year.toString());
 
   const lineChartOptions: Highcharts.Options = {
     chart: { type: 'line' },
     title: { text: title },
     xAxis: {
-      categories: categories,
+      categories: Array.from(new Set(formatYear)),
```
alibombaye, line 27: How about a little helper function for this? something like `generateYearsForXAxis` where you do this and lines 20 and 21.

frontend/fingertips-frontend/components/organisms/LineChart/lineChart.test.tsx
```diff
@@ -1,21 +1,54 @@
...
+        upperCi: 578.32766,
+      },
+    ],
   },
   {
-    date: new Date('2024-11-01T00:00:00.000Z'),
-    temperatureC: 0,
-    temperatureF: 32,
-    summary: 'Bracing',
+    areaCode: 'A1426',
```
alibombaye, line 35: I know this is just mock data but is this accurate in terms of the areaCode being the same as the first element. Perhaps it would be better to use a different area code here.

frontend/fingertips-frontend/components/organisms/LineChartTable/index.tsx
```diff
@@ -20,16 +20,22 @@ export function LineChartTable({ data, headings }: Readonly<TableProps>) {
...
-          </Table.Row>
-        ))}
+        {data.map((item) =>
+          item.healthData.map((point, index) => (
+            <Table.Row key={`${item.areaCode}-${point.year}--${index}`}>
+              {index === 0 && (
+                <Table.Cell rowSpan={item.healthData.length}>
+                  {item.areaCode}
+                </Table.Cell>
+              )}
```
alibombaye, line 30: I think one to discuss with the BAs/POs about how we want to show the table when there are multiple areaCodes. From what I remember there was some different designs as to what that looks like. Probably worth doing in a different ticket though if its vastly different.

frontend/fingertips-frontend/components/pages/chart/chart.test.tsx
```diff
@@ -1,21 +1,54 @@
...
+        upperCi: 578.32766,
+      },
+    ],
   },
   {
-    date: new Date('2024-11-01T00:00:00.000Z'),
-    temperatureC: 0,
-    temperatureF: 32,
-    summary: 'Bracing',
+    areaCode: 'A1426',
```
alibombaye, line 35: Same here. Duplicated.

frontend/fingertips-frontend/lib/getApiConfiguration.ts
```diff
@@ -1,4 +1,4 @@
-import { Configuration } from '@/generated-sources/api-client';
+import { Configuration } from '@/generated-sources/ft-api-client';
```
alibombaye, line 1: Does doing this break the home-page?

priya-thorpe-bjss: The homepage doesn't use this function.

frontend/fingertips-frontend/components/pages/chart/helper.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export function generateSeriesData(
+  data: HealthCareData[]
+): Highcharts.SeriesLineOptions[] {
+  return data.map((item) => ({
+    type: 'line',
+    name: `AreaCode ${item.areaCode}`,
+    data: item.healthData.map((point) => point.value),
+  }));
+}
```
alibombaye, line 22: Nice 👍  - can you add some unit tests for these please.

frontend/fingertips-frontend/app/chart/health-data.ts
gareth-allan-bjss: Hmm, interesting. Skimming over this PR, this type _is_ only used to represent the data that comes back from the API, but it's plumbed quite far down into the components. I had initially assumed we'd transform it into our own specific type a bit higher up and plug that into our components, but I'm not sure there's really any benefit to that above what Priya's done here. I think if we have a type that represents what the API returns (which is what this is), then we should use the generated code for it.

Maybe this is worth a quick in-person conversation though, as it seems like something that could go back and forth a lot in a PR comment thread.

gareth-allan-bjss (commented):
Some early comments. I've got a little bit more to look through.

frontend/fingertips-frontend/lib/chartHelpers/formatLineChartValues.ts
```diff
@@ -0,0 +1,19 @@
+import { HealthCareData } from '@/app/chart/health-data';
+import Highcharts from 'highcharts';
+
+export function orderedValues(data: HealthCareData[]): HealthCareData[] {
+  return data.map((item) => ({
+    ...item,
+    healthData: item.healthData.toSorted((a, b) => a.year - b.year),
+  }));
+}
```
gareth-allan-bjss, line 9: I think this function could have a better name, to make it clearer what it's doing. Something like `sortHealthDataByDate` perhaps?

frontend/fingertips-frontend/lib/chartHelpers/formatLineChartValues.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export function generateSeriesData(
+  data: HealthCareData[]
+): Highcharts.SeriesLineOptions[] {
+  return data.map((item) => ({
+    type: 'line',
+    name: `AreaCode ${item.areaCode}`,
```
gareth-allan-bjss, line 16: Ah. Our designs want us to (not unreasonably) show the name of the area rather than its area code. Which is going to need another API call to an endpoint that we don't have yet to give us details of the areas.

I think this is something for us to flag to Emma B, as she was going to be putting together follow-up stories for these data visualisations, so we can make sure to include displaying the area names on one of those.

frontend/fingertips-frontend/components/organisms/LineChart/lineChart.test.tsx
```diff
@@ -1,26 +1,11 @@
...
-  {
-    date: new Date('2024-11-01T00:00:00.000Z'),
-    temperatureC: 0,
-    temperatureF: 32,
-    summary: 'Bracing',
-  },
-];
+import { mockHealthData } from '@/mock/data/healthdata';
 
 test('should render the Highcharts react component', () => {
```
williamwillis11, line 7: can we make this test name unique from the one being added in the bar chart PR? maybe add linechart into the test name

frontend/fingertips-frontend/components/pages/chart/chart.test.tsx
```diff
@@ -1,39 +1,24 @@
...
   const lineChart = screen.getByTestId('lineChart-component');
   expect(lineChart).toBeInTheDocument();
 });
 
 test('should render the LineChart component title', () => {
-  render(registryWrapper(<Chart data={mockData} />));
+  render(registryWrapper(<Chart data={mockHealthData} />));
 
   const lineChartTitle = screen.getByText('Line Chart');
   expect(lineChartTitle).toBeInTheDocument();
```
williamwillis11, line 17: does it make sense to check it has the H1 tag?

frontend/fingertips-frontend/components/pages/chart/index.tsx
```diff
@@ -1,26 +1,28 @@
...
+  indicatorsSelected?: string[];
 };
 
-const headings = ['Date', 'TemperatureC', 'TemperatureF', 'Summary'];
+const headings = ['Area Code', 'Year', 'Value', 'Count', 'LowerCi', 'UpperCi'];
 
 export function Chart({ data }: Readonly<ChartProps>) {
   return (
     <>
       <H1>Line Chart</H1>
```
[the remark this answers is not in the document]
priya-thorpe-bjss: I have requested access to the UR design file. For now I will change the Line Chart text title to the one you suggested.

frontend/fingertips-frontend/components/organisms/LineChart/index.tsx
```diff
@@ -17,28 +21,22 @@ export function LineChart({
...
   accessibilityLabel,
 }: Readonly<LineChartProps>) {
-  const categories = data.map(
-    (item) => item.date?.toLocaleDateString('en-GB') ?? ''
-  );
-  const temperatureData = data.map((item) => item.temperatureC ?? '');
+  const orderedSeriesValues = orderedValues(data);
+  const seriesData = generateSeriesData(orderedSeriesValues);
 
   const lineChartOptions: Highcharts.Options = {
```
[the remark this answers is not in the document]
priya-thorpe-bjss: Good spot on the designs.
I have updated the chart and disabled the ticks, I have spoken to Emma B about the line colours and styles and they will be in another ticket.

frontend/fingertips-frontend/components/organisms/LineChart/index.tsx
```diff
@@ -1,44 +1,50 @@
...
   accessibilityLabel,
 }: Readonly<LineChartProps>) {
-  const categories = data.map(
-    (item) => item.date?.toLocaleDateString('en-GB') ?? ''
-  );
-  const temperatureData = data.map((item) => item.temperatureC ?? '');
+  const sortedSeriesValues = sortHealthDataByDate(data);
+  const seriesData = generateSeriesData(sortedSeriesValues);
 
   const lineChartOptions: Highcharts.Options = {
```
gareth-allan-bjss, line 24: I think if you add the following to the option object it'll take the "highcharts.com" watermark off the chart:
```suggestion
  const lineChartOptions: Highcharts.Options = {
    credits: {
      enabled: false
    },
```

frontend/fingertips-frontend/components/organisms/LineChartTable/index.tsx
gareth-allan-bjss: [DHSCFT-31](https://bjss-enterprise.atlassian.net/browse/DHSCFT-31) will cover displaying the table alongside the line chart, so happy to leave it as it is in this PR.

frontend/fingertips-frontend/lib/chartHelpers/formatLineChartValues.ts
gareth-allan-bjss: I've raised [DHSCFT-193](https://bjss-enterprise.atlassian.net/browse/DHSCFT-193) to ensure we display area names once the API endpoint is available.

gareth-allan-bjss (approved):

PR #40, opened 2025-01-06
DHSCFT-64: scoring setup tests

ataylor-bjss:
# Description

Jira ticket: [DHSCFT-64](https://bjss-enterprise.atlassian.net/browse/DHSCFT-64)

Adds testing for creation of scoring profile.

## Changes

- Adds testing for creation of scoring profile.

## Validation

Have run the tests and they pass. Have also done some manual testing to check that the scoring profile has been created and that it seems to behave as expected; this manual testing was done as part of https://github.com/dhsc-govuk/FingertipsNext/pull/29 since this is the ticket that covers the actual creation of the profile.
Screenshot 2025-01-07 at 13 14 46
scoring profile added to indicator index
Screenshot 2025-01-08 at 14 51 24

The tests have received some updates, proof of latest tests:


search-setup/indexOperations.ts
```diff
@@ -4,8 +4,9 @@ import {
   SearchField,
   SearchClient,
   SearchFieldDataType,
+  ScoringProfile,
 } from "@azure/search-documents";
-import { Data } from "./types.js";
+import { Data, ScoringWeight } from "./types.js";
```
williamwillis11, line 9: i can see it wasn't your change but wondering if we need the .js on the end of the import? looking at other imports in this project we don't usually specify the filetype of the import

williamwillis11: seen you have made this change - thanks!

williamwillis11:
making some additional enhancements to the test:

search-setup/__tests__/search.test.ts

gareth-allan-bjss:
Have you run the scripts against an AI search instance and validated that:

- The tests pass
- A scoring profile is created as expected on the index?

If so, can you add something to the PR's validation section to explain this please? I'm afraid we're at a point where our tests for this don't get run automatically, so we're reliant on manually validating and providing evidence in PRs.

search-setup/indexOperations.ts
```diff
@@ -61,3 +72,23 @@ function buildSearchIndexField(
...
+  weights: ScoringWeight[]
+): ScoringProfile {
+  let scoringProfile: ScoringProfile = 
+   {
+    name: name,
+    textWeights: {
+      weights: {
+      }
+    }
+  };
```
gareth-allan-bjss, line 87: Do we have the option to set this scoring profile as the default for the index? I think we'd want that, wouldn't we?

search-setup/indexOperations.ts
```diff
@@ -41,6 +42,16 @@ function buildSearchIndex(name: string): SearchIndex {
         ],
       },
     ],
+    scoringProfiles: [
+      buildScoringProfile(
+        "BasicScoringProfile",
```
adawari-thomas-bjss, line 47: could the scoring profile name be set as an environment variable to match the others?

search-setup/indexOperations.ts
gareth-allan-bjss: I think it makes sense. It at least means the search behaviour is consistent for this index, regardless of how you query it.

search-setup/indexOperations.ts
ataylor-bjss: Certainly that's possible but in my opinion that would be inconsistent with how we are creating the rest of the index. The scoring profile as I see it is a property of the index in the same way as the fields that are created & we don't pass the name of those in via an environment variable.
But happy to do it if we think it would be useful, who makes the call on this sort of thing?

adawari-thomas-bjss: Fair point, but I don't think the scoring profile name is immediately provided, unlike the names of the other fields in an index. A similar thing was done for the suggester on the geography index. What are your thoughts on this @gareth-allan-bjss?

search-setup/integration_tests/search.test.ts
```diff
@@ -0,0 +1,156 @@
...
+      const documents: DocumentResponse = await response.json();
+
+      expect(documents.value).toBeDefined();
+      expect(documents.value.length).toBeGreaterThan(0);
+
+      const firstDocument = documents.value[0];
+      expect(firstDocument).toHaveProperty("IID");
+      expect(firstDocument).toHaveProperty("Descriptive");
+      expect(firstDocument.Descriptive).toHaveProperty("Name");
+      expect(firstDocument.Descriptive).toHaveProperty("Definition");
```
gareth-allan-bjss, line 153: What's the thinking behind these additional checks @williamwillis11? Does it add any extra validation over the tests in the other describe block that validate the index configuration?

My concern is that we've now got two places to change if we rename a field (for example).

search-setup/integration_tests/search.test.ts
```diff
@@ -0,0 +1,156 @@
...
+          { IID: 20 },
+          { "Descriptive/Name": 10 },
+          { "Descriptive/Definition": 5 },
+        ],
+      });
+    });
+  });
+
+  describe("Index data from docs url", () => {
+    test("should populate data", async () => {
```
gareth-allan-bjss, line 137: You've changed all the other tests to use `it(...)` so should probably do this one too:
```suggestion
    it("should populate data", async () => {
```

search-setup/indexOperations.ts
gareth-allan-bjss: I think you both have valid points. If we're having to pass the scoring profile in the search query, specifying it in an environment variable would allow it to be passed into this application as well as the frontend, so they don't have to both contain implicit knowledge of the scoring profile name. Having said that, @ataylor-bjss is right that we don't do this for the entire index structure (it would get complicated fast), so there's already implicit knowledge.

I think perhaps the simplest thing would be to make this scoring profile the default and remove it from the parameters passed to the search query. Then we don't care what the profile name is, and can leave it as a hardcoded value.

search-setup/integration_tests/search.test.ts
williamwillis11: good spot - in another PR i have changed other test to it after Fawad made that start for clarity/consistency

search-setup/integration_tests/search.test.ts
williamwillis11: agreed its not adding anything extra just creating future maintenance - ill remove these

search-setup/indexOperations.ts
```diff
@@ -84,3 +95,23 @@ function buildSearchIndexField(
...
+   {
+    name: name,
+    textWeights: {
+      weights: {
+      }
+    }
+  };
+
+  for (const weighting of weights) {
+    scoringProfile.textWeights!.weights = { ...scoringProfile.textWeights!.weights, ...weighting };
```
alibombaye, line 113: What would happen if `textWeights` are not provided in scoringProfile? In general try and avoid using `!` where possible

search-setup/integration_tests/search.test.ts
```diff
@@ -0,0 +1,156 @@
...
+import {
+  DocumentResponse,
+  SearchIndexResponse,
+  IndexField,
+  ScoringProfile,
+  ScoringWeight,
+} from "../types";
+import { getEnvironmentVariable } from "../utils/helpers";
+
+config();
```
alibombaye, line 11: Can this go into `getEnvironmentVariable` function? I assume this is needed for that purpose.

williamwillis11: yep done this thanks

search-setup/indexOperations.ts
```diff
@@ -4,8 +4,13 @@ import {
   SearchField,
   SearchClient,
   SearchFieldDataType,
+  ScoringProfile,
```
gareth-allan-bjss, line 7: This `ScoringProfile` import is now unused.
```suggestion
```

search-setup/indexOperations.ts
```diff
@@ -4,8 +4,13 @@ import {
...
 } from "@azure/search-documents";
-import { GeographySearchData, IndicatorSearchData } from "./types.js";
+import {
+  ScoringWeight,
```
gareth-allan-bjss, line 10: `ScoringWeight` isn't used in this file.
```suggestion
```

search-setup/integration_tests/search.test.ts
```diff
@@ -0,0 +1,264 @@
...
+      });
+      expectFieldToMatch(index.fields[3], {
+        name: "Postcode",
+        type: "Edm.String",
+        retrievable: true,
+        searchable: true,
+        sortable: true,
+        filterable: true,
+      });
+    });
```
gareth-allan-bjss, line 236: `Name` and `Type` aren't address fields, they're other details of the geography. I think you need to either split these each into their own test, or have a single "should have the correct field configurations" test.

gareth-allan-bjss (approved):
A few minor comments, but I'm happy that if they're fixed (and the latest version of the integration tests have been run successfully against search indexes configured with this tool!) that this can be merged.

search-setup/indexOperations.ts
ataylor-bjss: This function has been removed so its use of ! is no longer an issue.

PR #45, opened 2025-01-13
DHSCFT-180: Added geo data for PoC and various updates

HowardMayBjss:
# Description
This adds the data required for PoC and makes various updates to properly use
Jira ticket: [DHSCFT-180](https://bjss-enterprise.atlassian.net/jira/software/projects/DHSCFT/boards/823?selectedIssue=DHSCFT-180)

## Changes

- Adds PoC geo data
- Changed 'key' field from ID field to area-code as this should be unique and avoids the need to manage a separate set of numbers
- Refactored project structure to keep it cleaner
- Renamed AI Search fields to area-type, area-code and area-name
- Refactored some env vars which don't need to change between environments. Made them constants
- Updated tests
- Added a postman collection for using suggestions api

## Validation

Integration tests continue to pass. Additional manual testing will be performed and added to the ticket.


search-setup/src/constants.ts
```diff
@@ -0,0 +1,3 @@
+export const indicatorSearchIndexName = "indicator-search-index";
+export const geographySearchIndexName = "geography-search-index";
+export const geographySearchSuggesterName = "geographySuggester";
```
alibombaye, line 3: If these are constants should these be capitalised? 

How about an enum instead though. Might be a nice way too?

search-setup/src/indexOperations.ts
```diff
@@ -65,19 +67,16 @@ export function buildGeographySearchIndex(name: string): SearchIndex {
...
     ],
     suggesters: [
       {
-        name: getEnvironmentVariable(
-          "AI_SEARCH_BY_GEOGRAPHY_INDEX_SUGGESTER_NAME"
-        ),
+        name: geographySearchSuggesterName,
         searchMode: "analyzingInfixMatching",
-        sourceFields: ["Name", "Postcode"],
+        sourceFields: ["areaName", "areaCode"],
```
alibombaye, line 79: Do you think enums/const would be useful for these sourceFields?

alibombaye (approved):

search-setup/assets/geographyData.json
gareth-allan-bjss, line 1: I happened to spot that the last area in this file (area code `Y07275`) has a newline at the end of its `areaType` value that probably shouldn't be there.

search-setup/src/main.ts
```diff
@@ -44,18 +44,22 @@ async function main(): Promise<void> {
...
     indexClient,
     buildIndicatorSearchIndex,
-    indicatorSearchIndexName,
+    INDICATOR_SEARCH_INDEX_NAME,
     sampleIndicatorData
   );
 
+  const geoData = JSON.parse(
+    fs.readFileSync("./assets/geographyData.json", "utf-8")
+  );
```
gareth-allan-bjss, line 53: I think instead of loading the JSON file this way you could just use an import statement, couldn't you:
```typescript
import geoData from "../assets/geographyData.json";
```

We already have the `resolveJsonModule` property set in this application's `tsconfig.json` file, and the benefit is that it'll check at compile time that the JSON file matches your `GeographySearchData` type.

search-setup/src/indexOperations.ts
```diff
@@ -5,16 +5,21 @@ import {
...
+} from "./constants.js";
 
 export async function createIndex(
   indexClient: SearchIndexClient,
   index: SearchIndex
 ): Promise<void> {
+  try {
+    console.log("Trying to delete an existing index");
+    await indexClient.deleteIndex(index);
+  } catch {}
```
gareth-allan-bjss, line 21: Why do we need to delete the index first? The idea behind using `createOrUpdateIndex` was that it would just update any pre-existing index if it were present.

search-setup/integration_tests/search.test.ts
```diff
@@ -8,7 +8,13 @@ import {
...
+import { getEnvironmentVariable } from "../src/utils/helpers";
+import {
+  GEOGRAPHY_SEARCH_INDEX_NAME,
+  GEOGRAPHY_SEARCH_SUGGESTER_NAME,
+  GeographySearchIndexColumnNames,
+  INDICATOR_SEARCH_INDEX_NAME,
+} from "../src/constants";
 
 let index: SearchIndexResponse;
 let searchEndpoint: string;
```
gareth-allan-bjss, line 20: Very minor point, but my IDE is pointing out that the `indexName` variable declared on the line below this one (line 21) is no longer used (but I can't comment on it because it's not been changed).

gareth-allan-bjss (commented):
Looks good, just a handful of minor suggestions.

search-setup/src/main.ts
```diff
@@ -21,6 +26,7 @@ async function createAndPopulateIndex<
   indexName: string,
   data: T[]
 ) {
+  console.log(`CreatingIndex ${indexName}`);
```
gareth-allan-bjss, line 29: Typo?
```suggestion
  console.log(`Creating index ${indexName}`);
```

search-setup/src/indexOperations.ts
HowardMayBjss: This updates at the index document level (a document is an individual area in this context). I did a test where I had 4 areas and then updated with only 3 areas. Rather than being left with 3 areas, you actually have 4. Because of this I have deleted the entire index and recreated it. This is still very quick - of the order of a second to achieve - and it gives us confidence that the index is true to the configuration. Once we move to configuring from the database then the whole approach will change and this code will likely be redundant.

search-setup/assets/fingertips-geo-suggest.postman_collection.json
```diff
@@ -0,0 +1,371 @@
...
+					"mode": "raw",
+					"raw": "    {  \n      \"search\": \"manc\",  \n      \"top\": 20,\n      \"searchFields\": \"areaName,areaCode\",\n      \"suggesterName\": \"geographySuggester\",\n      \"fuzzy\": true\n    }  ",
+					"options": {
+						"raw": {
+							"language": "json"
+						}
+					}
+				},
+				"url": {
+					"raw": "https://fingertips-ai-search.search.windows.net/indexes/geography-search-index/docs/suggest?api-version=2020-06-30",
```
gareth-allan-bjss, line 51: You've got your AI search instance's URL in here. I know it's not exactly a secret, but you probably don't want to default to pointing everyone at (what I assume is) an instance in your own Azure account.

search-setup/src/indexOperations.ts
gareth-allan-bjss: Ah, OK, I hadn't necessarily considered that scenario. Happy to leave this in then.

gareth-allan-bjss (approved):
Looks good, I'm happy for this to go in once you've addressed the SonarQube issue relating to an unused import.

PR #46, opened 2025-01-13
Feature/ftn 1234 geography api design

andyevans-bjss:
# Description

Jira ticket: Don't have a JIRA ticket for this
Adding area / geography API endpoints to the design first OpenAPI yaml file

## Changes

- Add 4 new area / geography related endpoints to the OpenAPI definition

## Validation

Reviewed with Fawad, Steve, Gareth and Howard


api/definition/swagger.yaml
```diff
@@ -12,7 +12,87 @@ info:
...
+      tags:
+        - areas
+      summary: Get area types
+      description: Get area types, optionally filtering by hierarchy type
+      operationId: getAreaTypes
+      parameters:
+        - $ref: "#/components/parameters/hierarchy_type"
+      responses:
+        '200':
+          description: The available area types e.g. CCB, PCN or GP Surgery
```
HowardMayBjss, line 47: Should than be ICB not CCB?

HowardMayBjss: The response is only a string, is there an areaType structure which can be returned which includes name, hierachy, description?

api/definition/swagger.yaml
```diff
@@ -88,6 +168,73 @@ paths:
...
+          type: string
+          example: "England"
+          description: The name of the area / geography
+    Area:
+      type: object
+      description: A area or geography that has associated public health data
+      required:
+        - code
+        - name
+        - hierarchyType
```
HowardMayBjss, line 192: I think this is quite subjective, but I feel this should be hierarchyName. For area we have areaName to identify different areas and then in addition to that areaType to differentiate area's at a different level. I see NHS and Administration as different instances of geographic heirachies - same type, different name.

steves-bjss (approved):

api/definition/swagger.yaml
```diff
@@ -190,6 +337,40 @@ components:
           type: integer
           example: 2023
         maxItems: 10
+    area_code:
```
HowardMayBjss, line 340: underscore has crept in here - looks inconsistent.

api/definition/swagger.yaml
andyevans-bjss: will change to ICB in description

api/definition/swagger.yaml
andyevans-bjss: will change

api/definition/swagger.yaml
andyevans-bjss: area_code is used throughout

PR #49, opened 2025-01-14
Feature/dhscft 54 correlation

Joe-Hendley:
# Description

Jira ticket: [DHSCFT-54](https://bjss-enterprise.atlassian.net/browse/DHSCFT-54)

Adds basic OpenTelemetry tracing to the frontend and backend components.

The details of how this works can be found [here](https://opentelemetry.io/docs/concepts/signals/traces/) but the gist is that each request has a corresponding 'Trace' with a unique `trace_id`, and each unit of work within that trace is called a 'Span' with a unique `span_id`.

At this stage our request-specific correlation ID is the `trace_id` and we don't need to worry about the rest. OpenTelemetry attaches trace information to requests using language-specific SDKs, propagates trace information automatically, and can be set up in a basic way very easily as per this PR.

The trace ID can be accessed easily within the code as per the examples in the weather handlers. My view is that we could add this as a wrapper to our loggers.

I've kept the console exporters commented out for this PR. It might be worth having a quick play around and viewing the output before deciding whether to have them enabled for development or not. I'd be inclined to not - they're very verbose by default, especially the front-end putting out 10 logs per request.

I think this will immediately lead to 2 larger pieces of work, integrating & configuring then sending what we need to Application Insights; each for the frontend and the api.

image
image


frontend/fingertips-frontend/instrumentation.ts
```diff
@@ -29,7 +29,40 @@ const configureApplicationInsights = async () => {
...
+    const {
+      SimpleSpanProcessor,
+      ConsoleSpanExporter,
+    } = require('@opentelemetry/sdk-trace-node');
+    const { Resource } = require('@opentelemetry/resources');
+    const {
+      ATTR_SERVICE_NAME,
+    } = require('@opentelemetry/semantic-conventions');
+
+    // const spanProcessor = new SimpleSpanProcessor(new ConsoleSpanExporter())
```
williamwillis11, line 47: do we want to commit this commented code?

Joe-Hendley: explained in description (tl;dr probably not, but staying until reviewed by steve / gareth)

PR #52, opened 2025-01-15
make some minor changes to the area API endpoints in the swagger

andyevans-bjss:
# Description

Simplify the areas API endpoints, adding suport for getting children at different levels

## Changes

- Remove the ability to get cousins and siblings
- Add the ability to get child areas at lower levels (e.g. grandchildren)

### DIFF
--- a/api/definition/swagger.yaml
+++ b/api/definition/swagger.yaml
@@ -58,14 +58,13 @@ paths:
       tags:
         - areas
       summary: Get area
-      description: Get the full details of a given area, including it's parent, optionally including it's children, siblings, cousins and ancestors
+      description: Get the full details of a given area, including it's parent, optionally including it's children and ancestors
       operationId: getArea
       parameters:
         - $ref: "#/components/parameters/area_code"
         - $ref: "#/components/parameters/includeChildren"
-        - $ref: "#/components/parameters/includeSiblings"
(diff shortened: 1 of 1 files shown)


api/definition/swagger.yaml
```diff
@@ -223,14 +222,6 @@ components:
               type: array
               items:
                 $ref: "#/components/schemas/Area"
-            siblings:
-              type: array
-              items:
-                $ref: "#/components/schemas/Area"
```
gareth-allan-bjss, line 229: As discussed offline, I think we still need the option to return siblings to allow us to show all areas in the same group as the one selected.

api/definition/swagger.yaml
```diff
@@ -347,28 +338,25 @@ components:
...
+        type: string
+        example: PCN
+      description: The type of area to request children for
+    include_children:
       in: query
-      name: includeCousins
-      description: include the cousin areas
-      schema:
+      name: include_children
+      description: Optionally, include the child areas. By default this is the direct children, to get children at a lower level supply the optional query paramter for child area type.
```
MonkeyDate, line 353: typo 'paramter'

williamwillis11:
agreed with @andyevans-bjss that this has been superseded by https://github.com/dhsc-govuk/FingertipsNext/pull/88

PR #58, opened 2025-01-17
[DHSCFT-TECH]: Clear the Areas.Areas table before populating it

gareth-allan-bjss:
# Description

Jira ticket: None.

There is an issue with the database population script that means it fails when run a second time against an already-populated DB. This is due to the fact that we are not clearing the `Areas.Areas` table before population and are using fixed values for the `Node` primary key field in our population script.

## Changes

- Add a `DELETE FROM` statement to `PopulateFingertips.sql` to clear the `Areas.Areas` table before populating it

## Validation

Ran the `docker compose --profile all up --build --remove-orphans -d` command twice on `main`:
Screenshot 2025-01-17 at 10 27 57
Screenshot 2025-01-17 at 10 29 37

The db-setup container logs contained the following:
```
An error occurred while the batch was being executed.
Updating database (Failed)
*** Could not deploy package.
Error SQL72014: Core Microsoft SqlClient Data Provider: Msg 2627, Level 14, State 1, Line 2 Violation of PRIMARY KEY constraint 'PK__Areas__7D8CACC0DB70DBE7'. Cannot insert duplicate key in object 'Areas.Areas'. The duplicate key value is (0x).
```

Then ran `docker compose --profile all down` to tear down all containers, checked out this branch and ran `docker compose --profile all up --build --remove-orphans -d` twice:
Screenshot 2025-01-17 at 10 56 37
Screenshot 2025-01-17 at 10 57 02

Finally, checked the `Areas.Areas` table still contains the expected data:
Screenshot 2025-01-17 at 10 58 26

### DIFF
--- a/database/fingertips-db/PostDeployment/PopulateFingertips.sql
+++ b/database/fingertips-db/PostDeployment/PopulateFingertips.sql
@@ -3005,10 +3005,13 @@ INSERT [dbo].[HealthMeasure] ([HealthMeasureKey], [AreaKey], [IndicatorKey], [Se
 SET IDENTITY_INSERT [dbo].[HealthMeasure] OFF
 GO
 
-INSERT [Areas].[Areas]   
+--delete all existing data so we always start from a known position
+DELETE FROM [Areas].[Areas]
+
+INSERT [Areas].[Areas]
 VALUES
 -- root level data
-('/', 0,'E92000001','England', 'Country','All') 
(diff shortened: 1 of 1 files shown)


alibombaye (approved):

HowardMayBjss (approved):

HowardMayBjss:
I downloaded the code and successfully started my containers. Thank you

PR #59, opened 2025-01-17
Feature/dhscft 66 search by indicator

HowardMayBjss:
# Description

Jira ticket: [Search by Indicator Id](https://bjss-enterprise.atlassian.net/browse/DHSCFT-66)

This PR builds on https://github.com/dhsc-govuk/FingertipsNext/pull/29

Key updates from the previous PR include

- Adding env vars to Docker file and ci.yml to support build and test
- Completed addition of dataSource and lastUpdated fields
- Aligned structure of Indicator document structure with indicator results type
- Updating unit tests and e2e tests
- Merging from main

There is some additional alignment with geography search e.g. using enum for field names but this will be done later. 

## Validation

No new validation has been added beyond updating existing tests


.github/workflows/ci.yml
```diff
@@ -4,6 +4,10 @@ env:
   dotnet-version: 9.0.x
   frontend-directory: ./frontend/fingertips-frontend
   NEXT_TELEMETRY_DISABLED: true
+  DHSC_AI_SEARCH_SERVICE_URL: "https://dummyurl.search.windows.net"
```
HowardMayBjss, line 7: At present the code needs these to all be defined even if ai_search isn't being used

frontend/fingertips-frontend/Dockerfile
```diff
@@ -28,7 +28,10 @@ COPY . .
 # Learn more here: https://nextjs.org/telemetry
 # Uncomment the following line in case you want to disable telemetry during the build.
 ENV NEXT_TELEMETRY_DISABLED=1
-
+ARG DHSC_AI_SEARCH_SERVICE_URL="https://dummyurl.search.windows.net"
```
HowardMayBjss, line 31: At present these are all needed at build time - this may be because of the use of instrumentation.ts

.github/workflows/ci.yml
ataylor-bjss: Is this correct? If the environment variable is set to use the mock service then environmentContext.ts shouldn't throw an error if it can't load the required search service environment variables.

search-setup/src/indexOperations.ts
```diff
@@ -123,3 +142,24 @@ function buildSearchIndexField(
     hidden,
   };
 }
+
+function buildScoringProfile(
```
ataylor-bjss, line 146: Gareth put in a comment about removing this function and just replacing its use with the JSON it would generate. That change was made but seems to have been reverted somehow.

HowardMayBjss: I don't see that change on FTN-58_search_by_indicator but will add it in to this PR

frontend/fingertips-frontend/lib/utils.ts
```diff
@@ -0,0 +1,26 @@
...
+// See https://github.com/emotion-js/emotion/blob/3c19ce5997f73960679e546af47801205631dfde/packages/is-prop-valid/src/props.js
+// for examples of valid props for HTML and SVG elements
+export const shouldForwardProp = (
+  propName: string,
+  target: unknown
+): boolean => {
+  return typeof target === 'string' ? isPropValid(propName) : true;
+};
+
+export const isBrowser = () => typeof window !== 'undefined';
```
gareth-allan-bjss, line 12: The other utilities in this file have been removed, so this appears to be a merge conflict resulting in them coming back.

search-setup/src/indexOperations.ts
gareth-allan-bjss: It was in a prior PR (#40). I think all we need to change in this file for the scoring profile is the name of the fields it references.

search-setup/integration_tests/search.test.ts
gareth-allan-bjss, line 1: You've added fields to the index. I think we should add test cases to validate they have been configured as expected here.

search-setup/src/indexOperations.ts
```diff
@@ -4,13 +4,18 @@ import {
...
+import { GeographySearchData, IndicatorSearchData } from "./types";
 import {
   GEOGRAPHY_SEARCH_SUGGESTER_NAME,
   GeographySearchIndexColumnNames,
 } from "./constants.js";
 
+interface ScoringWeight {
+  [fieldName: string]: number;
+}
+
```
gareth-allan-bjss, line 18: Sorry, missed this. It's not used and seems to have snuck back in as part of merging.

frontend/fingertips-frontend/Dockerfile
gareth-allan-bjss: I'm happy with changing the value in this file to `DHSC_AI_SEARCH_USE_MOCK_SERVICE=true`, but I'm leaving this thread to track the following places we need to add this value as well (as discussed offline):
- The `frontend/fingertips-frontend/.env.development` file (to enable AI search mocks when running the Next development server with `npm run dev` from the frontend dir
- The `environment` section of the `frontend` service in the `compose.yaml` (to enable AI search mocks when running the docker compose command
- The `environmentVariables` element of the `Build and deploy Container App` step in `.github/workflows/fingertips-frontend-deploy.yml` (once you've merged main, to ensure we're using AI search mocks when running the FE in a deployed environment) - @epercy-bjss: For reference, we're assuming we'll have the FE running against AI search mocks until you're ready to wire it up

HowardMayBjss: On review, only DHSC_AI_SEARCH_USE_MOCK_SERVICE=true is currently needed and the other vars have been removed. The various places where DHSC_AI_SEARCH_USE_MOCK_SERVICE is needed has been capture by Gareth and all updated.

search-setup/integration_tests/search.test.ts
HowardMayBjss: I've moved to using a snapshot so it captures more of the configuration and requires less maintenance

search-setup/src/indexOperations.ts
HowardMayBjss: good spot

README.md
```diff
@@ -37,6 +37,27 @@ Finally, you can stop all of the running containers with the following command,
...
+
+- fingertips-frontend/.env.development so that it is used when the nextjs app is run in dev mode
+- compose.yaml so that it is set when the frontend container is started
+- ci.yml so that it is set when the frontend is run in CI
+- fingertips-frontend-deploy.yml as a temporary measure until the Azure AI search is configured
+- fingertips-frontend/Dockerfile as it is currently needed at build time (this may be refactored away)
+
+To enable the system to connect to the Azure AI Search the following environment variables will need to be configured.
+
+- DHSC_AI_SEARCH_SERVICE_URL=< the url the Azure AI Search instance is exposed on >
```
gareth-allan-bjss, line 57: A minor point, but it might be worth giving an example of a more realistic URL format, as when setting this up for the search-setup app I had to double-check whether we needed to include the `https://` prefix, etc.

HowardMayBjss: I'll follow that up on my next branch. Good point though.

PR #69, opened 2025-01-20
Create GitHub Actions Workflow for API Build and Deploy

epercy-bjss:
# Description

Jira ticket: [DHSCFT-190](https://bjss-enterprise.atlassian.net/browse/DHSCFT-190)

Create a new GitHub workflow to build, test, and deploy the API component from the FingertipsNext repo.

## Changes

- Create a build workflow to build and test the API project
- Create a deploy workflow to deploy the service to Azure Container Apps, after a successful build to the `main` branch (using workflow_run trigger)
- Refactor workflows to ensure path-based filtering, so CI is only triggered when changes are made to the relevant folders

## Validation

Merge into main required prior to testing

### DIFF
--- a/.github/workflows/fingertips-api-build.yml
+++ b/.github/workflows/fingertips-api-build.yml
@@ -10,9 +10,13 @@ on:
     branches:
       ## ** is required to match branches with / in their name
       - "**"
+    paths:
+      - api/**
   pull_request:
     branches:
       - main
+    paths:
+      - api/**
 
 jobs:
   determine-version:
--- a/.github/workflows/fingertips-api-build.yml
+++ b/.github/workflows/fingertips-api-build.yml
@@ -2,6 +2,7 @@ name: Fingertips API Build
 
 env:
   api-directory: ./api
+  dotnet-version: 9.0.x
 
 on:
   workflow_dispatch:
@@ -27,9 +28,6 @@ jobs:
   api-build:
     name: Build API
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        dotnet-version: ["9.0.x"]
 
     defaults:
       run:
@@ -39,10 +37,10 @@ jobs:
       - name: Checkout source
         uses: actions/checkout@v4
 
-      - name: Setup dotnet ${{ matrix.dotnet-version }}
+      - name: Setup dotnet ${{ env.dotnet-version }}
         uses: actions/setup-dotnet@v4
         with:
-          dotnet-version: ${{ matrix.dotnet-version }}
+          dotnet-version: ${{ env.dotnet-version }}
 
       - name: Restore Packages
         run: dotnet restore
--- a/.github/workflows/fingertips-api-build.yml
+++ b/.github/workflows/fingertips-api-build.yml
@@ -1,8 +1,7 @@
 name: Fingertips API Build
 
 env:
-  frontend-directory: ./api
-  NEXT_TELEMETRY_DISABLED: true
+  api-directory: ./api
 
 on:
   workflow_dispatch:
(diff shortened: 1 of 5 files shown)


.github/workflows/fingertips-frontend-build.yml
```diff
@@ -10,9 +10,13 @@ on:
     branches:
       ## ** is required to match branches with / in their name
       - "**"
+    paths:
+      - frontend/fingertips-frontend/**
```
williamwillis11, line 14: should this path go as a var at the top like we have in the ci.yml file for this now?

epercy-bjss: There are no path-based filters on the existing 'ci.yml' file:

```
on:
  push:
    branches:
      ## ** is required to match branches with / in their name
      - "**"
  pull_request:
    branches:
      - main
 ```
 
 You cannot use variables in the path filter unfortunately

Comment (epercy-bjss) [reply .github/workflows/fingertips-frontend-build.yml:14]: To explain further, this refernce is used for the workflow triggers, which directs the pipeline to run _only_ when changes are made to a file within the path specified in this property

https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/triggering-a-workflow#using-filters

File: .github/workflows/fingertips-api-build.yml [hunk 2]
```diff
@@ -0,0 +1,75 @@
+name: Fingertips API Build
+
+env:
+  frontend-directory: ./api
```
gareth-allan-bjss, line 4: This isn't the frontend directory.

.github/workflows/fingertips-api-build.yml
```diff
@@ -0,0 +1,75 @@
...
+  NEXT_TELEMETRY_DISABLED: true
```
gareth-allan-bjss, line 5: I don't think it'll cause any problems, but we're not building a Next app here, so `NEXT_TELEMETRY_DISABLED` won't be doing anything.

epercy-bjss:
@gareth-allan-bjss those comments should be fixed now. I've also added the e2e tests that Will put together as a subsequent job on the API deployment workflow

.github/workflows/fingertips-api-deploy.yml
```diff
@@ -0,0 +1,81 @@
...
+            DB_PASSWORD=${{ steps.azure-keyvault-secrets.outputs.fingertips-sql-password }}
+          azureCredentials: ${{ secrets.AZURE_CREDENTIALS }}
+        env:
+          AZURE_CONFIG_DIR: $HOME/.azure
+
+  e2e-tests:
+    uses: ./.github/workflows/e2e-tests.yml
+    needs: api-deploy
+    with:
+      env-url: http://4.158.39.145/search
```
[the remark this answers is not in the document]
williamwillis11: you need to remove *search* from the end of this url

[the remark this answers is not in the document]
epercy-bjss: There are some issues with using around using environment-specific variables at the moment with the e2e workflow. Right now, it's not using a 'deployment' environment (as it's used for build and deploy). We will probably need to refactor this workflow as 'composite' actions, rather than a reusable workflow. That way, it can integrate into the build job **and** the deployment workflows (and I can access the deployment environment variables for the latter).

Are we ok to accept this for now, so we can get this running and I'll refactor the workflows in another ticket?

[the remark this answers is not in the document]
epercy-bjss: Fixed the URL to remove the path

.github/workflows/fingertips-api-build.yml
```diff
@@ -0,0 +1,69 @@
+name: Fingertips API Build
+
+env:
+  api-directory: ./api
+  dotnet-version: 9.0.x
```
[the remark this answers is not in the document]
epercy-bjss: I misunderstood your comment on the 'search-setup' PR. I've added this back in now.

gareth-allan-bjss (approved):

PR #71, opened 2025-01-21
Feature/DHSCFT-208_configure_e2e_tests

williamwillis11:
# Description

Jira ticket: [DHSCFT-208](https://bjss-enterprise.atlassian.net/browse/DHSCFT-208)

Configure Playwright to run against either deployed fingertips instance in CI or localhost

## Changes

- added e2e_test template which takes an optional env-url parameter
- in the build job call the template above but pass nothing and playwright falls back to localhost
- in the deploy job (triggered on merges to main) call the template and pass the env-url and playwright runs against that
- have removed the e2e_test job from ci.yml as its duplication until that file gets removed

## Validation

Have ran against localhost in build job -https://github.com/dhsc-govuk/FingertipsNext/actions/runs/12889783924

Have not been able to test merge to main triggering the e2e tests on deploy as we need to merge a PR to main to trigger this.


.github/workflows/e2e-tests.yml
```diff
@@ -0,0 +1,57 @@
...
+jobs:
+  execute-e2e-tests:
+    name: Execute e2e tests
+    runs-on: ubuntu-latest
+    container:
+      image: mcr.microsoft.com/playwright:v1.49.1-jammy
+
+    strategy:
+      matrix:
+        node-version: ["22.x"]
```
epercy-bjss, line 25: There has been a suggestion in one of my previous PRs that this node-version matrix is not required. I will be removing it from all other workflows. Please remove the matrix and amend any references to it. I've pointed at an env var instead.

.github/workflows/fingertips-frontend-deploy.yml
```diff
@@ -22,7 +22,6 @@ jobs:
       matrix:
         environment:
           - development
-    environment: ${{ matrix.environment }}
```
epercy-bjss, line 25: Why is this being removed?

epercy-bjss (commented):
Just a couple of comments, otherwise LGTM

.github/workflows/fingertips-frontend-deploy.yml
williamwillis11: removed by accident - readded

epercy-bjss (approved):

PR #74, opened 2025-01-22
Feature/dhscft 241 delete redundant weather forecast code

priya-thorpe-bjss:
# Description

Jira ticket: [DHSCFT-241](https://bjss-enterprise.atlassian.net/browse/DHSCFT-241)

## Changes
- Removed redundant weather forecast related code

## Validation
- Pages fully render
- E2E tests pass


api/DHSC.FingertipsNext.Modules.Core/Module.cs
gareth-allan-bjss, line 1: This file is also registering the `SearchClient` and the `CoreService`, so if we're keeping those we should presumably keep this file and just remove the lines that register `CoreController` and `WeatherFactory`. I think it does bring into question whether there's any point keeping the entire core module though. I'm not sure it's doing anything now we've stripped out the weather forecast stuff (other than serving as a bit of an example of the modular monolith approach). What do you think @steves-bjss?

api/DHSC.FingertipsNext.Modules.Core/DHSC.FingertipsNext.Modules.Core.csproj
```diff
@@ -27,4 +27,8 @@
 	  <ProjectReference Include="..\DHSC.FingertipsNext.Modules.Core.Repository\DHSC.FingertipsNext.Modules.Core.Repository.csproj" />
     </ItemGroup>
 
+    <ItemGroup>
+      <Folder Include="Controllers\V1\" />
+    </ItemGroup>
+
```
gareth-allan-bjss, line 33: Any idea why this change to the `.csproj` file has come in? As far as I can see this directory has been removed.

gareth-allan-bjss (commented):
Thanks for cleaning up those additional files.

api/DHSC.FingertipsNext.Modules.Core/Module.cs
gareth-allan-bjss: I'm happy now the core and search modules have been removed.

gareth-allan-bjss (approved):

PR #70, opened 2025-01-21
Feature/dhscft 193 show area names in chart

Joe-Hendley:
# Description

Jira ticket: [DHSCFT-193](https://bjss-enterprise.atlassian.net/browse/DHSCFT-193)

- add area name to API & spec
- use area name on FE instead of area code

SonarCloud issues are, at time of writing, entirely localised in the autogenerated mocks.

PR #80, opened 2025-01-23
[DHSCFT-138]: update line charts table to match design

adawari-thomas-bjss:
# Description

Jira ticket: [DHSCFT-138](https://bjss-enterprise.atlassian.net/browse/DHSCFT-138)

## Changes

- Updated line chart table component to match Figma design
- Updated table to display health data for single area
- Updated MSW handler to filter results by selected area codes
- Added tests to line chart table component 

## Validation

image


adawari-thomas-bjss:
Currently, the chart component displays the first item from the list of results filtered by area code. However, in a subsequent ticket, it will be changed such that the chart will display data for a specific area code within the filtered results.

frontend/fingertips-frontend/components/organisms/LineChartTable/index.tsx
```diff
@@ -1,19 +1,77 @@
...
+  textAlign: 'right',
+  paddingTop: '0',
+});
+
+const StyledTable = styled(Table)({
+  marginBottom: '0.9375em',
+});
+
+const StyledTableCell = styled(Table.Cell)({
+  textAlign: 'left',
```
alibombaye, line 35: Looking at the designs the numerical columns are right aligned. Only the period seems to be left aligned

frontend/fingertips-frontend/components/organisms/LineChartTable/index.tsx
```diff
@@ -24,23 +82,18 @@ export function LineChartTable({ data, headings }: Readonly<TableProps>) {
...
-            </Table.Row>
-          ))
-        )}
-      </Table>
+        {tableData.map((point, index) => (
+          <Table.Row key={`${data.areaCode}-${point.period}--${index}`}>
+            <StyledTableCell numeric>{point.period}</StyledTableCell>
+            <StyledTableCell></StyledTableCell>
+            <StyledTableCell numeric>{point.count}</StyledTableCell>
+            <StyledTableCell numeric>{point.value}</StyledTableCell>
```
HowardMayBjss, line 90: These should be shown as percentages and I think to 1 decimal place

frontend/fingertips-frontend/components/organisms/LineChartTable/index.tsx
```diff
@@ -1,19 +1,77 @@
...
+              England
+            </StyledAlignRightHeader>
+          </Table.Row>
+        }
+      ></StyledTable>
+      <StyledTable
+        head={
+          <Table.Row>
+            <StyledAlignRightHeader setWidth="three-quarters">
+              95% confidence limits
```
HowardMayBjss, line 68: I presume that this should be coupled to the Lower and Upper confidence limit values

frontend/fingertips-frontend/components/organisms/LineChartTable/index.tsx
```diff
@@ -1,19 +1,77 @@
...
     <div data-testid="lineChartTable-component">
-      <Table
+      <StyledTable
+        head={
+          <Table.Row>
+            <StyledAlignCenterHeader setWidth="three-quarters">
+              {data.areaCode}
+            </StyledAlignCenterHeader>
+            <StyledAlignRightHeader setWidth="one-quarter">
+              England
```
HowardMayBjss, line 59: The England column should have a grey background.

frontend/fingertips-frontend/components/organisms/LineChartTable/index.tsx
alibombaye: Yeah I was just going to raise this point. This label needs to be right aligned with the UpperCI column.

I think what you are going to have to do for this row is merge the first 6 cells for this label that is then right aligned and have an empty cell at the end.

frontend/fingertips-frontend/components/organisms/LineChartTable/index.tsx
```diff
@@ -1,19 +1,77 @@
...
+    count: healthPoint.count,
+    value: healthPoint.value,
+    lower: healthPoint.lowerCi,
+    upper: healthPoint.upperCi,
+  }));
+
 export function LineChartTable({ data, headings }: Readonly<TableProps>) {
-  data = sortHealthDataByDate(data);
+  const tableData = mapToTableData(data);
+  tableData.sort((a, b) => b.period - a.period);
```
alibombaye, line 49: I think I would prefer if we created a new const of the sorted data. Rather than changing the original.

you can do this with something like this;

```
const dataSortedByPeriod = [...tableData].sort(Sort logic here);
```

frontend/fingertips-frontend/components/organisms/LineChartTable/index.tsx
alibombaye: We don't the unit of the value. They could be percentage or raw values. There is a follow on ticket to figure that out. For now we have agreed to show the raw value we get from the API

HowardMayBjss:
The designs also include a scroll bar at the bottom of the table. I presume that is in scope for this ticket?

frontend/fingertips-frontend/lib/chartHelpers/chartHelpers.ts
```diff
@@ -19,3 +19,13 @@ export function generateSeriesData(
...
 }
+
+export const headings = [
+  'Period',
+  'Trend',
+  'Count',
+  'Value',
+  'Lower',
+  'Upper',
+  'Value ',
```
alibombaye, line 30: Why is Value in here twice?

frontend/fingertips-frontend/components/organisms/LineChartTable/index.tsx
```diff
@@ -1,19 +1,77 @@
...
   return (
     <div data-testid="lineChartTable-component">
-      <Table
+      <StyledTable
+        head={
+          <Table.Row>
+            <StyledAlignCenterHeader setWidth="three-quarters">
+              {data.areaCode}
```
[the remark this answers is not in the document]
adawari-thomas-bjss: Would be changed once https://github.com/dhsc-govuk/FingertipsNext/pull/70 is merged

frontend/fingertips-frontend/components/organisms/LineChartTable/index.tsx
adawari-thomas-bjss: This is to be done in a separate ticket https://bjss-enterprise.atlassian.net/browse/DHSCFT-240

adawari-thomas-bjss:
Concerning the scroll bar, that was removed from the scope of the ticket @HowardMayBjss

frontend/fingertips-frontend/lib/chartHelpers/chartHelpers.ts
adawari-thomas-bjss: The second value column corresponds with the England column

alibombaye: Ah I see it's for the benchmark heading.

I think it might be worth making this an enum then. Something like this;

```
export const LineChartTableHeadingEnum  {
  AreaPeriod = 'Period',
  AreaTrend = 'Trend',
  AreaCount = 'Count',
  AreaValue = 'Value',
  AreaLower = 'Lower',
  AreaUpper = 'Upper',
  BenchmarkValue = 'Value',
}

```

frontend/fingertips-frontend/components/organisms/LineChartTable/LineChartTable.test.tsx
```diff
@@ -1,74 +1,68 @@
...
+    );
+    expect(screen.getByText(/95% confidence limits/i)).toBeInTheDocument();
+    expect(screen.getByText(/England/i)).toBeInTheDocument();
+    expect(valueCellHeadings).toHaveLength(2);
+    expect(screen.getAllByRole('cell')).toHaveLength(
+      MOCK_HEALTH_DATA.healthData.length * CELLS_PER_ROW
+    );
+
+    headings
+      .filter((heading) => heading.trim() !== 'Value')
```
alibombaye, line 43: I don't understand why we are doing this filter here?

adawari-thomas-bjss: the 'Value' heading appears twice on the screen, so the test fails because it finds multiple elements containing 'Value'. So, the filter is used to test the other headings, excluding 'Value', while 'Value' is tested separately.

frontend/fingertips-frontend/components/organisms/LineChartTable/index.tsx
alibombaye: Nice that's looking lot more like the designs.

HowardMayBjss:
The vertical spacing of the column titles seems a little off. Also the size of the text relative to the box seems a little too small. 
image
image

frontend/fingertips-frontend/components/organisms/LineChartTable/LineChartTable.test.tsx
alibombaye: ah I see. In which case I think it might be worthwhile getting these headings via the data-testids instead here. I'm not a fan of doing something different here based upon a particular value.

frontend/fingertips-frontend/components/organisms/LineChartTable/index.tsx
```diff
@@ -1,46 +1,170 @@
...
+              95% confidence limits
+            </StyledConfidenceLimitsHeader>
+          </Table.Row>
+        }
+      ></StyledTable>
+      <StyledTable
+        head={
+          <Table.Row>
+            {headings?.map((heading) =>
+              heading === 'Period' || heading === 'Trend' ? (
```
alibombaye, line 135: If you use the enum to iterate. We can use the enum here instead of hardcoding these

frontend/fingertips-frontend/components/organisms/LineChartTable/index.tsx
```diff
@@ -1,46 +1,170 @@
...
+    (a, b) => b.period - a.period
+  );
   return (
     <div data-testid="lineChartTable-component">
-      <Table
+      <StyledTable
+        head={
+          <Table.Row>
+            <StyledAlignCenterHeader setWidth="one-quarter"></StyledAlignCenterHeader>
+            <StyledAlignCenterHeader setWidth="one-quarter"></StyledAlignCenterHeader>
```
alibombaye, line 108: Do we need these 2?

frontend/fingertips-frontend/components/organisms/LineChartTable/index.tsx
```diff
@@ -1,46 +1,170 @@
...
+      ></StyledTable>
+      <StyledTable
         head={
           <Table.Row>
-            {headings?.map((heading) => (
-              <Table.CellHeader key={`header-${heading}`}>
-                {heading}
-              </Table.CellHeader>
+            <StyledAlignRightHeader setWidth="one-quarter"></StyledAlignRightHeader>
+            <StyledAlignRightHeader setWidth="one-quarter"></StyledAlignRightHeader>
```
alibombaye, line 121: Do we need these? I'm a little confused how this is working

alibombaye: I think it might be worth putting a border around the cells to see how these cells are getting rendered. I'm free for a chat to take a look at this.

frontend/fingertips-frontend/components/organisms/LineChartTable/index.tsx
```diff
@@ -1,46 +1,170 @@
...
+            <StyledAlignCenterHeader setWidth="one-quarter"></StyledAlignCenterHeader>
+            <StyledAlignCenterHeader>{data.areaCode}</StyledAlignCenterHeader>
+            {[...Array(7).keys()].map((i) => (
+              <Table.CellHeader key={i}></Table.CellHeader>
+            ))}
+            <StyledGreyHeader>England</StyledGreyHeader>
+          </Table.Row>
+        }
+      ></StyledTable>
+      <StyledTable
```
alibombaye, line 117: Ah I see - you have multiple tables. Whereas it should just be one.  This will start causing alignment issues 

What you can do is use `colspan` to make certain cells cover more than one cell. So for example the cell with the areaName and England. The areaName cell will have a colspan of 7. 

Sorry I didn't pick up on this before

frontend/fingertips-frontend/components/organisms/LineChartTable/index.tsx
adawari-thomas-bjss: removed

frontend/fingertips-frontend/components/organisms/LineChartTable/index.tsx
adawari-thomas-bjss: fixed

frontend/fingertips-frontend/components/organisms/LineChartTable/LineChartTable.test.tsx
```diff
@@ -1,74 +1,63 @@
...
+      MOCK_HEALTH_DATA.healthData.length * CELLS_PER_ROW
+    );
+    Object.values(LineChartTableHeadingEnum).forEach((heading, index) =>
+      expect(
+        screen.getByTestId(`header-${heading}-${index}`)
+      ).toBeInTheDocument()
+    );
+    screen.getAllByTestId('grey-table-cell').forEach((greyCell) => {
+      expect(greyCell).toHaveStyle(`background-color: ${LIGHT_GREY}`);
+    });
```
alibombaye, line 44: I think let's not bother with checking the styling as part of the unit tests. Unless we have some specific logic to do a specific style for a specific condition. 

Will use the snapshot tests for that purpose. Otherwise the tests will start to get too much.

alibombaye: Looking at the code now. I can see why you have done it. Can you do a seperate tests to say that the benchmark column should be in grey or something along those lines instead.

frontend/fingertips-frontend/components/organisms/LineChartTable/index.tsx
```diff
@@ -1,45 +1,155 @@
...
+            </StyledAlignRightTableCell>
+            <StyledAlignRightTableCell numeric>
+              {point.lower.toFixed(1)}
+            </StyledAlignRightTableCell>
+            <StyledAlignRightTableCell numeric>
+              {point.upper.toFixed(1)}
+            </StyledAlignRightTableCell>
+            <StyledBenchmarkValueTableCell data-testid="grey-table-cell"></StyledBenchmarkValueTableCell>
+          </Table.Row>
+        ))}
```
alibombaye, line 152: That's looking lot better now. Thanks for changing 👍

alibombaye (approved):

HowardMayBjss:
A new version of the design has been released with some chnages 
image
I think we will need to discuss whether we refactor as part of this ticket or create a separate ticket.

frontend/fingertips-frontend/components/organisms/LineChartTable/index.tsx
```diff
@@ -1,46 +1,222 @@
...
+});
+
+const StyledConfidenceLimitsHeader = styled(StyledAlignLeftHeader)({
+  width: '18%',
+  padding: '0.5em',
+});
+
+const StyledGreyHeader = styled(StyledAlignRightHeader)({
+  backgroundColor: LIGHT_GREY,
+  borderTop: `solid white 0.1875em`,
```
alibombaye, line 68: looking at the design I don't think the line is white but the same shade of grey as the rest. But with the colour of the cell it looks like its white.

frontend/fingertips-frontend/components/organisms/LineChartTable/index.tsx
```diff
@@ -1,46 +1,222 @@
...
+  textAlign: 'left',
+});
+
+const StyledAlignRightTableCell = styled(StyledTableCell)({
+  textAlign: 'right',
+});
+
+const StyledBenchmarkValueTableCell = styled(StyledTableCell)({
+  backgroundColor: LIGHT_GREY,
+  borderTop: `solid white 0.1875em`,
```
alibombaye, line 86: same here I don't think it should be white.

frontend/fingertips-frontend/components/organisms/LineChartTable/index.tsx
adawari-thomas-bjss: good spot

alibombaye: by the way just use `px` sizes for the border. The em doesn't offer any advantages here.

alibombaye (approved):

HowardMayBjss (approved):

PR #82, opened 2025-01-24
Change Azure Container Registry Server and Authentication Method

epercy-bjss:
# Description

Jira ticket: [DHSCFT-271](https://bjss-enterprise.atlassian.net/browse/DHSCFT-271)

Current builds are pushing images to a manually created ACR in Azure and use a username/password authentication. Refactoring is required in order to point a new ACR created in Terraform, then authenticate using Service Principal authentication. This change will also be reflected in the deployment workflows, to pull images.

## Changes

- Update GitHub Action to a supported version
- Change authentication to use Service Principal authentication
- Update ACR login server to point at the new ACR created in Terraform

## Validation

Branch has been run manually and is able to push the image successfully to the new ACR

### DIFF
--- a/.github/workflows/build-container.yml
+++ b/.github/workflows/build-container.yml
@@ -20,11 +20,11 @@ on:
         required: true
         type: string
     secrets:
-      acr-username:
-        description: 'Azure Container Registry username'
+      arm-client-id:
+        description: 'Service Principal Client ID'
         required: true
-      acr-password:
-        description: 'Azure Container Registry password'
+      arm-client-secret:
+        description: 'Service Principal Client Secret'
         required: true
       acr-server:
         description: 'Azure Container Registry server'
@@ -46,8 +46,8 @@ jobs:
         uses: docker/login-action@v3
         with:
           registry: ${{ secrets.ACR_SERVER }}
-          username: ${{ secrets.ARM_CLIENT_ID }}
-          password: ${{ secrets.ARM_CLIENT_SECRET }}
+          username: ${{ secrets.arm-client-id }}
+          password: ${{ secrets.arm-client-secret }}
 
       - name: Generate container metadata
         id: meta
--- a/.github/workflows/build-container.yml
+++ b/.github/workflows/build-container.yml
@@ -43,11 +43,11 @@ jobs:
         uses: docker/setup-buildx-action@v3
 
       - name: Azure Container Registry Login
-        uses: Azure/docker-login@v1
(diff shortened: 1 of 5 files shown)


.github/workflows/fingertips-frontend-build.yml
williamwillis11, line 1: were no changes needed to .github/workflows/fingertips-frontend-deploy.yml ?

epercy-bjss: Helps if I push that commit! :-D

.github/workflows/build-container.yml
```diff
@@ -20,11 +20,11 @@ on:
         required: true
         type: string
     secrets:
-      acr-username:
-        description: 'Azure Container Registry username'
+      arm-client-id:
```
HowardMayBjss, line 23: This is using github secrets to manage credentials rather than having secrets stored in Azure Key Store. I'm asking myself the pros and cons of this, but I haven't used github secrets before. Are there any issue with using secrets in a public repo? How does it work with multiple environments. Who can manage these keys? I know that this PR hasn't introduced the use of github secrets but I would value understanding it better at this point.

epercy-bjss: With regards to GitHub secrets, they can only be accessed by members with permissions to read the secrets within the GitHub organization. Generally, anyone with 'write' access to the repository can update them (through GitHub CLI). So members of the public, will not be able to view any secrets, only our references to them. When you execute a workflow that references a secret, the value gets obfuscated from the console. GitHub secrets _can_ be scoped to an environment, by setting up a [deployment environment](https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-deployments/managing-environments-for-deployment) and executing a deployment workflow in GitHub which references them. We are doing this for environment-scoped variables elsewhere in the project. But for secrets scoped at the Repo level, they can be used more generically throughout your workflows.

Key Vault is also perfectly valid to reference secrets, but doing so would be an environment-specific value, as the key vault resides in each workload Resource Group. To have generic Key Vault secrets, we could add a new Key Vault into the core Resource Group and make this global.

HowardMayBjss: That all makes sense, I would make the observation that .github secrets might not work so well for dynamic environments (?), but given we have no intent of supporting that at present then what we have here seems good to me.

HowardMayBjss (approved):

PR #83, opened 2025-01-24
Feature/DHSCFT-209_e2e_tests_data

williamwillis11:
# Description

Jira ticket: [DHSCFT-209](https://bjss-enterprise.atlassian.net/browse/DHSCFT-209)

Building on the work done in [PR](https://github.com/dhsc-govuk/FingertipsNext/pull/71) and [PR](https://github.com/dhsc-govuk/FingertipsNext/pull/72) the e2e test - search_results_and_view_chart.spec.ts will now run in 2 modes:
1. CI against every push in the .github/workflows/fingertips-frontend-build.yml job - test will work against mock data
2. CD against every merge to main in the .github/workflows/fingertips-frontend-deploy.yml - test will work against data added by search-setup from the json files in the path: search-setup/assets/*

## Changes

- Added support for e2e test to run in modes described above
- Made e2e test dynamic from data added in search setup
- Added e2e test summary to github actions job page
- Added support to upload a screenshot of any e2e test failures into the job artefacts
- Updated playwright to latest version

## Validation

E2E tests pass locally
E2E tests passing in CI mode - https://github.com/dhsc-govuk/FingertipsNext/actions/runs/12953150933
E2E tests will be checked in CD mode once this merges to main - i have run the journey manually against the deployed site and it is working with search 'mortality'


frontend/fingertips-frontend/playwright/testHelpers.ts
```diff
@@ -0,0 +1,16 @@
+interface IndicatorDocument {
```
williamwillis11, line 1: note that i didnt use IndicatorDocument from 

search-setup/src/constants.ts

as it defines lastUpdated as a Date which isn't how the data is stored in:

search-setup/assets/indicatorData.json

the other option would be to change the lastUpdated property to be `lastUpdated: Date | string;` in IndicatorDocument in search-setup/src/constants.ts but i didnt want to make that change without agreeing it with a dev first

priya-thorpe-bjss (approved):
Looks good. 
A couple of questions about the type for last Updated, 
- why are there two types for last updated? I would have thought it better to have one type for consistency across the the code, so we can use the same interface and avoid confusion.
- will date type only be a string going forward?

frontend/fingertips-frontend/playwright/testHelpers.ts
williamwillis11: have addressed this but in a way that i want to improve on in a future PR

PR #84, opened 2025-01-27
DHSCFT-270: add client factory for api

alibombaye:
# Description

Jira ticket: [DHSCFT-270](https://bjss-enterprise.atlassian.net/browse/DHSCFT-270)

Adds a factory function to return instance of the Areas or Indicators Api. Following review comments.

## Validation

Test still work, runs as expected


frontend/fingertips-frontend/lib/apiClient/apiClientFactory.test.ts
```diff
@@ -0,0 +1,16 @@
+import { AreasApi, IndicatorsApi } from '@/generated-sources/ft-api-client';
+import { ApiClientFactory } from './apiClientFactory';
+
+describe('apiClientFactory', () => {
+  it('getAreasApiClient return an instace of the AreasApi', () => {
```
williamwillis11, line 5: instace typo

frontend/fingertips-frontend/lib/apiClient/apiClientFactory.test.ts
```diff
@@ -0,0 +1,16 @@
...
+    const apiClient = ApiClientFactory.getAreasApiClient();
+
+    expect(apiClient).toBeInstanceOf(AreasApi);
+  });
+
+  it('getIndicatorsApiClient return an instace of the IndicatorsApi', () => {
```
williamwillis11, line 11: instace typo

alibombaye: Good spot - Updated

frontend/fingertips-frontend/lib/apiClient/apiClientFactory.test.ts
alibombaye: Good spot - Updated

frontend/fingertips-frontend/package.json
```diff
@@ -4,7 +4,7 @@
   "private": true,
   "scripts": {
     "dev:local-api": "concurrently \"npm run api:start\" \"next dev\" && npm run api:stop",
-    "dev": "cross-env MOCK_SERVER=true next dev",
+    "dev": "cross-env DHSC_AI_SEARCH_USE_MOCK_SERVICE=true MOCK_SERVER=true next dev",
```
gareth-allan-bjss, line 7: Does this need to be added? There's a `frontend/fingertips-frontend/.env.development` file that includes this value, and should be getting loaded when `npm run dev` is run.

In fact, I think we should probably move MOCK_SERVER=true into this file too.

frontend/fingertips-frontend/components/pages/chart/index.tsx
```diff
@@ -38,41 +37,34 @@ export function Chart({
         aria-label="Go back to the previous page"
       />
       <H2>View Dementia QOF prevalence</H2>
-      <GridRow>
-        <GridCol setWidth="one-third">
-          <AreaFilter availableAreaTypes={[]} />
```
gareth-allan-bjss, line 43: Just to check, you're intending to remove the `AreaFilter`, right? This was the component that was added prematurely?

frontend/fingertips-frontend/package.json
alibombaye: ah ok when I did an npm run dev - I wasn't getting any data comeback on the results page. So I added this to fix it. But I'll check again.

alibombaye: So `.env.development` is being gitignored. Which is why it didn't work. Do we want to gitignore it?

alibombaye: Doesn't seem to work 😢  - If you get a chance can you try locally too

alibombaye: My bad it does work. Just realised I can't just use any random word to test with

gareth-allan-bjss (approved):

PR #93, opened 2025-01-28
Feature/dhscft tech playback demo

steves-bjss:
Containts test data used during the recent playback demo, containing a cutdown data set for Hypertension which allows chart visualisations to be seen more easily.


search-setup/assets/indicatorData.json
```diff
@@ -39,6 +39,14 @@
...
         "lastUpdated": "20-Jan-24"
     },
+    {
+        "indicatorId": "219",
+        "name": "Hypertension: QOF prevalence (all ages)",
+        "definition": "The percentage of patients with established hypertension, as recorded on practice disease registers (proportion of total list size).",
+        "latestDataPeriod": "2024",
+        "dataSource": "Quality and Outcomes Framework (QOF), NHS England",
+        "lastUpdated": "24-Sep-24"
+    },
```
gareth-allan-bjss, line 49: We also have `frontend/fingertips-frontend/assets/mockIndicatorData.json` which is a very similar file and drives the mock AI search instance when running locally. I think it would be worth adding this same entry to it, so we have consistency between running locally and a deployed environment.

steves-bjss: Good shout - I've added it in.

gareth-allan-bjss (approved):

PR #95, opened 2025-01-28
Feature/dhscft 159 initial scatter plot

priya-thorpe-bjss:
# Description

Jira ticket: [DHSCFT-159](https://bjss-enterprise.atlassian.net/browse/DHSCFT-159)

## Changes

- Added a second api call with another indicator.
- Created a scatter chart that displays data from two api calls.

## Validation
- Tests pass
- Have tested with the api

Screenshot 2025-01-28 at 16 35 45


frontend/fingertips-frontend/app/chart/page.tsx
```diff
@@ -25,11 +25,19 @@ export default async function ChartPage(
...
     indicatorId: Number(indicatorsSelected[0]),
     areaCodes: areaCodes,
   });
 
+  const dataSetTwo = await indicatorApi.getHealthDataForAnIndicator({
+    indicatorId: Number(indicatorsSelected[1]),
+    areaCodes: areaCodes,
+  });
+
+  const scatterData = [data, dataSetTwo];
```
gareth-allan-bjss, line 39: I wonder if we'd be better having this page just make `getHealthDataForAnIndicator` requests for all indicators provided (i.e. in a loop) and have the `Chart` component deal with which values it uses. For now it could just use `data[0]` for the existing charts and `[data[0], data[1]]` for the scatter plot.

I don't think it makes sense to pass a separate `scatterData` prop to `Chart`, given we know we'll ultimately want to remove it.

frontend/fingertips-frontend/components/pages/chart/index.tsx
```diff
@@ -49,6 +52,15 @@ export function Chart({
           <br />
         </>
       ) : null}
+      {scatterData ? (
+        <ScatterChart
+          data={scatterData}
+          ScatterChartTitle="Compare indicators within the area group"
+          yAxisTitle="<b>y: Indicator 1 (value)</b></br> rate per information"
+          xAxisTitle="<b>x: Indicator 2 (value)</b></br> rate per information"
```
gareth-allan-bjss, line 60: Should we be passing formatting into the prop values? Perhaps we need to consider splitting these props up if there are separate elements to the label? i.e.:

- `y: Indicator 1 (value)`
- `x: Indicator 2 (value)`
- `rate per information`

frontend/fingertips-frontend/playwright/page-objects/pages/chartPage.ts
```diff
@@ -1,14 +1,15 @@
...
+import { screen } from '@testing-library/react';
 
 export default class ChartPage extends BasePage {
   readonly backLink = 'chart-page-back-link';
   readonly lineChartComponent = 'lineChart-component';
   readonly lineChartTableComponent = 'lineChartTable-component';
   readonly barChartComponent = 'barChart-component';
   readonly populationPyramidComponent = 'populationPyramid-component';
-
+  readonly scatterChart = 'scatterChart-component';
```
gareth-allan-bjss, line 12: Think you need a newline after this:
```suggestion
  readonly scatterChart = 'scatterChart-component';
  
```

gareth-allan-bjss: Oh, and it should probably be called `scatterChartComponent` for consistency with the other variable names.

frontend/fingertips-frontend/app/chart/page.tsx
priya-thorpe-bjss: yes, currently i've put in a fix that doesn't affect the data rendering on the other charts, so the ticket can go in.
I have created a subtask to return data from the api call as an object and handle data for the other charts in that ticket.

priya-thorpe-bjss: Some of the table work has gone in which is affecting the fix, I think the best solution would be to keep it as it was originally and then do the refactoring of the api call and data handling in the subtask. This saves having to change the data and tests three times.

frontend/fingertips-frontend/components/pages/chart/index.tsx
```diff
@@ -52,31 +51,33 @@ export function Chart({
           <br />
         </>
       ) : null}
-      {scatterData ? (
+      {indicatorsSelected.length > 1 ? (
```
gareth-allan-bjss, line 54: I think (based on [this page](https://confluence.collab.test-and-trace.nhs.uk/pages/viewpage.action?pageId=419245267)) this should only show the scatter chart if `indicatorsSelected.length` is exactly 2:
```suggestion
      {indicatorsSelected.length == 2 ? (
```

frontend/fingertips-frontend/components/organisms/ScatterChart/scatterChart.test.tsx
```diff
@@ -0,0 +1,29 @@
...
+  const yAxisPropsTitle = 'DifferentYTitle';
+  render(
+    <ScatterChart data={[mockHealthData[1]]} yAxisTitle={yAxisPropsTitle} />
+  );
+
+  const highcharts = screen.getByTestId('highcharts-react-component');
+
+  expect(highcharts).toBeInTheDocument();
+  expect(highcharts).toHaveTextContent(yAxisPropsTitle);
+});
```
gareth-allan-bjss, line 16: The `ScatterChart` component has a few more properties controlling titles than just the `yAxisTitle`. We should presumably have assertions in this test for those too.

frontend/fingertips-frontend/components/pages/chart/index.tsx
```diff
@@ -8,12 +8,14 @@ import { SearchStateManager } from '@/lib/searchStateManager';
...
 import { PopulationData } from '@/lib/chartHelpers/preparePopulationData';
+import { ScatterChart } from '@/components/organisms/ScatterChart';
 
 type ChartProps = {
-  data: HealthDataForArea[];
+  data: HealthDataForArea[][];
   populationData?: PopulationData;
   searchedIndicator?: string;
   indicatorsSelected?: string[];
+  // scatterData?: HealthDataForArea[][];
```
gareth-allan-bjss, line 18: This just needs removing now.

frontend/fingertips-frontend/components/organisms/ScatterChart/scatterChart.test.tsx
```diff
@@ -3,10 +3,20 @@ import { render, screen } from '@testing-library/react';
 import { mockHealthData } from '@/mock/data/healthdata';
 import { expect } from '@jest/globals';
 
-it('should render the Highcharts react component with passed yAxisPropsTitle parameter', () => {
+it('should render the Highcharts react component with the passed axis and subtitle parameters', () => {
   const yAxisPropsTitle = 'DifferentYTitle';
+  const xAxisPropsTitle = 'DifferentXTitle';
+  const yAxisSubtitle = 'DifferentXSubtitle';
```
gareth-allan-bjss, line 9:
```suggestion
  const yAxisSubtitle = 'DifferentYSubtitle';
```

frontend/fingertips-frontend/components/organisms/ScatterChart/scatterChart.test.tsx
gareth-allan-bjss: You still need to add the `expect(highcharts).toHaveTextContent(<whatever>)` assertions to this test.

frontend/fingertips-frontend/app/chart/page.tsx
gareth-allan-bjss: I'm happy with the solution that creates an array of results for now.

frontend/fingertips-frontend/lib/chartHelpers/scatterChartHelper.ts
```diff
@@ -0,0 +1,10 @@
+import { HealthDataForArea } from '@/generated-sources/ft-api-client';
+
+export function sortHealthDataByYearDescending(
+  data: HealthDataForArea[] = []
+): HealthDataForArea[] {
+  return data.map((item) => ({
+    ...item,
+    healthData: item.healthData.toSorted((a, b) => b.year - a.year),
+  }));
+}
```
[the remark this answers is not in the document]
gareth-allan-bjss: No, it does the reverse sort.

PR #119, opened 2025-02-03
Create database workflows

epercy-bjss:
# Description

Jira ticket: [DHSCFT-191](https://bjss-enterprise.atlassian.net/browse/DHSCFT-191)

Create database workflows to build and deploy the database project to Azure SQL Database. API workflow also updated to wait for DB deploy if changes are made to both directories.

## Changes

- Create a build workflow that builds the database project and uploads the dacpac to GitHub Artifacts
- Create a deployment workflow that is triggered to execture after a successful build
- Artifact is picked up from either the same sha (if workflow_run), or the latest artifact from the main branch if workflow_dispatch (manually run).
- Add a new task to the API deployment workflow to check for changes on the same sha to the 'database' directory. Poll for completion of the database deployment workflow if so.

## Validation

Database deployment and API deployment job to wait for database deployment has been tested from branch, but refactoring was needed to revert to a 'workflow_run' trigger. This will need to be tested again once merged in.

### DIFF
--- a/.github/workflows/fingertips-api-deploy.yml
+++ b/.github/workflows/fingertips-api-deploy.yml
@@ -23,6 +23,8 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
+        with:
+          fetch-depth: 2
 
       - name: Check for changes in database directory
         id: check
@@ -59,7 +61,7 @@ jobs:
       gitversion-version: "6.0.x"
 
   api-deploy-development:
-    needs: determine-version
+    needs: [determine-version, check-db-changes]
     uses: ./.github/workflows/deploy-container.yml
     with:
       tf_state_file_name: fingertips.api.tfstate
--- a/.github/workflows/fingertips-api-deploy.yml
+++ b/.github/workflows/fingertips-api-deploy.yml
@@ -37,22 +37,50 @@ jobs:
             echo "DB_CHANGED=false" >> $GITHUB_ENV
           fi
 
-      - name: Check Database Deployment Status
+      - name: Wait for Database Deployment to Complete
         uses: actions/github-script@v7
         id: database_status
         if: env.DB_CHANGED == 'true'
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
-          retries: 3
+          retries: 10  # Retry up to 10 times
           script: |
-            const runs = await github.rest.actions.listWorkflowRuns({
-              owner: context.repo.owner,
-              repo: context.repo.repo,
-              workflow_id: 'fingertips-db-deploy.yml',
-              head_sha: '${{ github.sha }}',
-              status: 'completed'
-            });
-            core.setOutput("db_completed", runs.data.workflow_runs.length > 0 ? 'true' : 'false');
+            async function waitForDBDeployment() {
+              const workflowName = 'fingertips-db-deploy.yml';
+              const headSHA = '${{ github.sha }}';
+              const maxRetries = 20;
+              const delay = 30 * 1000; // 30 seconds per check
+              
+              for (let i = 0; i < maxRetries; i++) {
+                const runs = await github.rest.actions.listWorkflowRuns({
+                  owner: context.repo.owner,
+                  repo: context.repo.repo,
+                  workflow_id: workflowName,
+                  head_sha: headSHA,
+                });
+  
+                if (runs.data.workflow_runs.length > 0) {
+                  const latestRun = runs.data.workflow_runs[0]; // Get the most recent run
+                  
+                  if (latestRun.status === 'completed') {
+                    console.log(`Database deployment completed with status: ${latestRun.conclusion}`);
+                    core.setOutput("db_completed", 'true');
+                    return;
+                  } else {
+                    console.log(`Database deployment is still running. Status: ${latestRun.status}`);
+                  }
+                } else {
+                  console.log("No database deployment run found yet.");
+                }
+  
+                await new Promise(resolve => setTimeout(resolve, delay)); // Wait before retrying
+              }
+  
+              console.log("Database deployment did not complete within the timeout period.");
+              core.setFailed("Database deployment did not complete in time.");
+            }
+  
+            await waitForDBDeployment();
 
   determine-version:
     uses: ./.github/workflows/determine-semver.yml
--- a/.github/workflows/fingertips-api-deploy.yml
+++ b/.github/workflows/fingertips-api-deploy.yml
@@ -49,7 +49,7 @@ jobs:
               owner: context.repo.owner,
               repo: context.repo.repo,
               workflow_id: 'fingertips-db-deploy.yml',
-              head_sha: ${{ github.sha }},
+              head_sha: '${{ github.sha }}',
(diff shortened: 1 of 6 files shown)


.github/workflows/fingertips-api-deploy.yml
```diff
@@ -10,14 +10,79 @@ on:
       - main
 
 jobs:
+  check-db-changes:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 2
```
williamwillis11, line 19: in the other yml files we have 

fetch-depth: 0

do we want this to be 2?

epercy-bjss: It needs to be set to `2`, as it is looking for files changed between the last commit and the current commit in subsequent `git diff` command.

.github/workflows/fingertips-api-deploy.yml
```diff
@@ -10,14 +10,79 @@ on:
...
+        if: env.DB_CHANGED == 'true'
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          retries: 10
+          script: |
+            async function waitForDBDeployment() {
+              const workflowName = 'fingertips-db-deploy.yml';
+              const headSHA = '${{ github.event.workflow_run.head_branch }}';
+              const maxRetries = 20;
+              const delay = 30 * 1000; // 30 seconds per check
```
williamwillis11, line 45: would calling this const 

delayMs

make it clearer that this delay is in milliseconds

.github/workflows/fingertips-api-deploy.yml
```diff
@@ -10,14 +10,79 @@ on:
...
+                    console.log(`Database deployment is still running. Status: ${latestRun.status}`);
+                  }
+                } else {
+                  console.log("No database deployment run found yet.");
+                }
+  
+                await new Promise(resolve => setTimeout(resolve, delay));
+              }
+  
+              console.log("Database deployment did not complete within the timeout period.");
```
williamwillis11, line 72: is it worth saying what the timeout period is in this log?

.github/workflows/fingertips-db-build.yml
```diff
@@ -0,0 +1,51 @@
...
+      - name: Setup .NET
+        uses: actions/setup-dotnet@v4
+        with:
+          dotnet-version: ${{ env.dotnet-version }}
+
+      - name: Build database project
+        run: dotnet build fingertips-db.sqlproj --configuration ${{ env.dotnet-configuration }}
+
+      - name: Upload .dacpac file artifact
+        uses: actions/upload-artifact@v4
```
williamwillis11, line 46: do we want to define a retention period for this uploaded file? or happy with the 90 day default?

williamwillis11: also looking at the other defaults it seems to compress the upload if we don't specify not to, is that ok?

.github/workflows/fingertips-db-deploy.yml
```diff
@@ -0,0 +1,60 @@
...
+          $RUN_ID = $response.workflow_runs[0].id
+        }
+
+        Write-Output "Looking for $RUN_ID"
+        echo "RUN_ID=$RUN_ID" | Out-File -FilePath $env:GITHUB_ENV -Append
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+    - name: Download DACPAC
+      uses: actions/download-artifact@v4
```
williamwillis11, line 49: where is this downloading to?

.github/workflows/fingertips-api-deploy.yml
epercy-bjss: Amended as suggested

.github/workflows/fingertips-api-deploy.yml
epercy-bjss: Added `...(30s)` to the end of the console log

.github/workflows/fingertips-db-build.yml
epercy-bjss: Yea, I think that's all fine, I've tested the upload/download of the artifact and it works as expected

.github/workflows/fingertips-db-deploy.yml
epercy-bjss: I downloads to the default working firectory and uncompresses. Again, this has been tested.

.github/workflows/fingertips-frontend-deploy.yml
```diff
@@ -10,6 +10,71 @@ on:
       - main
 
 jobs:
+  check-api-changes:
```
gareth-allan-bjss, line 13: Do you need to add this new `check-api-changes` job to the `needs` property for `frontend-deploy-development`? Presumably the deployment won't actually wait for it to finish otherwise?

.github/workflows/fingertips-frontend-deploy.yml
```diff
@@ -10,6 +10,71 @@ on:
...
+        run: |
+          if git diff --name-only HEAD~1 HEAD | grep '^api/'; then
+            echo "API changes detected"
+            echo "API_CHANGED=true" >> $GITHUB_ENV
+          else
+            echo "No API changes detected"
+            echo "API_CHANGED=false" >> $GITHUB_ENV
+          fi
+
+      - name: Wait for API Deployment to Complete
```
gareth-allan-bjss, line 33: It doesn't have to be part of this PR, but we should probably look to extract this to a re-usable function or script at some point, given we're now doing this check in a couple of places.

.github/workflows/fingertips-db-deploy.yml
```diff
@@ -0,0 +1,82 @@
...
+      ${{
+        github.event_name == 'workflow_dispatch' ||
+        (
+          github.event_name == 'workflow_run' &&
+          (
+            github.event.workflow_run.event == 'push' ||
+            github.event.workflow_run.event == 'workflow_dispatch'
+          )
+        )
+      }}
```
gareth-allan-bjss, line 76: What's this doing?

.github/workflows/fingertips-frontend-deploy.yml
epercy-bjss: Well spotted. I've added this in now

.github/workflows/fingertips-frontend-deploy.yml
epercy-bjss: Yes, that would probably be a good plan. If we don't find a better solution to this workflow strategy, then I'll refactor as a reusable workflow.

.github/workflows/fingertips-db-deploy.yml
epercy-bjss: So each deployment workflow can also be run manually from branch (using 'workflow_dispatch'). We have a condition here to allow the end-to-end-tests to run in that eventy. However, if the workflow has been triggered by a successful build (workflow_run), it will also run the end-to-end tests if the build was triggered by either a 'push', or if the build was also run manually.

PR #100, opened 2025-01-29
Feature/dhscft tech fix opentel error

epercy-bjss:
# Description

Fix for console error registering Application Insights Opentelemetry:

```
Error: @opentelemetry/api: Attempted duplicate registration of API: context
```

## Changes

- Avoids multiple API registrations by ensuring telemetryInitialized prevents duplicate setup.
- Ensures OpenTelemetry and Azure Monitor do not conflict by only initializing one of them.
- Refactored the tracing setup to be cleaner and added explicit logging for better debugging.
- Live Metrics has been disabled due to a bug with Next.js v15 - https://github.com/Azure/azure-sdk-for-js/pull/32720

## Validation

Running the container locally has been succesful and traces are coming through to the Application Insights resource in Azure from the local container

PR #127, opened 2025-02-05
Fix the lint script saved in package.json

williamwillis11:
# Description

Jira ticket: [DHSCFT-305](https://bjss-enterprise.atlassian.net/browse/DHSCFT-305)

The lint script in package json was incorrectly doing `next lint` rather than `eslint` - this therefore meant it was missing some issues in main.

## Changes

- change lint script to use eslint
- configure eslint to use next linting as well as whats already configured
- configure eslint to ignore .next and auto generated files from the mock service worker
- add eslint-plugin-playwright to add linting to the playwright tests
- addressed lint issues found in frontend/fingertips-frontend/instrumentation.ts (see Validation to prove this didnt regress that implementation which @epercy-bjss mentioned needed some work from him)
- addressed lint issues found in frontend/fingertips-frontend/highcharts.d.ts @WolfieKnee  fyi
- bumped typescript

## Validation

ran lint and playwright tests locally - success
ran typecheck locally - success
both pass in CI - success

checked instrumentation.ts changes havent stopped the app from connecting to azure app insights:
 
First screenshot shows the same terminal errors as in main (confirmed with @epercy-bjss):

Screenshot 2025-02-06 at 11 24 28

Second shows we are still connecting to azure application insights:

image (34)

Third shows how we generated an error in the webapp:

Screenshot 2025-02-06 at 11 28 51


frontend/fingertips-frontend/package.json
```diff
@@ -64,9 +64,10 @@
     "axe-playwright": "^2.0.3",
     "concurrently": "^9.1.2",
     "cross-env": "^7.0.3",
-    "eslint": "^9.19.0",
+    "eslint": "^9.17.0",
```
gareth-allan-bjss, line 67: Why downgrade ESlint?

williamwillis11: wasnt intentional - not sure how i managed to do that! good spot - ill bump back up

frontend/fingertips-frontend/instrumentation.ts
```diff
@@ -31,24 +31,29 @@ const configureApplicationInsights = async () => {
...
+    const { ConsoleSpanExporter } = await import(
+      '@opentelemetry/sdk-trace-base'
+    );
+    const { Resource } = await import('@opentelemetry/resources');
+    const { ATTR_SERVICE_NAME } = await import(
+      '@opentelemetry/semantic-conventions'
+    );
 
-    const spanProcessor = new SimpleSpanProcessor();
+    const spanProcessor = new SimpleSpanProcessor(new ConsoleSpanExporter());
```
Joe-Hendley, line 49: are we sure we want to have the console span exporter on by default? It's *very* verbose - hitting the homepage produces 339 lines of output in my console

williamwillis11: yeh good point - SimpleSpanProcessor wants an SpanExporter to be passed in by default otherwise it complains with

An argument for '_exporter' was not provided.

any ideas?

gareth-allan-bjss: Can you just not give it a `SpanProcessor`?

gareth-allan-bjss: Oh, or it looks like there's a [NoopSpanProcessor](https://open-telemetry.github.io/opentelemetry-js/classes/_opentelemetry_sdk_trace_base.NoopSpanProcessor.html)?

frontend/fingertips-frontend/highcharts.d.ts
WolfieKnee, line 1: works for me.

frontend/fingertips-frontend/instrumentation.ts
alibombaye: I know @epercy-bjss has done some work in this area and has a PR which does tidy this up https://github.com/dhsc-govuk/FingertipsNext/pull/100 It's not merged in yet. 

Would this be a better approach?

Joe-Hendley: It needs a no-op Exporter, which unfortunately doesn't seem to exist. We can roll our own with but this might not be necessary with the incoming changes to this file from https://github.com/dhsc-govuk/FingertipsNext/pull/100

```
    const noopSpanExporter: SpanExporter = {
      export: function (
        spans: ReadableSpan[],
        resultCallback: (result: ExportResult) => void
      ): void {},
      shutdown: function (): Promise<void> {
        return new Promise((resolve) => resolve());
      },
    };
```

williamwillis11: @Joe-Hendley are you happy to push this solution to the branch?

PR #129, opened 2025-02-06
Feature/DHSCFT-215_benchmarkcomponent

obarojohnson:
# Description

Jira ticket: [DHSCFT-215](https://bjss-enterprise.atlassian.net/browse/DHSCFT-215)

Benchmark UI Component , implementation  base on this figma design.

## Changes

- Create a new  LegendLabel component  with theme base on the props type.
- A new Benchmark component that houses the Label components  
- The components are all stateless 

## Unit testing's

The only testing provided is the component unit testing  , this ensure that the component colours are rather correctly
and also some manually testing to see how it appears.

## Benchmark Component

A snapshot of the manually testing of the component in  full screen mode.

image

Mobile view of the same component

image

The component that uses NHS-Gov components with minimal styling will ensure the UI is compatible with existing component text.

image


.vscode/settings.json
```diff
@@ -1,5 +1,8 @@
 {
   "dataworkspace.excludedProjects": [],
   "editor.defaultFormatter": "esbenp.prettier-vscode",
-  "editor.formatOnSave": true
+  "editor.formatOnSave": true,
+     // Jest configuration    
+     "jest.debugMode": true,
```
gareth-allan-bjss, line 6: Thanks for sorting this config, but I don't think we want to turn on debug mode, do we?
```suggestion
```

frontend/fingertips-frontend/components/organisms/BenchmarkLegend/LegendLabel.tsx
```diff
@@ -0,0 +1,173 @@
...
+  WORST = 'worst',
+  WORSE = 'worse',
+  BEST = 'best',
+  NOT_COMPARED = 'not_compared',
+}
+
+export const enum LegendLabelGroupType {
+  RAG = 'rag',
+  QUINTILE = 'quintile',
+  OTHERS = 'others',
```
gareth-allan-bjss, line 25: This others group is actually quintiles with a value judgement (and the `QUINTILE`) group is quintiles without a value judgement, so we should name them accordingly.

frontend/fingertips-frontend/components/organisms/BenchmarkLegend/LegendLabel.tsx
```diff
@@ -0,0 +1,173 @@
...
+    lineHeight: '16px',
+    letterSpacing: '1px',
+    margin: '0.5px',
+    fontFamily: 'GDS Transport',
+    ...theme,
+  };
+});
+
+export const LegendLabel: React.FC<LegendLabelProps> = ({
+  label,
```
gareth-allan-bjss, line 161: I don't think the label should be passed in to this component, as it will always be related to the `type` passed. I think this component should just be told the type and group and should pick the appropriate text itself.

frontend/fingertips-frontend/components/organisms/BenchmarkLegend/LegendLabel.tsx
```diff
@@ -0,0 +1,173 @@
...
+    padding: '5px 8px 4px 8px',
+    alignItems: 'center',
+    fontSize: '12px',
+    fontStyle: 'normal',
+    fontWeight: '300',
+    lineHeight: '16px',
+    letterSpacing: '1px',
+    margin: '0.5px',
+    fontFamily: 'GDS Transport',
+    ...theme,
```
gareth-allan-bjss, line 156: I wasn't necessarily expecting us to have to re-style the govuk-react label so much. Do we need to apply all this styling (changing the font, etc.?). I think you may be speaking to @alibombaye about this already.

frontend/fingertips-frontend/components/organisms/BenchmarkLegend/LegendLabel.tsx
gareth-allan-bjss, line 1: I think we're going to want to use this label in other places than the legend for maps/heatmaps, so it probably shouldn't be called a legend label. How about `BenchmarkLabel` or something similar? As it's essentially a label for a benchmark value.

Also, shouldn't it be in a separate directory from the `BenchmarkLegend` component? I think we want to model this as it's own component that can be used independently.

frontend/fingertips-frontend/components/organisms/BenchmarkLegend/BenchmarkLegend.tsx
```diff
@@ -0,0 +1,72 @@
...
+  title?: string;
+  group: string;
+  labels: LegendLabelData[];
+}
+
+export interface BenchmarkProps {
+  model?: BenchmarkData[] | undefined;
+}
+
+export const Benchmark: React.FC<BenchmarkProps> = ({ model }) => {
```
gareth-allan-bjss, line 45: This means that when we're adding the legend to a page we'll need to provide configuration each time for which labels to show. I think as we discussed in our kickoff session we probably just want to bake this knowledge into this component, so the pages using it don't need to apply configuration.

frontend/fingertips-frontend/components/organisms/BenchmarkLegend/BenchmarkLegend.test.tsx
gareth-allan-bjss, line 1: Again, I think this should have a snapshot test.

frontend/fingertips-frontend/components/organisms/BenchmarkLegend/LegendLabel.tsx
obarojohnson: Oh! this is the style from the  figma design apart from  **display: 'inline-block',** to make it responsive as you suggestion

frontend/fingertips-frontend/components/organisms/BenchmarkLegend/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+  },
+});
+
+export interface BenchmarkData {
+  title?: string;
+  group: string;
+  types: string[];
+}
+
+export const Benchmark = () => {
```
gareth-allan-bjss, line 29: I think this should probably be called `BenchmarkLabel`:
```suggestion
export const BenchmarkLabel = () => {
```

frontend/fingertips-frontend/components/organisms/BenchmarkLegend/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+  const model: BenchmarkData[] = [
```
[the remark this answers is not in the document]
gareth-allan-bjss: I think this is fine for now.

frontend/fingertips-frontend/components/organisms/BenchmarkLegend/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+  return (
+    <div>
+      {model?.map((item, index) => (
+        <DefaultBenchmarkLegendGroupPanelStyle key={index}>
+          <div>
+            {item.title !== null && (
+              <div>
+                <LabelText className="legend_panel_header">
+                  {item.title}
+                </LabelText>
```
gareth-allan-bjss, line 56: Is there [heading](https://govuk-react.github.io/govuk-react/?path=/docs/heading--docs) level we could use instead of styling a `LabelText` component? From [the docs](https://govuk-react.github.io/govuk-react/?path=/docs/label--docs), it looks like the `LabelText` component is intended as a label for form input and is meant to be wrapped in a `Label` component, so I don't think it's the right thing to be using here.

frontend/fingertips-frontend/components/organisms/BenchmarkLegend/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+              </div>
+            )}
+            {item.types.map((type, label_key) => (
+              <BenchmarkLabel
+                group={item.group}
+                type={type}
+                key={label_key}
+                data-testid={item?.group + '_' + type + '_' + label_key}
```
gareth-allan-bjss, line 64: I think you can still remove the `?` here:
```suggestion
                data-testid={item.group + '_' + type + '_' + label_key}
```

frontend/fingertips-frontend/components/organisms/BenchmarkLabel/index.tsx
```diff
@@ -0,0 +1,197 @@
...
+  QUINTILES_WITH_VALUE = 'quintiles_wv',
+}
+
+interface BenchmarkLabelProps {
+  type?: BenchmarkLabelType | string;
+  group?: BenchmarkLabelGroupType | string;
+}
+
+export const getDefaultBenchmarkTagStyle = (
+  group: BenchmarkLabelGroupType = BenchmarkLabelGroupType.QUINTILES_WITH_VALUE,
```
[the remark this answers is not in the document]
obarojohnson: you re right we are definitely going to pass the group anyway :)

frontend/fingertips-frontend/components/organisms/BenchmarkLabel/__snapshots__/BenchmarkLabel.test.tsx.snap
```diff
@@ -0,0 +1,66 @@
...
+  padding-left: 8px;
+  outline: 2px solid transparent;
+  outline-offset: -2px;
+  -webkit-letter-spacing: 1px;
+  -moz-letter-spacing: 1px;
+  -ms-letter-spacing: 1px;
+  letter-spacing: 1px;
+  -webkit-text-decoration: none;
+  text-decoration: none;
+  text-transform: uppercase;
```
gareth-allan-bjss, line 25: I think this is resulting in the text on the labels coming out as upper-case (i.e. `BEST` rather than `Best`). Adding the `textTransform: 'unset'` value to the styling of the `Tag` component, as we do for a `TrendTag` seems to resolve the issue.

frontend/fingertips-frontend/components/organisms/BenchmarkLabel/index.tsx
```diff
@@ -0,0 +1,197 @@
...
+  const theme = getDefaultBenchmarkTagStyle(group, legendType);
+  return {
+    display: 'inline-block',
+    padding: '5px 8px 4px 8px',
+    alignItems: 'center',
+    fontSize: '12px',
+    fontWeight: '300',
+    fontStyle: 'normal',
+    lineHeight: '16px',
+    margin: '0.0225em',
```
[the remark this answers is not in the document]
obarojohnson: I will remove that then  ,  that make sense.

frontend/fingertips-frontend/components/organisms/BenchmarkLegend/index.tsx
obarojohnson: This is the main component with many labels.  I agreed a suitable name should be given, instead of Benchmark , should I called it BenchmarkLabelPanel ? since this component can be reuse.
image

obarojohnson: or maybe BenchmarkLegend

frontend/fingertips-frontend/components/organisms/BenchmarkLegend/index.tsx
obarojohnson: Yes I think I can use h4 for it , and control the font Weight

frontend/fingertips-frontend/components/organisms/BenchmarkLabel/__snapshots__/BenchmarkLabel.test.tsx.snap
obarojohnson: I forget to remove it that transformation, good spot there thanks.

frontend/fingertips-frontend/components/organisms/BenchmarkLabel/BenchmarkLabel.test.tsx
```diff
@@ -0,0 +1,134 @@
...
+    [
+      'returns default style when group is not provided',
+      undefined,
+      BenchmarkLabelType.BETTER,
+      {
+        backgroundColor: '#812972',
+        tint: 'SOLID',
+      },
+    ],
+  ])('%s', (_, group, type, expected) => {
```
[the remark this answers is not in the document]
obarojohnson: Brilliant suggestion here

frontend/fingertips-frontend/components/organisms/BenchmarkLabel/__snapshots__/BenchmarkLabel.test.tsx.snap
obarojohnson: yeah , I should be using capitalisation instead.

frontend/fingertips-frontend/components/organisms/BenchmarkLabel/BenchmarkLabel.test.tsx
```diff
@@ -0,0 +1,121 @@
+import { render } from '@testing-library/react';
+import {
+  BenchmarkLabel,
+  BenchmarkLabelGroupType,
+  BenchmarkLabelType,
+  getDefaultBenchmarkTagStyle,
+} from '@/components/organisms/BenchmarkLabel/index';
+
+describe('testing the function getBenchmarkLegendColourStyle', () => {
+  test.each([
```
gareth-allan-bjss, line 10: It might get quite big, but now it's parameterised I think it would be nice if this test validated that the right colours were used for all label types in all groups. I think it's pretty key functionality and we'd want to identify it were accidentally changed.

frontend/fingertips-frontend/components/organisms/BenchmarkLabel/index.tsx
```diff
@@ -0,0 +1,195 @@
...
+
+export const BenchmarkTagStyle = styled(Tag)<{
+  legendType: BenchmarkLabelType;
+  group: BenchmarkLabelGroupType;
+}>(({ legendType, group }) => {
+  const theme = getDefaultBenchmarkTagStyle(group, legendType);
+  return {
+    padding: '5px 8px 4px 8px',
+    alignItems: 'center',
+    margin: '0.0225em',
```
gareth-allan-bjss, line 143: I think this margin value is too small. We're using `0.3125em` in the `TrendTag` and with this value as it is the benchmark labels look a lot closer together in the legend than on the design.

frontend/fingertips-frontend/components/organisms/BenchmarkLabel/__snapshots__/BenchmarkLabel.test.tsx.snap
gareth-allan-bjss: Actually, I don't think we should be using `capitalize`, as this results in `Not Compared` vs the design's `Not compared` (the start of the second word is capitalised. I think given you're doing the right thing with the case in the label values we should just use `unset` here.

frontend/fingertips-frontend/components/organisms/BenchmarkLabel/BenchmarkLabel.test.tsx
obarojohnson: I am trying to understand what you meant here—are you saying that we should explicitly test each colour separately within the function or on the component itself? I believe that, at the moment, colour is tested based on the function rather than the component, but I can add some tests to check whether the Label component has the correct colour. I suppose that is the suggestion right?

frontend/fingertips-frontend/components/organisms/BenchmarkLabel/BenchmarkLabel.test.tsx
```diff
@@ -0,0 +1,222 @@
...
+    [
+      BenchmarkLabelGroupType.RAG,
+      BenchmarkLabelType.LOWER,
+      {
+        backgroundColor: 'var(--other-light-blue, #5694CA)',
+        color: 'var(--other-black, #0B0C0C)',
+      },
+    ],
```
gareth-allan-bjss, line 18: You've got this scenario again lower down (lines 46 - 53).
```suggestion
```

frontend/fingertips-frontend/components/organisms/BenchmarkLabel/BenchmarkLabel.test.tsx
gareth-allan-bjss: Discussed offline - I was asking to cover all colour permutations in the existing test, which you've now done.

gareth-allan-bjss (approved):

PR #130, opened 2025-02-07
Feature/DHSCFT-313 Make e2e test stable

williamwillis11:
# Description

Jira ticket: [DHSCFT-313](https://bjss-enterprise.atlassian.net/browse/DHSCFT-313)

Added defensive code to stop playwright getting ahead of itself.

## Validation

e2e tests pass locally against local and docker

### DIFF
--- a/frontend/fingertips-frontend/playwright/page-objects/pages/resultsPage.ts
+++ b/frontend/fingertips-frontend/playwright/page-objects/pages/resultsPage.ts
@@ -25,19 +25,29 @@ export default class ResultsPage extends BasePage {
   }
 
   async checkURLIsCorrect(queryParams = '') {
-    await this.checkURL(
+    await this.checkURLMatches(
       `results?${SearchParams.SearchedIndicator}=${queryParams}`
     );
   }
 
-  async clickIndicatorCheckboxes(indicatorIds: string[]) {
-    for (const indicatorId of indicatorIds) {
-      const checkbox = this.page.getByTestId(
-        `${this.indicatorCheckboxPrefix}-${indicatorId}`
+  async selectIndicatorCheckboxes(indicatorIDs: string[]) {
+    for (const indicatorID of indicatorIDs) {
+      let checkbox = undefined;
+      checkbox = this.page.getByTestId(
+        `${this.indicatorCheckboxPrefix}-${indicatorID}`
       );
-      await checkbox.waitFor({ state: 'visible' });
+
+      await expect(checkbox).toBeAttached();
+      await expect(checkbox).toBeVisible();
+      await expect(checkbox).toBeInViewport();
       await expect(checkbox).toBeEnabled();
-      await checkbox.click({ delay: 100 });
+      await expect(checkbox).toBeEditable();
+      await expect(checkbox).toBeChecked({ checked: false });
+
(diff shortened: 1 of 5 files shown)


frontend/fingertips-frontend/playwright/page-objects/pages/resultsPage.ts
```diff
@@ -57,6 +69,30 @@ export default class ResultsPage extends BasePage {
...
+    const responsePromise = this.page.waitForResponse(
+      (response) =>
+        response
+          .url()
+          .includes(
+            `results?${SearchParams.SearchedIndicator}=${searchTerm}`
+          ) && response.status() === 200
+    );
+
+    await this.page.getByTestId(this.indicatorSearchButton).click();
```
gareth-allan-bjss, line 85: Not a major thing, but this could just be a call to `clickIndicatorSearchButton`.

frontend/fingertips-frontend/playwright/page-objects/pages/resultsPage.ts
```diff
@@ -25,17 +24,30 @@ export default class ResultsPage extends BasePage {
...
   }
 
-  async checkURLIsCorrect(indicator: string) {
-    await this.checkURL(
-      `results?${SearchParams.SearchedIndicator}=${indicator}`
+  async checkURLIsCorrect(queryParams = '') {
+    await this.checkURLMatches(
+      `results?${SearchParams.SearchedIndicator}=${queryParams}`
     );
   }
```
gareth-allan-bjss, line 31: Not necessarily something we have to do in this PR, but I think this highlights that this method is a bit odd. We're now passing more values in than the expected searched indicator value, but relying on the calling method to build the URL parameters. We should probably have separate parameters for the indicator and whatever other parameters we want to check.

frontend/fingertips-frontend/playwright/page-objects/pages/resultsPage.ts
```diff
@@ -25,17 +24,30 @@ export default class ResultsPage extends BasePage {
...
-    for (const indicatorId of indicatorIds) {
-      await this.page
-        .getByTestId(`${this.indicatorCheckboxPrefix}-${indicatorId}`)
-        .click();
+  async selectIndicatorCheckboxes(indicatorIDs: string[]) {
+    for (const indicatorID of indicatorIDs) {
+      let checkbox = undefined;
+      checkbox = this.page.getByTestId(
+        `${this.indicatorCheckboxPrefix}-${indicatorID}`
+      );
```
gareth-allan-bjss, line 38: Can this just be:
```suggestion
      const checkbox = this.page.getByTestId(
        `${this.indicatorCheckboxPrefix}-${indicatorID}`
      );
```

gareth-allan-bjss (approved):

frontend/fingertips-frontend/playwright/page-objects/pages/resultsPage.ts
williamwillis11: yes that was actually just left over when i wanted to make sure checkbox was getting set back to undefined each time. but later in the investigation realised that this wasnt the issue so ill make this change to remove the unnecessary setting it to undefined in each call

PR #136, opened 2025-02-10
DHSCFT-258: Only return health data for all ages and sexes

gareth-allan-bjss:
# Description

Jira ticket: [DHSCFT-258](https://bjss-enterprise.atlassian.net/browse/DHSCFT-258)

Updates the health data endpoint so that it only returns data points that relate to all sexes and all ages. It is currently returning multiple data points for each time period, split by the age and sex dimension. This is not the desired behaviour when doing a "basic" query, such as that used to retrieve data for the line chart.

**Note:** This change means that the population pyramid is no longer shown, as it relies on the inequalities data being returned for the special indicators that it uses. Once support has been added to the API for requesting this data (this should be done as part of [DHSCFT-312](https://bjss-enterprise.atlassian.net/browse/DHSCFT-312)) the population pyramid can be updated to request the appropriate data again (likely under [DHSCFT-148](https://bjss-enterprise.atlassian.net/browse/DHSCFT-148)).

## Changes

- Add a `HasValue` column to the `AgeDimension` table to indicate whether a row has a value or is an "all ages" value. This matches the approach used in the `SexDimension` table
- Made the `HealthDataRepository` return only data points which have age and sex dimensions with the `HasValue` flag set to false. This is to filter out "inequalities" data from the default health data endpoint response. In future tasks that add support for inequalities the intention is to add query parameters to the endpoint to allow you to request these values to be included, at which point the repository query will be extended to conditionally use the where clauses being added here.
- Made the frontend's Mock Service Worker handler filter the indicator data it was returning by the selected area codes even when the indicator ID didn't match. It's currently only filtering when the indicator ID matches.
- Removed the check for the population pyramid from the E2E tests

## Validation
The chart page when running against mocks:
Screenshot 2025-02-10 at 15 18 12

The chart page when running the whole application as part of the Docker compose command:
Screenshot 2025-02-10 at 16 05 27

API results before this change (note the "sex" field values):
Screenshot 2025-02-10 at 16 15 17

API results for the same indicator after this change (again, see the "sex" field values):
Screenshot 2025-02-10 at 16 12 35

Running the E2E tests against Docker:
Screenshot 2025-02-10 at 16 09 42


database/fingertips-db/Tables/AgeDimension.sql
```diff
@@ -1,9 +1,10 @@
 ﻿--The age (range) of the population that the data is for
 CREATE TABLE [dbo].[AgeDimension](
-	[AgeKey] [smallint] IDENTITY(1,1) NOT NULL, --surrogate key of the age
-	[Name] [nvarchar](50) NOT NULL, 			-- the name of the age e.g. over 75		
-	[AgeID] [smallint] NOT NULL, 				--the age id, taken from the original Fingertips Pholio database
- CONSTRAINT [PK_AgeDimension] PRIMARY KEY CLUSTERED 
+	[AgeKey] [smallint] IDENTITY(1,1) NOT NULL, -- Surrogate key of the age
+	[Name] [nvarchar](50) NOT NULL,             -- The name of the age e.g. over 75
+	[AgeID] [smallint] NOT NULL,                -- The age id, taken from the original Fingertips Pholio database
+	[HasValue] [bit] NOT NULL,                  -- A flag to indicate if the data has a value (if not it is for all persons)
```
HowardMayBjss, line 6: As you have identified, this will fail because it can't be NULL and there is existing data which has not value for this new column. A workaround is to delete the data from the table before updating the schema.

HowardMayBjss: Alternately delete the tables manually before deploying!

steves-bjss (approved):

HowardMayBjss (approved):

PR #138, opened 2025-02-10
[DHSCFT-137] one indicator two areas

adawari-thomas-bjss:
# Description

Jira ticket: [DHSCFT-137](https://bjss-enterprise.atlassian.net/browse/DHSCFT-137)

## Changes

- Updated line chart table to display data for 2 areas
- Added change to prevent line chart and line chart table from being displayed when more than 1 indicator is selected
- Added change to prevent line chart and line chart table from being displayed when more than 2 areas are selected
- Added change to prevent line chart and line chart table from being displayed when each selected area has more than 1 health data point
- Updated playwright tests

## Validation

- border lines to be removed when scroll is added to table

2 Areas
image

Single area
image


HowardMayBjss:
I can't find the design for this in Figma - could you give some direction as to where it is please :)

adawari-thomas-bjss:
> I can't find the design for this in Figma - could you give some direction as to where it is please :)

There isn't a design for this in Figma. This ticket updates the table to accept multiple areas, a subsequent ticket will add a scroll bar to display a single area at a time.

frontend/fingertips-frontend/playwright/page-objects/pages/chartPage.ts
```diff
@@ -24,17 +24,20 @@ export default class ChartPage extends BasePage {
     await this.page.getByTestId(this.backLink).click();
   }
 
-  async checkChartAndChartTable() {
-    await expect(this.page.getByTestId(this.lineChartComponent)).toBeVisible();
-    await expect(
-      this.page.getByTestId(this.lineChartTableComponent)
-    ).toBeVisible();
+  async checkChartAndChartTable(isMultipleIndicators: boolean) {
```
williamwillis11, line 27: am going to suggest a change to this

frontend/fingertips-frontend/components/pages/chart/chart.test.tsx
```diff
@@ -99,3 +99,42 @@ it('should not render the scatterChart component when only 1 indicator is select
 
   expect(scatterChart).not.toBeInTheDocument();
 });
+
+describe('should not display line chart', () => {
+  it('should not display line chart and line chart table when multiple indicators are selected', () => {
+    render(
+      <Chart data={[mockHealthData['1']]} indicatorsSelected={['0', '1']} />
```
gareth-allan-bjss, line 106: You're passing the same `data` prop here that you're passing in `should not display line chart and line chart table when more than 2 area codes are selected`, so presumably it has more than 2 area codes. i.e. if you removed the check for the number of indicators selected this test would still pass, wouldn't it?

frontend/fingertips-frontend/components/organisms/LineChartTable/index.tsx
```diff
@@ -163,75 +138,128 @@ const getCellHeader = (heading: string, index: number): ReactNode => {
...
+            <Table.Row>
+              {data.map((area, index) => (
+                <React.Fragment key={area.areaName}>
+                  <Table.CellHeader
+                    colSpan={getAreaNameCellSpan(index)}
+                  ></Table.CellHeader>
+                  <StyledAreaNameHeader>{area.areaName}</StyledAreaNameHeader>
+                  <Table.CellHeader colSpan={2}></Table.CellHeader>
+                </React.Fragment>
+              ))}
```
gareth-allan-bjss, line 187: This results in longer area names displaying strangely:
Screenshot 2025-02-11 at 16 25 33
I appreciate the designs _do_ look like this, but Leonie has raised the point that it doesn't look right. I'm double-checking that we definitely want to deviate from the designs, but I think it makes sense.

I think if we were to change this to something like the following it would allow us to have longer area names display nicely:
```suggestion
              {data.map((area, index) => (
                <React.Fragment key={area.areaName}>
                  {index === 0 ? <Table.CellHeader /> : null}
                  <StyledAreaNameHeader colSpan={5}>
                    {area.areaName}
                  </StyledAreaNameHeader>
                </React.Fragment>
              ))}
```

frontend/fingertips-frontend/components/organisms/LineChartTable/LineChartTable.test.tsx
```diff
@@ -5,121 +5,232 @@ import {
...
+    it('should have grey cell color for benchmark column', () => {
+      const benchmarkValueIndex = Object.values(
+        LineChartTableHeadingEnum
+      ).findIndex(
+        (value) => value === LineChartTableHeadingEnum.BenchmarkValue
+      );
+
+      render(
+        <LineChartTable
+          data={MOCK_HEALTH_DATA}
```
gareth-allan-bjss, line 71: Should this be `MOCK_HEALTH_DATA[0]`?

frontend/fingertips-frontend/playwright/tests/search_results_and_view_chart.spec.ts
```diff
@@ -88,6 +92,38 @@ test.describe('Search via indicator', () => {
       await homePage.checkSearchFieldIsPrePopulatedWith(searchTerm);
     });
 
+    await test.step('Select single indicator and view charts', async () => {
+      await homePage.clearSearchIndicatorField();
+      await homePage.typeIndicator(searchTerm);
+      await homePage.clickSearchButton();
+
+      await resultsPage.checkURLIsCorrect(searchTerm);
```
gareth-allan-bjss, line 100: We use `waitForURLToContain` on the first run through the journey. Is that to avoid flakiness (that this might reintroduce?).

frontend/fingertips-frontend/playwright/page-objects/pages/chartPage.ts
```diff
@@ -2,13 +2,30 @@ import { SearchParams } from '@/lib/searchStateManager';
...
+export enum IndicatorMode {
+  ONE_INDICATOR = 'ONE_INDICATOR',
+  TWO_INDICATORS = 'TWO_INDICATORS',
+  MULTIPLE_INDICATORS = 'MULTIPLE_INDICATORS', // 3+ indicators
+}
+
+export enum AreaMode {
+  ONE_AREA = 'ONE_AREA',
+  TWO_AREAS = 'TWO_AREAS',
+  THREE_AREAS = 'THREE_AREAS',
```
gareth-allan-bjss, line 14: Probably worth making it clear that this is 3+ areas.

frontend/fingertips-frontend/playwright/page-objects/pages/chartPage.ts
```diff
@@ -24,17 +41,91 @@ export default class ChartPage extends BasePage {
...
+        `${IndicatorMode.ONE_INDICATOR}_${AreaMode.ENGLAND_AREA}`,
+        singleIndicatorConfig,
+      ],
+      [
+        `${IndicatorMode.TWO_INDICATORS}_${AreaMode.ENGLAND_AREA}`,
+        { visibleComponents: defaultVisible, hiddenComponents: defaultHidden },
+      ],
+    ]);
+
+    const config = scenarioConfigs.get(`${indicatorMode}_${areaMode}`);
```
gareth-allan-bjss, line 92: This isn't a must-change thing, but could you use a nested map (i.e. indicator number as the first dimension and areas as the second) instead? It feels a bit odd to create a string to use as the key. I suppose it might be easier to understand though.

frontend/fingertips-frontend/playwright/tests/search_results_and_view_chart.spec.ts
williamwillis11: good spot

frontend/fingertips-frontend/playwright/page-objects/pages/chartPage.ts
williamwillis11: don't love my approach either but thinking of (once the pick area stuff is in) changing the e2e test to use a different approach like:

https://playwright.dev/docs/test-parameterize
or
https://playwright.dev/docs/test-parameterize#create-tests-via-a-csv-file
or from a json file 

so the same test loops for the three agreed scenarios we are going to cover - i need to think more on this but thinking we get Adas in and revisit

frontend/fingertips-frontend/playwright/tests/search_results_and_view_chart.spec.ts
```diff
@@ -88,6 +92,38 @@ test.describe('Search via indicator', () => {
...
+      await chartPage.clickBackLink();
+      await resultsPage.checkURLIsCorrect(
+        `${searchTerm}&${SearchParams.IndicatorsSelected}=${indicatorIDs[0]}`
+      );
+
+      await resultsPage.clickBackLink();
+
+      await homePage.checkOnHomePage();
+    });
+
```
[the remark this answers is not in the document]
williamwillis11: agreed - the coverage at the e2e level is a mess and i need to straighten it out:

https://github.com/dhsc-govuk/FingertipsNext/pull/138#discussion_r1952273607

frontend/fingertips-frontend/components/pages/chart/chart.test.tsx
adawari-thomas-bjss: the success of the test now depends on the `areasSelected` prop that's just been added, instead of the `data` prop.

frontend/fingertips-frontend/playwright/page-objects/pages/chartPage.ts
gareth-allan-bjss: OK, sounds reasonable. Happy to leave this as-is in this PR.

PR #142, opened 2025-02-11
Fix container push conditions

epercy-bjss:
# Description

Jira ticket: [DHSCFT-318](https://bjss-enterprise.atlassian.net/browse/DHSCFT-318)

Ensure container push runs with push and PR changes

## Changes

- Add conditions to the container push stages
- Ensure that deployments only happen from the `main` branch

## Validation

Worklow run manually from branch successfully

### DIFF
--- a/.github/workflows/fingertips-workflow.yml
+++ b/.github/workflows/fingertips-workflow.yml
@@ -209,6 +209,7 @@ jobs:
     uses: ./.github/workflows/build-container.yml
     needs: [check-changes, determine-version, e2e-tests-local]
     if: >
+      always() && !failure() && !cancelled() &&
       github.event_name == 'workflow_dispatch' ||
       needs.check-changes.outputs.frontend_changed == 'true'
     with:
@@ -224,6 +225,7 @@ jobs:
     uses: ./.github/workflows/build-container.yml
     needs: [check-changes, determine-version, e2e-tests-local]
     if: >
+      always() && !failure() && !cancelled() &&
       github.event_name == 'workflow_dispatch' ||
       needs.check-changes.outputs.frontend_changed == 'true'
     with:
--- a/.github/workflows/fingertips-workflow.yml
+++ b/.github/workflows/fingertips-workflow.yml
@@ -249,6 +249,7 @@ jobs:
         working-directory: ./database/fingertips-db
     if: >
       always() && !failure() && !cancelled() &&
+      github.ref_name == 'main' &&
       github.event_name != 'pull_request' &&
       (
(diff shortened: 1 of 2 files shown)


.github/workflows/fingertips-workflow.yml
```diff
@@ -242,8 +259,10 @@ jobs:
       run:
         working-directory: ./database/fingertips-db
     if: >
-      always() && !failure() && !cancelled() &&
+      always() &&
+      contains(needs.*.result, 'success') &&
       github.event_name != 'pull_request' &&
+      github.ref_name == 'main' &&
```
gareth-allan-bjss, line 265: Howcome the other jobs doing this are using the following, but not this one?:
```
        github.head_ref == 'main' ||
        github.ref_name == 'main'
```

.github/workflows/fingertips-workflow.yml
```diff
@@ -450,17 +484,24 @@ jobs:
...
+      always() &&
+      contains(needs.*.result, 'success') &&
       github.event_name != 'pull_request' &&
-      github.event_name == 'workflow_dispatch' ||
       (
-        needs.check-changes.outputs.db_changed == 'true' ||
-        needs.check-changes.outputs.api_changed == 'true' ||
-        needs.check-changes.outputs.frontend_changed == 'true' ||
-        needs.check-changes.outputs.search_changed == 'true'
+        github.ref_name == 'main' &&
```
gareth-allan-bjss, line 491: This one's not referencing `github.head_ref` either. Is that deliberate?

.github/workflows/fingertips-workflow.yml
epercy-bjss: Because it cannot run from a pull request. The branch name ref is different on pull_request triggers than it is on 'push' triggers. As we are explicitly preventing pull_requests for this job, it's not needed. I've actually seen a couple more I can remove this from, thanks for pointing it out.

.github/workflows/fingertips-workflow.yml
epercy-bjss: Again, as per the above. The condition to look at branch names should only be:

      `(
        github.head_ref == 'main' ||
        github.ref_name == 'main'
      ) `
      
 For the API and Frontend deployment as we want the job to run Terraform PR validation for pull requests

.github/workflows/fingertips-workflow.yml
epercy-bjss: Pushed a change to remove a further reference to the PR main branch

PR #120, opened 2025-02-04
DHSCFT-290: show areas

alibombaye:
# Description

Jira ticket: [DHSCFT-290](https://bjss-enterprise.atlassian.net/browse/DHSCFT-290)
Screenshot 2025-02-14 at 14 50 08

- Shows the groupTypes associated with the areaType selected
- Shows the available areas based upon the areaType selected
- Handles change to API contract areaTypeKey and new schema changes
- Updates mock data
- Extract the IndicatorSelection form to its own component.
- Updates the validation on IndicatorSearchForm.
- Handles passing all search state when submitting indicator selection.

## Validation

Unit tests passing
E2E tests passing against MSW
Tested manually against MSW
Tested manually against API (https://github.com/dhsc-govuk/FingertipsNext/pull/158) Working as expected.
E2E tests against API (https://github.com/dhsc-govuk/FingertipsNext/pull/158) Working as expected.

PR #144, opened 2025-02-12
Feature/dhscft 139 line chart with confidence intervals

priya-thorpe-bjss:
# Description

Jira ticket: [dhscft 139](https://bjss-enterprise.atlassian.net/browse/DHSCFT-139)

## Changes

- Removed 'search' from error page when clicking the back button and updated tests
- Created a reusable confidenceIntervalCheckbox component
- Displaying error bars when the checkbox is clicked
- Updated confidence intervals in mock data

## Validation
- Manual validation
- Tests pass

Screenshot 2025-02-13 at 16 46 06


frontend/fingertips-frontend/components/organisms/LineChart/index.tsx
HowardMayBjss, line 1: I've had a look at this running and noticed that the confidence bars are all being added to the one line as shown here 
image

frontend/fingertips-frontend/components/molecules/ConfidenceIntervalCheckbox/index.tsx
HowardMayBjss, line 1: The styling of this component looks like it needs a little more refinement 
image compared to the design 
image

HowardMayBjss: I recognise that the positioning of the checkbox between the 'area selectors' and the actual chart may not be made possible by the High charts library. Again might be worth a chat with BA / Leonie

gareth-allan-bjss: We agreed as part of kick-off that we'd show the checkbox above the chart, as it's a simpler implementation.

frontend/fingertips-frontend/lib/searchStateManager.ts
```diff
@@ -143,6 +152,43 @@ export class SearchStateManager {
     const searchStateManager = new SearchStateManager(newState);
     return searchStateManager;
   }
+  public static setStateFromSearchStateParams(params?: SearchStateParams) {
```
alibombaye, line 155: This is not needed for now so should be good to remove.

frontend/fingertips-frontend/components/molecules/ConfidenceIntervalCheckbox/confidenceIntervalCheckbox.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+      />
+    );
+    expect(screen.getByRole('checkbox')).not.toBeChecked();
+
+    const confidenceIntervalBars = container.getElementsByClassName(
+      'highcharts-errorbar-series'
+    );
+
+    expect(confidenceIntervalBars).toHaveLength(0);
+  });
```
gareth-allan-bjss, line 76: I don't think these tests are validating what they say they are validating. How are there error bars if you're just rendering the checkbox?

frontend/fingertips-frontend/components/molecules/ConfidenceIntervalCheckbox/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+  chartName: string;
+  showConfidenceIntervalsData: boolean;
+  searchState?: SearchStateParams;
+};
+
+export function ConfidenceIntervalCheckbox({
+  chartName,
+  showConfidenceIntervalsData,
+}: Readonly<ConfidenceIntervalCheckboxProps>) {
+  const searchParams = useSearchParams();
```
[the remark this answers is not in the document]
gareth-allan-bjss: Presumably we could do the same with the path as well?

frontend/fingertips-frontend/components/molecules/ConfidenceIntervalCheckbox/confidenceIntervalCheckbox.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+    expect(screen.getByRole('checkbox')).toBeChecked();
+
+    expect(mockReplace).toHaveBeenCalledWith(
+      `${mockPath}?${SearchParams.ConfidenceIntervalSelected}=example+chart`,
+      {
+        scroll: false,
+      }
+    );
+  });
+});
```
gareth-allan-bjss, line 95: Can we have a test that validates starting with `showConfidenceIntervalsData={true}` and unchecking the checkbox results in the correct URL too please?

frontend/fingertips-frontend/components/organisms/LineChart/index.tsx
```diff
@@ -31,19 +35,33 @@ export function LineChart({
...
+  const searchParams = useSearchParams();
+
+  const existingParams = new URLSearchParams(searchParams);
+  const searchStateManager =
+    SearchStateManager.setStateFromParams(existingParams);
+
+  const showConfidenceIntervalsData =
+    searchStateManager.getSearchState()[
+      SearchParams.ConfidenceIntervalSelected
+    ];
```
gareth-allan-bjss, line 47: I think it would make sense to pass `showConfidenceIntervalsData` into this chart as a prop, so that the chart doesn't have to deal with the URL query parameters/search state manager, it just has to know what chart name to look for in the list. It would make it easier to test as well, since you wouldn't have to mock `useSearchParams`.

frontend/fingertips-frontend/components/organisms/LineChart/index.tsx
```diff
@@ -79,6 +97,10 @@ export function LineChart({
   return (
     <div data-testid="lineChart-component">
       <H3>{lineChartTitle}</H3>
+      <ConfidenceIntervalCheckbox
+        chartName="lineChart"
```
gareth-allan-bjss, line 101: Can we make the value passed to `chartName` a constant in this file please? Then you can use it in the function that filters `showConfidenceIntervalsData` and there's no risk of the two values becoming out of sync.

frontend/fingertips-frontend/components/organisms/LineChart/lineChart.test.tsx
gareth-allan-bjss, line 1: It's probably worth having a test here to validate that if we pass in the correct "show confidence intervals" parameter that the line chart deals with it correctly and shows the checkbox (and that it doesn't show it if it doesn't).

frontend/fingertips-frontend/components/organisms/LineChart/lineChartHelpers.test.ts
```diff
@@ -207,11 +282,90 @@ describe('generateSeriesData', () => {
   it('should repeat symbols when there are more series than symbols', () => {
```
gareth-allan-bjss, line 282: We should have an equivalent of this test to verify we're repeating colours when we run out too, as we're now responsible for that logic, rather than Highcharts.

frontend/fingertips-frontend/components/organisms/LineChart/lineChartHelpers.test.ts
```diff
@@ -207,11 +282,90 @@ describe('generateSeriesData', () => {
...
+      undefined,
+      false
+    ) as SeriesLineOptions[];
 
     expect(generatedSeriesData[0].marker?.symbol).toBe('arc');
-    expect(generatedSeriesData[1].marker?.symbol).toBe('circle');
-    expect(generatedSeriesData[2].marker?.symbol).toBe('arc');
+    expect(generatedSeriesData[1].marker?.symbol).toBe(undefined);
+    expect(generatedSeriesData[2].marker?.symbol).toBe('circle');
+    expect(generatedSeriesData[3].marker?.symbol).toBe(undefined);
```
gareth-allan-bjss, line 296: This is no longer verifying that the symbols are repeated, as you're not checking the values for the last series.

It might also be worth a comment to explain why some series have no symbols (they're the `errorbar` series).

frontend/fingertips-frontend/components/molecules/result/index.tsx
```diff
@@ -92,6 +92,7 @@ export function SearchResult({
             name="indicator"
             value={result.indicatorId}
             defaultChecked={indicatorSelected}
+            sizeVariant="SMALL"
```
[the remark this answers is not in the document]
gareth-allan-bjss: Wait, has this been applied to the wrong checkbox? Should it have been applied to the one in `frontend/fingertips-frontend/components/molecules/ConfidenceIntervalCheckbox/index.tsx`?

frontend/fingertips-frontend/components/molecules/ConfidenceIntervalCheckbox/index.tsx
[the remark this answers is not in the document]
alibombaye: I have discussed this with Priya and think it's one to do later and also after https://github.com/dhsc-govuk/FingertipsNext/pull/120 has been merged in which does this for the results page. 

But yeah that's what that searchState prop is for. That I think can possibly removed for this PR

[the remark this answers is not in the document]
gareth-allan-bjss: OK, I wondered if this was a timing thing. I think it's fair enough to wait then. @priya-thorpe-bjss we should perhaps put a ticket in to make sure we do this though, as I think it is something we want to get to.

frontend/fingertips-frontend/components/organisms/LineChart/lineChartHelpers.ts
```diff
@@ -1,22 +1,43 @@
...
+
+      const confidenceIntervalSeries: Highcharts.SeriesOptionsType = {
+        type: 'errorbar',
+        name: `${item.areaName}`,
+        data: item.healthData.map((point) => [
+          point.year,
+          point.lowerCi,
+          point.upperCi,
+        ]),
+        visible: showConfidenceIntervalsData,
```
gareth-allan-bjss, line 32: Can we do anything with the styling of the error bars? Compare the current implementation:
Screenshot 2025-02-14 at 17 00 01

With the designs:
Screenshot 2025-02-14 at 17 00 11

I think we need to try and make the error bars the correct shade of grey, make the horizontal lines narrower and possibly make all lines a bit thicker.

gareth-allan-bjss: Looks like there's a [`color` property](https://api.highcharts.com/highcharts/plotOptions.errorbar.color), a [`whiskerLength` property](https://api.highcharts.com/highcharts/plotOptions.errorbar.whiskerLength) and a [`lineWidth` property](https://api.highcharts.com/highcharts/plotOptions.errorbar.lineWidth) that should do the trick.

See https://www.highcharts.com/docs/chart-and-series-types/error-bar-series#styling-the-error-bars & https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/plotoptions/error-bar-styling/.

gareth-allan-bjss: Something like this looks reasonable to me:
```suggestion
        type: 'errorbar',
        name: `${item.areaName}`,
        data: item.healthData.map((point) => [
          point.year,
          point.lowerCi,
          point.upperCi,
        ]),
        visible: showConfidenceIntervalsData,
        color: '#B1B4B6',
        whiskerLength: '20%',
        lineWidth: 2,
```

frontend/fingertips-frontend/components/molecules/result/index.tsx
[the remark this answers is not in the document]
priya-thorpe-bjss: Yes, wrong one

frontend/fingertips-frontend/components/molecules/ConfidenceIntervalCheckbox/index.tsx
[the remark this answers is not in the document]
priya-thorpe-bjss: Yes, we can put a ticket in. I have removed the searchState prop.

frontend/fingertips-frontend/components/molecules/ConfidenceIntervalCheckbox/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+  const pathname = usePathname();
+  const { replace } = useRouter();
+  const params = new URLSearchParams(searchParams);
+
+  const handleClick = (chartName: string, checked: boolean) => {
+    const searchState = SearchStateManager.setStateFromParams(params);
```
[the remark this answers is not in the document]
priya-thorpe-bjss: Yes, I have moved it out of the function

frontend/fingertips-frontend/components/organisms/LineChart/lineChartHelpers.ts
priya-thorpe-bjss: Thanks

frontend/fingertips-frontend/components/organisms/LineChart/lineChartHelpers.test.ts
priya-thorpe-bjss: ah yes, missed the last one

frontend/fingertips-frontend/components/organisms/LineChart/index.tsx
gareth-allan-bjss: Agreed that we'll do it in the follow-up ticket already discussed, as the cleanest option is to pull all the URL handling up to the chart page itself.

frontend/fingertips-frontend/components/organisms/LineChart/lineChart.test.tsx
gareth-allan-bjss: I think this is still valid. We should have a test that validates the checkbox is checked if the correct URL query param is passed in, and one that validates it isn't if no query param or a different value is passed in.

I think I may have missed this point when we talked earlier, apologies.

frontend/fingertips-frontend/components/organisms/LineChart/lineChart.test.tsx
```diff
@@ -33,3 +49,22 @@ it('should render the LineChart title', () => {
...
+  expect(screen.getByRole('checkbox')).toBeChecked();
+});
+
+it('should validate checkbox is not checked when passed in an incorrect URL query param of randomChart', async () => {
+  const confidenceIntervalData = ['randomChart'];
+  render(<LineChart healthIndicatorData={mockHealthData[1]} />);
+
+  expect(confidenceIntervalData.some((ci) => ci === 'lineChart')).toBe(false);
+  expect(screen.getByRole('checkbox')).not.toBeChecked();
+});
```
gareth-allan-bjss, line 70: I don't think these are validating the behaviour of the line chart page. Your first expectation is just testing the behaviour of JS' `some` function, and the second is just validating that clicking/not clicking the checkbox results in it being selected. Neither of which are things we've implemented. 

I think you need to mock `useSearchParams`, return your `confidenceIntervalData` value from the correct URL parameter and then validate whether the checkbox is checked (without interacting with it in the test).

gareth-allan-bjss: Discussed offline. Given our next task is to refactor this code so that it isn't pulling query parameters directly from the URL I'm happy for you to just remove these tests here and add some equivalents back in once we've refactored.

gareth-allan-bjss (approved):

PR #150, opened 2025-02-13
Feature/dhscft 333 run e2e ci in docker

williamwillis11:
# Description

Jira ticket: [DHSCFT-333](https://bjss-enterprise.atlassian.net/browse/DHSCFT-333)

Run E2E tests in dockerised containers in CI instead of using localhost+Mock Service Worker.

## Changes

- Changed e2e-tests.yml to take an optional container parameter which if passed is used (for CD) else falls back to using docker (for CI)
- Cacheing dependencies now to speed up the (up to) 2 dependency installation steps
- new step to install the playwright dependencies (only needed for CI when we arnt using the container which already has them)
- added a new secret fingertips_e2e_db_password to house the db password which is used in new step below
- new step to create a .env file for db-setup (again only needed in CI for the docker image)
- Removed send-slack-on-fail parameter as its no longer needed
- make CI e2e tests run start as soon as check-changes is done as it can run at the same time now as the previous needs jobs
- Updated Readme

## Validation
 Ran successfully locally against mocks and docker
Ran CI successfully in docker - 
Will check CD on merge


epercy-bjss (commented):
I'd be greatful if you could solve your local IDE problem and change the double quotes back to single quotes. It is the preferred 'quote' in most cases. Otherwise LGTM.

HowardMayBjss (approved):

PR #156, opened 2025-02-14
Feature/dhscft 57 create more data

andyevans-bjss:
# Description

Jira ticket: dhscft 57

Add further data to the PoC data set, including indicator dates and category IDs needed for deprivation data

## Validation

Imported the resulting files into the star schema. Review of file contents


DataCreator/DataCreator/data/out/areas.csv
HowardMayBjss, line 1: Is there a reason to have changed this to a CSV - this will require consuming projects to be updated to take a csv rather than a json. Same for indicators.json

HowardMayBjss, line 1: There are only 117 areas in this file. We are still expecting there to be 1000's?

DataCreator/DataCreator/data/out/areas.json
HowardMayBjss, line 1: ICBs are missing parents

andyevans-bjss: I think this isn't an issue for now as the logic to insert into the Areas table will use children. If it becomes an issue we can go back and create a new PR to resolve

PR #157, opened 2025-02-17
Add support for missing values

epercy-bjss:
# Description

Jira ticket: [DHSCFT-350](https://bjss-enterprise.atlassian.net/browse/DHSCFT-350)

Update HealthData table’s schema and API to account for null values, and make the frontend application deal with the situation gracefully.

## Changes

- Update the HealthData table’s schema to mark the necessary fields as nullable
- Update the API’s Swagger definition to specify the necessary fields as optional
- Regenerate the frontend’s API clients
- Fix null references in Heatmap, PopulationData and PreparePopulationData

## Validation

Build, test, e2e all passing successfully. Testing the site locally seems to be the same experience as prior to the change.

### DIFF
--- a/frontend/fingertips-frontend/components/organisms/Heatmap/heatmapUtil.ts
+++ b/frontend/fingertips-frontend/components/organisms/Heatmap/heatmapUtil.ts
@@ -30,17 +30,16 @@ function getValueForAreaCode(
   areaCode: string,
   indicatorRowData: IndicatorRowData
 ): number | null {
-  let result = null;
   for (const healthDataForArea of indicatorRowData.rowData) {
     if (healthDataForArea.areaCode === areaCode) {
       for (const dataPoint of healthDataForArea.healthData) {
         if (dataPoint.year === indicatorRowData.year) {
-          result = dataPoint.value;
-          break;
+          // Coerce undefined to null
+          return dataPoint.value !== undefined ? dataPoint.value : null;
         }
       }
       break;
     }
   }
-  return result;
+  return null;
 }
--- a/frontend/fingertips-frontend/components/organisms/Heatmap/heatmapUtil.ts
+++ b/frontend/fingertips-frontend/components/organisms/Heatmap/heatmapUtil.ts
@@ -29,17 +29,16 @@ export function generateHeatmapData(
 function getValueForAreaCode(
   areaCode: string,
   indicatorRowData: IndicatorRowData
-): number | null {
+): number | undefined {
   for (const healthDataForArea of indicatorRowData.rowData) {
     if (healthDataForArea.areaCode === areaCode) {
       for (const dataPoint of healthDataForArea.healthData) {
         if (dataPoint.year === indicatorRowData.year) {
-          // Coerce undefined to null
-          return dataPoint.value !== undefined ? dataPoint.value : null;
+          return dataPoint.value;
         }
       }
       break;
     }
   }
-  return null;
+  return undefined;
 }
--- a/frontend/fingertips-frontend/components/organisms/LineChartTable/index.tsx
+++ b/frontend/fingertips-frontend/components/organisms/LineChartTable/index.tsx
@@ -15,10 +15,10 @@ interface TableProps {
 
 interface LineChartTableRowData {
   period: number;
-  count: number;
-  value: number;
-  lower: number;
-  upper: number;
+  count: number | null;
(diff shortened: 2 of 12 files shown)


frontend/fingertips-frontend/components/organisms/LineChartTable/index.tsx
```diff
@@ -15,10 +15,10 @@ interface TableProps {
...
 interface LineChartTableRowData {
   period: number;
-  count: number;
-  value: number;
-  lower: number;
-  upper: number;
+  count: number | null;
+  value: number | null;
+  lower: number | null;
+  upper: number | null;
```
gareth-allan-bjss, line 21: Could this not use optional properties?
```suggestion
  count?: number;
  value?: number;
  lower?: number;
  upper?: number;
```

frontend/fingertips-frontend/components/organisms/LineChartTable/index.tsx
```diff
@@ -106,19 +106,22 @@ const StyledSpan = styled('span')({
...
   areaData.healthData.map((healthPoint) => ({
     period: healthPoint.year,
-    count: healthPoint.count,
-    value: healthPoint.value,
-    lower: healthPoint.lowerCi,
-    upper: healthPoint.upperCi,
+    count: healthPoint.count ?? null,
+    value: healthPoint.value ?? null,
+    lower: healthPoint.lowerCi ?? null,
+    upper: healthPoint.upperCi ?? null,
```
gareth-allan-bjss, line 112: Again, I think if you use optional properties you don't need this change.

frontend/fingertips-frontend/components/organisms/Heatmap/heatmapUtil.ts
```diff
@@ -30,17 +30,16 @@ function getValueForAreaCode(
...
 ): number | null {
-  let result = null;
   for (const healthDataForArea of indicatorRowData.rowData) {
     if (healthDataForArea.areaCode === areaCode) {
       for (const dataPoint of healthDataForArea.healthData) {
         if (dataPoint.year === indicatorRowData.year) {
-          result = dataPoint.value;
-          break;
+          // Coerce undefined to null
+          return dataPoint.value !== undefined ? dataPoint.value : null;
```
gareth-allan-bjss, line 38: This could use the nullish coalescing operator too, I think:
```suggestion
          return dataPoint.value ?? null;
```

gareth-allan-bjss: Although I think the better option might be to update the return type of this function to allow `undefined` instead of `null`.

frontend/fingertips-frontend/components/organisms/LineChartTable/index.tsx
epercy-bjss: Fixed as suggested

frontend/fingertips-frontend/components/organisms/LineChartTable/index.tsx
epercy-bjss: Reverted as suggested

frontend/fingertips-frontend/components/organisms/Heatmap/heatmapUtil.ts
epercy-bjss: Changed to return undefined and updated the tests to also expect 'undefined'

frontend/fingertips-frontend/components/organisms/LineChartTable/index.tsx
```diff
@@ -289,19 +289,19 @@ export function LineChartTable({
                   {sortedAreaData[index].count}
                 </StyledAlignRightTableCell>
                 <StyledAlignRightTableCell numeric>
-                  {convertToPercentage(sortedAreaData[index].value)}
+                  {convertToPercentage(sortedAreaData[index].value ?? null)}
```
gareth-allan-bjss, line 292: I think you could just make `convertToPercentage` accept an optional parameter (e.g. `const convertToPercentage = (value?: number): string => {`) and then you wouldn't have to coerce the value to `null` here.

epercy-bjss: Done

gareth-allan-bjss (approved):

PR #164, opened 2025-02-18
Feature/dhscft 345 parent health data on table

WolfieKnee:
# Description

Jira ticket: [DHSCFT-345](https://bjss-enterprise.atlassian.net/browse/DHSCFT-345)
Display area group data on lineChart and its table

## Changes
- Chart Page: determine parent area code and add to healthIndicatorData Fetch. Pass to Chart with parentAreaCode.
- Chart: separate out parent data and pass to LineChartTable and LineChart
- LineChartTable: conditionally render parent data as group
- LineChart: conditionally add group series

no e2e changes as overall journey behaviour is unchanged. 

## Validation
- Typecheck, linting, formatting and unit tests all pass
- same `npm run dev` output for (http://localhost:3000/chart?si=mortality&is=333&as=E06000047&as=E06000006&ats=counties-and-unitary-authorities&gts=regions&cis=lineChart) below 
image


frontend/fingertips-frontend/components/organisms/LineChart/lineChartHelpers.ts
```diff
@@ -39,6 +40,22 @@ export function generateSeriesData(
...
 
+  if (parentIndicatorData) {
+    const groupSeries: Highcharts.SeriesOptionsType = {
+      type: 'line',
+      name: `Group: ${parentIndicatorData.areaName}`,
+      data: parentIndicatorData.healthData.map((point) => [
+        point.year,
+        point.value,
+      ]),
+      color: 'green',
```
gareth-allan-bjss, line 51: Is this the same colour as used on the designs?

Actually, looking more closely, I'm not sure the `black` used for the England benchmark is quite right, as it appears the designs call for a very dark grey.

frontend/fingertips-frontend/components/organisms/LineChart/lineChartHelpers.test.ts
```diff
@@ -291,6 +292,132 @@ describe('generateSeriesData', () => {
...
+        lineWidth: 2,
+        whiskerLength: '20%',
+      },
+    ];
+
+    const generatedSeriesData = generateSeriesData(
+      mockData,
+      symbols,
+      chartColours,
+      undefined,
```
gareth-allan-bjss, line 419: Can we pass some benchmark data here. I think it's important to verify that the benchmark's series comes before the parent's in the returned array, otherwise they'll be in the wrong order on the chart's legend.

frontend/fingertips-frontend/components/organisms/LineChart/lineChartHelpers.ts
WolfieKnee: changed to align with GovukColours

frontend/fingertips-frontend/components/organisms/LineChart/lineChartHelpers.test.ts
WolfieKnee: Added in benchmark data

frontend/fingertips-frontend/app/chart/page.tsx
```diff
@@ -30,12 +30,31 @@ export default async function ChartPage(
   await connection();
 
   const indicatorApi = ApiClientFactory.getIndicatorsApiClient();
+  const areaApi = ApiClientFactory.getAreasApiClient();
+
+  let parentAreaCode: string | undefined;
+
+  if (indicatorsSelected.length === 1 && areaCodes.length <= 2) {
+    try {
+      const areaData = await areaApi.getArea({ areaCode: areaCodes[0] }); // DHSCFT-256 assumes one common parent
```
steves-bjss, line 39: Is it guaranteed that the parent area code is always the first element in the array?

frontend/fingertips-frontend/app/chart/page.tsx
```diff
@@ -30,12 +30,31 @@ export default async function ChartPage(
...
+      if (areaData?.parent) {
+        parentAreaCode = areaData.parent.code;
```
steves-bjss, line 41: This could be simplified with something like this:  parentAreaCode = areaData?.parent?.code;  

This will do the null checking for you and parentAreaCode will be set to null, even if areaData is null.

steves-bjss (approved):

frontend/fingertips-frontend/app/chart/page.tsx
WolfieKnee: This is not the parent area, just the first of the selected areas. Since the filter will only allow area from the same group (i.e. same parent) to be selected all the areas in areaSelected will have the same parent. Since it is possible to select just one area areaCodes[0] is used.

PR #165, opened 2025-02-18
Feature/dhscft 278 line chart in tab

Joe-Hendley:
https://bjss-enterprise.atlassian.net/browse/DHSCFT-278

put the line chart & table in tabs

can't test the tabs in jest - govuk-react uses some sort of css magic which jsdom doesn't handle properly and probably never will) 
https://github.com/testing-library/jest-dom/issues/444

on discussion with @gareth-allan-bjss , @williamwillis11  & @Nayira-BJSS, we've agreed that the combination of "are the charts there when they should be" unit tests and playwright snapshots down the line will cover this sufficiently.


frontend/fingertips-frontend/components/layouts/tabContainer/tabContainer.test.tsx
```diff
@@ -0,0 +1,42 @@
...
+    const titles = screen.getAllByRole('link');
+    titles.map((_, index) => {
+      expect(titles[index].textContent).toBe(expectedTitleText[index]);
+    });
+
+    const panels = screen.getAllByRole('paragraph');
+    panels.map((_, index) => {
+      expect(panels[index].textContent).toBe(expectedContentText[index]);
+    });
+  });
```
alibombaye, line 41: Do we want a test for when we click on the tab. Are we able to check the selected prop on the tab is correct?

frontend/fingertips-frontend/components/layouts/tabContainer/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+  const handleClick = (
+    event: React.MouseEvent<HTMLAnchorElement>,
+    index: number
+  ) => {
+    event.preventDefault();
+    setTabIndex(index);
+  };
+
+  return (
```
williamwillis11, line 23: can each tab have a unique data-testid?

williamwillis11:
Screenshot 2025-02-19 at 08 41 54
just checking if this navigation will work with the tab component? ie tab to the tabs and then use the arrow keys to navigate through them?

i know the tests for this would need to be at a higher level, just wondering :)

frontend/fingertips-frontend/components/layouts/tabContainer/tabContainer.test.tsx
Joe-Hendley: Unfortunately the `selected` prop is a TS construct that gets abstracted away into `display = none` or `display = block` - which doesn't get applied properly on JSDOM so we can't test it with Jest

Joe-Hendley:
> just checking if this navigation will work with the tab component? ie tab to the tabs and then use the arrow keys to navigate through them?
> 
> i know the tests for this would need to be at a higher level, just wondering :)

So, tab + arrow navigation doesn't work on these, it works with tab + enter but then tab continues you down the page.

frontend/fingertips-frontend/components/pages/chart/index.tsx
```diff
@@ -59,19 +60,33 @@ export function Chart({
...
-            benchmarkData={englandBenchmarkData}
-            xAxisTitle="Year"
-            accessibilityLabel="A line chart showing healthcare data"
-          />
-          <LineChartTable
-            healthIndicatorData={dataWithoutEngland}
-            englandBenchmarkData={englandBenchmarkData}
-          />
-        </>
+        <TabContainer
```
gareth-allan-bjss, line 63: Probably worth expanding the various tests in `frontend/fingertips-frontend/components/pages/chart/chart.test.tsx` that checks the line chart/it's table are in the page to verify the tabs are or aren't shown when expected.

frontend/fingertips-frontend/components/pages/chart/index.tsx
gareth-allan-bjss, line 1: The designs have the "See how the indicator has changed over time" title outside of the tab group (because it applies to the chart and the table), whereas you currently have it inside the chart's tab:
Screenshot 2025-02-19 at 09 21 41
Screenshot 2025-02-19 at 09 21 50

I think we're going to need a slight restructure of how the heading is shown.

Joe-Hendley: Good spot - that's what I get for assuming past me got the layout right

gareth-allan-bjss: Looks good now!

frontend/fingertips-frontend/components/layouts/tabContainer/index.tsx
```diff
@@ -9,7 +9,13 @@ interface TabItem {
   content: JSX.Element;
 }
 
-export const TabContainer = ({ items }: { items: TabItem[] }) => {
+export const TabContainer = ({
+  id = '',
+  items,
+}: {
+  id?: string;
```
gareth-allan-bjss, line 16: Should the `id` just be a mandatory property? That way we can guarantee the `data-testid`s will always be unique (unlike if there is no `id` passed currently).

gareth-allan-bjss (approved):

williamwillis11 (approved):

gareth-allan-bjss (approved):

PR #166, opened 2025-02-18
[DHSCFT-312] update api to accept inequalities query parameter

adawari-thomas-bjss:
# Description

Jira ticket: [DHSCFT-312](https://bjss-enterprise.atlassian.net/browse/DHSCFT-312)

## Changes

- Included the `inequalities` query parameter to the indicator health data endpoint
- filtered the data being returned by the endpoint to include inequality data depending on the query parameter passed in the url

## Validation

age band data being returned when the age inequality is passed as a param

image


api/DHSC.FingertipsNext.Modules.HealthData.Repository/HealthDataRepository.cs
```diff
@@ -12,14 +12,14 @@ public HealthDataRepository(HealthDataDbContext healthDataDbContext)
...
-    public async Task<IEnumerable<HealthMeasureModel>> GetIndicatorDataAsync(int indicatorId, string[] areaCodes, int[] years)
+    public async Task<IEnumerable<HealthMeasureModel>> GetIndicatorDataAsync(int indicatorId, string[] areaCodes, int[] years, string[] inequalities)
     {
         return await _dbContext.HealthMeasure
             .Where(hm => hm.IndicatorDimension.IndicatorId == indicatorId)
             .Where(hm => areaCodes.Length == 0 || areaCodes.Contains(hm.AreaDimension.Code))
             .Where(hm => years.Length == 0 || years.Contains(hm.Year))
-            .Where(hm => !hm.SexDimension.HasValue)
-            .Where(hm => !hm.AgeDimension.HasValue)
+            .Where(hm => inequalities.Contains("sex") || !hm.SexDimension.HasValue)
```
MonkeyDate, line 21: what is the intended behaviour if no inequalities are supplied?

adawari-thomas-bjss: in that case, the `inequalities.Contains` check is false, so the query defaults to its original behaviour and doesn't return inequality data

api/DHSC.FingertipsNext.Modules.HealthData/Controllers/V1/IndicatorsController.cs
```diff
@@ -29,12 +30,14 @@ public class IndicatorsController(IIndicatorsService indicatorsService)
     public async Task<IActionResult> GetIndicatorDataAsync(
         [FromRoute] int indicatorId,
         [FromQuery] string[]? area_codes = null,
-        [FromQuery] int[]? years = null)
+        [FromQuery] int[]? years = null,
+        [FromQuery] string[]? inequalities = null)
```
gareth-allan-bjss, line 34: Not a major issue (and you're just following the existing pattern) but why set the default value for these optional properties to `null` and then have logic to convert that to `[]` below? Why not just set the default to `[]`?

api/DHSC.FingertipsNext.Modules.HealthData.UnitTests/Controllers/V1/IndicatorsControllerTests.cs
```diff
@@ -30,7 +30,8 @@ await _indicatorService
             .GetIndicatorDataAsync(
                 1,
                 ArgEx.IsEquivalentTo<string[]>(["ac1", "ac2"]),
-                ArgEx.IsEquivalentTo<int[]>([1999, 2024])
+                ArgEx.IsEquivalentTo<int[]>([1999, 2024]),
+                []
```
gareth-allan-bjss, line 34: This test is intended to cover the scenario where all parameters are specified and make sure they are passed through correctly. You should be passing some `inequalities` values into the controller here and validating it is passed to the indicator service as expected.

api/DHSC.FingertipsNext.Modules.HealthData.UnitTests/Services/IndicatorServiceTests.cs
```diff
@@ -62,7 +62,8 @@ public async Task GetIndicatorData_PassesFirst10DistinctFiltersToRepository_When
         await _indicatorService.GetIndicatorDataAsync(
             1,
             inputAreaCodes,
-            inputYears
+            inputYears,
+            []
```
gareth-allan-bjss, line 66: I think this scenario also needs to pass in some `inequalities` values and verify they're passed through as expected.

It also raises the question of whether we need to apply the same `Distinct()`/`Take(10)` filter to these parameters too?

api/DHSC.FingertipsNext.Modules.HealthData.UnitTests/Repository/HealthDataRepositoryTests.cs
```diff
@@ -270,6 +268,120 @@ public async Task Repository_ShouldOnlyIncludeResultsWithoutAnAgeDimensionValue(
...
+
+        // act
+        var result = await _repository.GetIndicatorDataAsync(500, [], [], ["age"]);
+
+        // assert
+        result.ShouldNotBeEmpty();
+        result.Count().ShouldBe(1);
+        result.ShouldBeEquivalentTo(new List<HealthMeasureModel>()
+        {
+            personsHealthMeasure
```
gareth-allan-bjss, line 342: I think this test should be set up so that the `personsHealthMeasure` has no value for the age dimension and the `maleHealthMeasure` has no value for the sex dimension and we expect both to be returned. That way it will validate (as the previous test did for the sex dimension filtering) that both values with and without age dimensions are returned, but values with a sex dimension are not.

gareth-allan-bjss (commented):
Looks good! Just a few comments, largely around the tests.

api/DHSC.FingertipsNext.Modules.HealthData.UnitTests/Services/IndicatorServiceTests.cs
gareth-allan-bjss: Actually, discussed offline with @andyevans-bjss, and we're looking to remove the limit of 10 for the other fields, so we can ignore that here.

api/DHSC.FingertipsNext.Modules.HealthData/Controllers/V1/IndicatorsController.cs
gareth-allan-bjss: Ah, because you can't do that apparently.

MonkeyDate (approved):

api/DHSC.FingertipsNext.Modules.HealthData.UnitTests/Services/IndicatorServiceTests.cs
adawari-thomas-bjss: touching on this, there are only a limited number of values that can be passed into `inequalities` due to the query param being an enum

MonkeyDate: @adawari-thomas-bjss  It's not just the front end that **could** call the api, it could be called directly with any data being part of the request.

PR #169, opened 2025-02-19
e2e test added

MonkeyDate:
# Description

Jira ticket: [DHSCFT-236](https://bjss-enterprise.atlassian.net/browse/DHSCFT-236)

End-2-end for landing on the search results page with selected ares in the page url. Additionally adding `data-testid` to some elements in the results page to enable more robust testing.

## Changes

- Add `data-testid` attribute to 3 elements within the area filter organism
- Added new e2e test

## Validation

Manual execution of tests for scenarios that were stepped through in the browser.


frontend/fingertips-frontend/playwright/page-objects/pages/resultsPage.ts
```diff
@@ -13,13 +13,66 @@ export default class ResultsPage extends BasePage {
...
+  }
+
+  areaFilterCombobox() {
+    return this.page
+      .getByTestId(this.areaFilterContainer)
+      .getByTestId(this.areaTypeSelector)
+      .locator('select');
+  }
+
+  areaFilterOptions() {
```
alibombaye, line 67: Do these need to be async? the other functions seem to be async

MonkeyDate: No. They return a `Locator` rather than data from within the page.

alibombaye (approved):

PR #167, opened 2025-02-18
Feature/dhscft 57 Improve data import performance and add extra data

andyevans-bjss:
# Description

Jira ticket: dhscft 57

Improve import performance, now runs in about 10 seconds

## Changes

- Add extra area data
- Resolve issue with category data
- Add CI 99.8%
- Remove 'CIA Test' indicators
- Add flag for HasInequalities flag to each indicator
- Add EarliestYear property to each indicator

PR #174, opened 2025-02-20
fix issues with missing data json files

andyevans-bjss:
# Description

Jira ticket:DHSCFT-5

Fix issue with missing indicators.json file


gareth-allan-bjss (approved):
Not easy to review the content of the files, but I'm happy that the ones that were cleared in #167 have been reinstated here, and the search-setup areas and indicator files match the ones in `DataCreator/DataCreator/data/out/`.

PR #176, opened 2025-02-20
recent result tag

Joe-Hendley:
## Description

add a tag if a result is less than a month old

https://bjss-enterprise.atlassian.net/browse/DHSCFT-171

we use the date from the nextjs server to guard against any issues caused by improperly set dates/times on user devices

## Validation

Validated through unit testing on the result page. Unsure if we want to validate we actually use the server time (somehow).

### DIFF
--- a/frontend/fingertips-frontend/app/results/page.tsx
+++ b/frontend/fingertips-frontend/app/results/page.tsx
@@ -14,6 +14,7 @@ import { determineApplicableGroupTypes } from '@/lib/areaFilterHelpers/determine
 import { determineSelectedGroupType } from '@/lib/areaFilterHelpers/determineSelectedGroupType';
 import { AreaTypeKeys } from '@/lib/areaFilterHelpers/areaType';
 import { IndicatorSelectionState } from '@/components/forms/IndicatorSelectionForm/indicatorSelectionActions';
+import { currentServerDate } from '@/lib/dateHelpers/serverDate';
 
 export default async function Page(
   props: Readonly<{
@@ -90,6 +91,8 @@ export default async function Page(
       errors: {},
     };
 
+    const serverDate: Date = await currentServerDate();
+
     return (
       <SearchResults
         initialIndicatorSelectionState={initialState}
@@ -99,6 +102,7 @@ export default async function Page(
         availableGroupTypes={availableGroupTypes}
         selectedAreasData={selectedAreasData}
         searchState={stateManager.getSearchState()}
+        serverDate={serverDate}
       />
     );
   } catch (error) {
--- a/frontend/fingertips-frontend/components/molecules/result/searchResult.test.tsx
+++ b/frontend/fingertips-frontend/components/molecules/result/searchResult.test.tsx
@@ -84,6 +84,35 @@ it('should contain expected text', () => {
   );
 });
 
+it('should display tag if indicator date is within one month of server date', () => {
+  const currentDate = new Date('December 7, 2024');
+  render(
+    <SearchResult
+      result={MOCK_DATA[0]}
+      searchState={initialSearchState}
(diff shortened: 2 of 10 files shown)


frontend/fingertips-frontend/components/pages/results/index.tsx
```diff
@@ -24,6 +24,7 @@ import {
   searchIndicator,
 } from '@/components/forms/IndicatorSearchForm/indicatorSearchActions';
 import { IndicatorSelectionForm } from '@/components/forms/IndicatorSelectionForm';
+// import { currentServerDate } from '@/lib/timeHelpers/timeHelpers';
```
alibombaye, line 27: Needs to be removed?

frontend/fingertips-frontend/app/results/page.tsx
```diff
@@ -99,6 +102,7 @@ export default async function Page(
         availableGroupTypes={availableGroupTypes}
         selectedAreasData={selectedAreasData}
         searchState={stateManager.getSearchState()}
+        serverDate={serverDate}
```
alibombaye, line 105: Can you add a unit test to check this prop is passed as expected. The unit test for the page are already broken down by prop level. So should be a case of adding a new one rather than updating any existing one.

frontend/fingertips-frontend/app/results/page.tsx
```diff
@@ -90,6 +91,8 @@ export default async function Page(
       errors: {},
     };
 
+    const serverDate: Date = await getServerDate();
```
alibombaye, line 94: Not sure I understand the value of abstracting this out to its own function here and not just calling `new Date()` straight in here.

frontend/fingertips-frontend/lib/dateHelpers/dateComparison.test.ts
```diff
@@ -0,0 +1,63 @@
...
+    },
+    {
+      title: 'year boundary',
+      indicatorDate: new Date(2024, 11, 31),
+      serverDate: new Date(2025, 0, 1),
+      expectedResult: true,
+    },
+  ])('$title', ({ serverDate, indicatorDate, expectedResult }) => {
+    expect(isWithinOneMonth(serverDate, indicatorDate)).toBe(expectedResult);
+  });
```
alibombaye, line 62: Nice 👍

frontend/fingertips-frontend/app/results/page.tsx
alibombaye: I'm not sure how I feel about the name `serverDate` what is wrong with maybe just `currentDate`? we just happen to call it on the server.

frontend/fingertips-frontend/components/molecules/result/searchResult.test.tsx
```diff
@@ -84,6 +84,35 @@ it('should contain expected text', () => {
...
+    <SearchResult
+      result={MOCK_DATA[0]}
+      searchState={initialSearchState}
+      handleClick={mockHandleClick}
+      serverDate={currentDate}
+    />
+  );
+  expect(screen.getAllByRole('strong').at(0)?.textContent).toContain(
+    'Updated in last month'
+  );
```
alibombaye, line 99: I think making this assertion similar to how you have done the one below might be better

```
expect(screen.getByText('Updated in last month')).toBeInTheDocument();
```

frontend/fingertips-frontend/app/results/page.tsx
Joe-Hendley: Initially I was trying to slot it in on a higher level than the page, which required a helper function, and then I didn't clock that the app page render happened on the server rather  than locally.

alibombaye (approved):

PR #177, opened 2025-02-21
add empty dependency array to useeffect

Joe-Hendley:
No ticket attached. Solves performance issues & timeout with the highcharts map rendering.

Previously, the useEffect that was used to load the map didn't have a dependency array, causing it to reload the map in an infinite loop.

Validated locally. 

There are still some performance issues when mousing over regions with complex borders, but this should resolve the immediate issues.

### DIFF
--- a/frontend/fingertips-frontend/components/organisms/ThematicMap/index.tsx
+++ b/frontend/fingertips-frontend/components/organisms/ThematicMap/index.tsx
@@ -95,7 +95,7 @@ export function ThematicMap({
     loadHighchartsModules(async () => {
       setOptions(mapOptions);
     });
-  });
+  }, []);
 
   if (!options) {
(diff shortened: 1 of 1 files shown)


gareth-allan-bjss:
Some React documentation that makes it clear on the difference between passing an empty array of dependencies and passing no dependencies: https://react.dev/reference/react/useEffect#examples-dependencies

gareth-allan-bjss (approved):
Nice fix

Joe-Hendley:
The linter suggestion to use `mapOptions` was iffy and also caused looping, but if we use the prop that drives mapOptions it keeps the linter happy *and* doesn't cause endless renders

gareth-allan-bjss (approved):

PR #180, opened 2025-02-21
Import data into DB from CSV files

epercy-bjss:
# Description

Jira ticket: [DHSCFT-354](https://bjss-enterprise.atlassian.net/browse/DHSCFT-354)

CSV files have been generated for data import into the Fingertips database. Post Deployment script has now been updated to handle this import process.

## Changes

- CSV files generated from DataCreator project
- Post deployment script updated to BULK IMPORT csv files into temporary tables
- CSV file location is passed into the post deployment script dynamically
- Data then transformed into the main schema before the temporary table is dropped
- All tables are dropped at the beginning of the process to have a clean state evey time
- Workloaw does not need to change, as the Post Deployment is handled by the dacpac deploy job

## Validation

Tested locally, but there are dependencies on two other tickets before tests will pass:

https://bjss-enterprise.atlassian.net/browse/DHSCFT-51
https://bjss-enterprise.atlassian.net/browse/DHSCFT-291

Deployed to a separate database and tested manually through the branch. E2E test failing until prerequiste tickets are implemented.


.github/workflows/fingertips-workflow.yml
```diff
@@ -284,7 +284,7 @@ jobs:
           connection-string: ${{ secrets.AZURE_SQL_CONNECTION_STRING }}
           path: 'fingertips-db.dacpac'
           action: 'publish'
-          arguments: '/p:DropObjectsNotInSource=true /p:BlockOnPossibleDataLoss=false'
+          arguments: '/p:DropObjectsNotInSource=true /p:BlockOnPossibleDataLoss=false /v:UseAzureBlob=1 /v:BlobStorageLocation="https://dlsfingertipsuks001.blob.core.windows.net/sql-data" /v:MasterKeyPassword=${{ secrets.SQL_MASTER_KEY_PASSWORD }}'
```
HowardMayBjss, line 287: Is there a better way of handling the blobStorageLocation rather than embedding it in the workflow file? Also this is not environment specific so any changes in dev will impact prod - or is there an env specific file name used?

DataCreator/DataCreator/DataManager.cs
```diff
@@ -200,27 +199,12 @@ private static List<AgeEntity> AddAgeIds(List<HealthMeasureEntity> healthMeasure
             return allAges.Where(age=>usedAgeIds.Contains(age.AgeID)).ToList();
         }
 
-        private static void AddSexIds(List<HealthMeasureEntity> healthMeasures)
```
HowardMayBjss, line 203: Why is this PR removing this code? Or does it just need a merge from main?

.github/workflows/fingertips-workflow.yml
epercy-bjss: Change the `blobStorageLocation` parmaeter to use GitHub environment-specific variable

DataCreator/DataCreator/DataManager.cs
epercy-bjss: I believe SexID is being removed from the schema

HowardMayBjss (approved):

database/fingertips-db/fingertips-db.sqlproj
```diff
@@ -26,6 +25,30 @@
     <Build Include="Tables\SexDimension.sql" />
     <Build Include="Security\Areas.sql" />
     <Build Include="Tables\Areas.sql" />
+    <Build Include="Tables\AreaTypes.sql" />
```
dlzhry2, line 28: Does the change mean we can remove `<PostDeploy Include="PostDeployment\PopulateFingertips.sql" />` from line 18 in this file?

dlzhry2: Oh my bad, I had wrongly assumed you had deleted it and it was now a different process. It is still there and will be used! Marking my comment as resolved/invalid.

PR #186, opened 2025-02-25
DHSCFT-285: show groups and select a group.

alibombaye:
# Description

Jira ticket: [DHSCFT-285](https://example.com)

Shows the group drop down based upon the group type
Default select a group
Handle selecting a group
Filter available areas based upon the group and areaType selected.

### DIFF
--- a/frontend/fingertips-frontend/components/organisms/AreaFilter/AreaFilter.test.tsx
+++ b/frontend/fingertips-frontend/components/organisms/AreaFilter/AreaFilter.test.tsx
@@ -12,8 +12,9 @@ import {
   nhsIntegratedCareBoardsAreaType,
   nhsRegionsAreaType,
 } from '@/lib/areaFilterHelpers/areaType';
-import { AreaType } from '@/generated-sources/ft-api-client';
+import { Area, AreaType } from '@/generated-sources/ft-api-client';
 import {
+  allNhsRegions,
   eastEnglandNHSRegion,
   northEastAndYorkshireNHSRegion,
 } from '@/mock/data/areas/nhsRegionsAreas';
@@ -92,11 +93,13 @@ describe('Area Filter', () => {
   });
 
   describe('Area type', () => {
+    const areaTypeDropDownLabel = 'Select an area type';
+
     it('should disable the select area type drop down when there are areas selected', () => {
       render(<AreaFilter selectedAreasData={mockSelectedAreasData} />);
 
       expect(
-        screen.getByRole('combobox', { name: /Select an area type/i })
+        screen.getByRole('combobox', { name: areaTypeDropDownLabel })
       ).toBeDisabled();
     });
 
@@ -104,7 +107,7 @@ describe('Area Filter', () => {
       render(<AreaFilter selectedAreasData={[]} />);
 
       expect(
-        screen.getByRole('combobox', { name: /Select an area type/i })
+        screen.getByRole('combobox', { name: areaTypeDropDownLabel })
       ).not.toBeDisabled();
     });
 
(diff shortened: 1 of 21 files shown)


frontend/fingertips-frontend/components/organisms/AreaFilter/AreaFilter.test.tsx
```diff
@@ -282,6 +348,109 @@ describe('Area Filter', () => {
...
+
+      expect(
+        screen.getByRole('combobox', { name: groupDropDownLabel })
+      ).toBeInTheDocument();
+      expect(
+        screen.getByRole('combobox', { name: groupDropDownLabel })
+      ).not.toBeDisabled();
+    });
+
+    it('should render all applicable group provided', () => {
```
williamwillis11, line 378: groups?

frontend/fingertips-frontend/lib/areaFilterHelpers/determineSelectedGroup.ts
```diff
@@ -0,0 +1,15 @@
+import { Area } from '@/generated-sources/ft-api-client';
+import { areaCodeForEngland } from '../chartHelpers/constants';
+
+export const determineSelectedGroup = (
+  selectedGroup?: string,
+  availableGroups?: Area[]
+): string => {
+  if (selectedGroup) return selectedGroup;
+
+  if (availableGroups && availableGroups.length > 0) {
```
williamwillis11, line 10: would

(availableGroups?.length)

be equivalent?

alibombaye: No doesn't like that. Says it can be possibly undefined even with using the `?`. Will leave this as is.

frontend/fingertips-frontend/components/forms/IndicatorSelectionForm/indicatorSelectionActions.test.ts
```diff
@@ -20,6 +21,10 @@ const initialState: IndicatorSelectionState = {
 let formData: FormData;
 
 describe('Indicator Selection Actions', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
```
williamwillis11, line 25: just wondering out loud if we should add:
https://jestjs.io/docs/cli#--clearmocks

to the jest npm script? that way it will do this by default for all tests, which we want right? and we dont have to have this code in each file

alibombaye: Interesting. I've never used that before. I think it might be better to be explicit within the tests that this is going on. 

If we did want to do this I think we should try that as a seperate tech task and see how that turns out. Rather than as part of this PR at the moment.

frontend/fingertips-frontend/lib/areaFilterHelpers/determineSelectedGroup.ts
MonkeyDate: return availableGroups?.length ? availableGroups[0].code : areaCodeForEngland;

frontend/fingertips-frontend/mock/server/handlers.ts
```diff
@@ -114,8 +124,25 @@ export function getGetAreaTypeMembers200Response(areaTypeKey: AreaTypeKeys) {
   return mockAvailableAreas[areaTypeKey];
 }
 
-export function getGetArea200Response(areaCode: string) {
-  return mockAreaData[areaCode];
+export function getGetArea200Response(
+  areaCode: string,
+  includeChildren: string,
```
williamwillis11, line 129: could this be a boolean?

then we can just do 
 if (includeChildren) instead of  if (includeChildren === 'true')

alibombaye: Since this comes from the query param as a string. We would still need to do some kind of conversion to a boolean type.

williamwillis11 (approved):

PR #190, opened 2025-02-26
Feature/dhscft 400 add ai search single document lookup

Joe-Hendley:
https://bjss-enterprise.atlassian.net/browse/DHSCFT-400

Adds a single document retrieval service to the existing AI Search integration

adds the unitValue index (already exists in mock & real data)

It follows the same implementation & test patterns as the working searches, but as of writing needs manual validation against a real AI Search instance


frontend/fingertips-frontend/lib/search/searchTypes.ts
```diff
@@ -34,3 +35,7 @@ export interface IIndicatorSearchService {
 export interface IAreaSearchService {
   getAreaSuggestions(partialAreaName: string): Promise<AreaDocument[]>;
 }
+
+export interface IIndicatorDocumentService {
+  getIndicator(indicatorId: string): Promise<IndicatorDocument>;
```
HowardMayBjss, line 40: We could model this as a new function on the existing IndicatorSearchService which would significantly reduce the amount of new code. What are the benefits of creating a separate service?

Joe-Hendley: Good shout. I've condensed the functionality & tests into the existing service.

frontend/fingertips-frontend/lib/search/indicatorSearchServiceMock.ts
```diff
@@ -27,4 +27,10 @@ export class IndicatorSearchServiceMock implements IIndicatorSearchService {
       })
       .slice(0, 20);
   }
+
+  public async getIndicator(indicatorId: string): Promise<IndicatorDocument> {
+    return this.mockIndicatorData.filter((indicator) => {
```
dlzhry2, line 32: Could potentially use: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

Will return earlier and return undefined if not found. Feel free to dismiss if not - it's just for the mock data ultimately.

HowardMayBjss (approved):

frontend/fingertips-frontend/lib/search/indicatorSearchService.ts
```diff
@@ -57,4 +57,8 @@ export class IndicatorSearchService implements IIndicatorSearchService {
 
     return results.slice(0, 20);
   }
+
+  async getIndicator(indicatorId: string): Promise<IndicatorDocument> {
+    return this.searchClient.getDocument(indicatorId);
```
gareth-allan-bjss, line 62: What happens if a document with the ID specified isn't found?

frontend/fingertips-frontend/lib/search/indicatorSearchService.test.ts
```diff
@@ -177,6 +180,44 @@ describe('IndicatorSearchService', () => {
...
+    expect(SearchClient).toHaveBeenCalledWith(
+      'test-url',
+      INDICATOR_SEARCH_INDEX_NAME,
+      expect.any(Object)
+    );
+
+    expect(mockGetDocument).toHaveBeenCalledWith(indicatorId);
+  });
+
+  it('should get a single indicator a document', async () => {
```
gareth-allan-bjss, line 199: Typo?
```suggestion
  it('should get a single indicator document', async () => {
```

search-setup/src/indexOperations.ts
```diff
@@ -87,6 +87,13 @@ export function buildIndicatorSearchIndex(name: string): SearchIndex {
         sortable: false,
         filterable: true,
       },
+      {
+        name: IndicatorSearchIndexColumnNames.INDICATOR_UNIT_LABEL,
+        type: 'Edm.String',
+        searchable: false,
+        sortable: false,
+        filterable: false,
+      },
```
gareth-allan-bjss, line 96: There are integration tests for this application (in `search-setup/integration_tests`) that should be updated to validate this change, I think.

frontend/fingertips-frontend/lib/search/indicatorSearchServiceMock.ts
Joe-Hendley: added complexity in implementing due to async/typescript things

dlzhry2: I think you can type hint `Promise<IndicatorDocument | undefined>` and update the signature to match in the interface/base SearchService class, since undefined is a legitimate result for this type of search.

I'll mark resolved though and happy for it without since it is for the mock and current code will work + tests pass.

Joe-Hendley: Yeah, that's what I've ended up on - yesterday I was conflating the promise not resolving at all with being unable to return an element. Fresh day, fresh eyes.

frontend/fingertips-frontend/lib/search/indicatorSearchService.ts
Joe-Hendley: As discussed in Dan's comment, brain failed to find. Updated return type.

gareth-allan-bjss: I don't think `getDocument` is defined as potentially returning undefined. Have you been able to validate that this is what actually happens when a document doesn't exist? My concern is that it'll raise an error (because the API returns a 404) instead. I think we should understand what will happen in this scenario.

Joe-Hendley: Correct, corrected.

frontend/fingertips-frontend/lib/search/indicatorSearchService.ts
```diff
@@ -57,4 +57,16 @@ export class IndicatorSearchService implements IIndicatorSearchService {
...
   }
+
+  async getIndicator(
+    indicatorId: string
+  ): Promise<IndicatorDocument | undefined> {
+    const indicatorDocument = await this.searchClient
+      .getDocument(indicatorId)
+      .catch(() => {
+        return undefined;
+      });
```
gareth-allan-bjss, line 68: Can you add a test that verifies `getIndicator` returns undefined if `getDocument` throws an error please?

search-setup/integration_tests/search.test.ts
```diff
@@ -97,6 +97,21 @@ describe('AI search index creation and data loading', () => {
...
+      expect (result.value[0].indicatorName).toBeDefined();
+      expect (result.value[0].indicatorDefinition).toBeDefined();
+      expect (result.value[0].dataSource).toBeDefined();
+      expect (result.value[0].earliestDataPeriod).toBeDefined();
+      expect (result.value[0].latestDataPeriod).toBeDefined();
+      expect (result.value[0].lastUpdatedDate).toBeDefined();
+      expect (result.value[0].associatedAreaCodes).toBeDefined();
+      expect (result.value[0].hasInequalities).toBeDefined();
+      expect (result.value[0].unitLabel).toBeDefined();
+    })
```
gareth-allan-bjss, line 113: I think this is a duplication of the intent of the snapshot test (`should have correct index name and number of fields`), so I'm not really sure we need to add it.

search-setup/integration_tests/search.test.ts
```diff
@@ -57,10 +55,8 @@ describe('AI search index creation and data loading', () => {
         body: JSON.stringify({
           search: searchTerm,
           searchFields: 'indicatorID,indicatorDefinition,indicatorName',
-          select: 'indicatorID,indicatorDefinition,indicatorName',
```
gareth-allan-bjss, line 60: Not a major issue, but I don't think you need to remove this `select` property any more.

gareth-allan-bjss (approved):

search-setup/integration_tests/search.test.ts
Joe-Hendley: I've put it back.

gareth-allan-bjss (approved):

PR #192, opened 2025-02-26
DHSCFT-341: Proposed RAG benchmarking calcs

stuc-bjss:
# Description

Jira ticket: [DHSCFT-341](https://bjss-enterprise.atlassian.net/browse/DHSCFT-341)

Proposed basic implementation of the RAG benchmarking calculation

## Changes

- Added comparison_method to query string of HealthData Indicator controller
- Loaded data as requested but also for England (hard coded)
- Separate out the data for England and use it to benchmark the other areas
- Polarity currently fixed - but changeable for unit tests.
- Add benchmark property to returned datapoints

image

## Validation

- Unit tests covering, no rag, rag with lowerIsGood, rag with higherIsGood and noJudgement
- Doesn't apply benchmarking values to itself - eg England v England
- Can request the benchmark area in the results
- Return null benchmark data not found
- Only make one DB query


api/DHSC.FingertipsNext.Modules.HealthData.Schemas/HealthDataPointBenchmark.cs
```diff
@@ -0,0 +1,27 @@
...
+/// <summary>
+/// Benchmark type and value
+/// </summary>
+public class HealthDataPointBenchmark
+{
+    /// <summary>
+    /// The value
+    /// </summary>
+    [JsonPropertyName("value")]
+    public string? Value { get; init; } = String.Empty;
```
HowardMayBjss, line 14: I think this is a mandatory field, and I think '?' makes it optional so isn't needed.

HowardMayBjss, line 14: I think we as a team need to work on these names. In my mind 'Value' is quite generic and doesn't communicate the intent of the parameter very well. I think making this an enumerated type would help.

api/DHSC.FingertipsNext.Modules.HealthData.Schemas/HealthDataPointBenchmark.cs
```diff
@@ -0,0 +1,27 @@
...
+    
+    /// <summary>
+    /// Type
+    /// </summary>
+    [JsonPropertyName("type")]
```
HowardMayBjss, line 19: Again I think we need to work on the names

api/DHSC.FingertipsNext.Modules.HealthData.Schemas/HealthDataPointBenchmark.cs
```diff
@@ -0,0 +1,27 @@
...
+    public string? Type { get; init; } = String.Empty;
+    
+    /// <summary>
+    /// The compared value
+    /// </summary>
+    [JsonPropertyName("comparedTo")]
```
HowardMayBjss, line 25: I'm not sure about putting the compared area value here, but given we expect to be comparing to areas other than England, then I think there is merit in putting the compared area areaCode or areaName.

api/DHSC.FingertipsNext.Modules.HealthData.Schemas/HealthDataPointBenchmark.cs
stuc-bjss: Fixed

stuc-bjss: Determination? Just let me know what the property should be called.
I looked at enums in .NET and it appeared that they always had to be integer values - so I may have missed something about how they work.

api/DHSC.FingertipsNext.Modules.HealthData.Schemas/HealthDataPointBenchmark.cs
stuc-bjss: Just let me know what the property should be called, Gareth's suggestion was `type`.

api/DHSC.FingertipsNext.Modules.HealthData.Schemas/HealthDataPointBenchmark.cs
stuc-bjss: I've removed it. Not sure that adding `areaCode` or `areaName` here is right, same for `type` as they would have the same values in every data point in the result, but would make sense higher up where it would only need to be said once?

api/DHSC.FingertipsNext.Modules.HealthData.Service/IndicatorService.cs
```diff
@@ -6,43 +6,115 @@
...
+        var (healthDataForAreas, benchmarkHealthData) =
+            SeparateHealthAndBenchMarkData(allHealthDataForAreas, comparisonMethod, wasBenchMarkAreaAdded);
+
+        return ApplyBenchMarkedAreaData(healthDataForAreas, benchmarkHealthData, comparisonMethod, Polarity);
+    }
+
+    private static (string[] areaCodesForSearch, bool wasBenchMarkAreaAdded) GetAreaCodesForSearch(string[] areaCodes,
+        string comparisonMethod)
+    {
+        var areaCodesForSearch = new List<string>(areaCodes.Distinct().Take(10).ToArray());
```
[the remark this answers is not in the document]
stuc-bjss: never done .NET before and the List vs Array thing is confusing (why aren't they the same thing) I shall try to find another way to do it.

api/definition/swagger.yaml
```diff
@@ -397,6 +398,32 @@ components:
             least 5 data points to use.
           enum: [Increasing, Decreasing, NoChange, CannotBeCalculated,
             NotYetCalculated]
+        benchmark:
```
HowardMayBjss, line 401: can we rename this benchmarkComparison

api/DHSC.FingertipsNext.Modules.HealthData.Service/IndicatorService.cs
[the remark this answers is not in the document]
steves-bjss: Arrays have a static size, so if you want to grow them you have to create a new array.  Lists are dynamically sized and automatically leave headroom to add values.  Wherever possible I'd suggest we don't use arrays - stick to using interfaces, such as IEnumerable, ICollection, IList for passing parameters around.  It allows more flexibility.  Happy to talk it through.

api/DHSC.FingertipsNext.Modules.HealthData.Service/IndicatorService.cs
```diff
@@ -6,43 +6,115 @@
...
-/// The business logic for accessing indicator data.
+///     The business logic for accessing indicator data.
 /// </summary>
 /// <remarks>
-/// Does not include anything specific to the hosting technology being used.
+///     Does not include anything specific to the hosting technology being used.
 /// </remarks>
 public class IndicatorService(IRepository _repository, IMapper _mapper) : IIndicatorsService
 {
+    public const string AreaCodeEngland = "E92000001";
```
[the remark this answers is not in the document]
stuc-bjss: Discussed with Steve, needs a whole new project in .NET to do this, probably not worth it at the moment.

HowardMayBjss (approved):

PR #196, opened 2025-02-27
Feature/DHSCFT-380_skeleton_chart_views

WolfieKnee:
# Description

Jira ticket: [DHSCFT-380](https://bjss-enterprise.atlassian.net/browse/DHSCFT-380)

Chart Views - Create skeleton components

## Changes
- Add ViewsContext to end of Charts page
- Add skeleton of views/dashboards.
- Update existing tests to support interim conditions
- Unit tests added that ViewsContext returns the correct View

## Validation
- functional testing against local api
- typecheck, lint, prettier, unit tests and e2e tests passing


frontend/fingertips-frontend/components/views/ViewsContext.tsx
```diff
@@ -0,0 +1,46 @@
...
+      return <OneIndicatorOneAreaView searchState={searchState} />;
+    } else if (areaCodes.length >= 2) {
+      return <OneIndicatorTwoOrMoreAreasView searchState={searchState} />;
+    }
+  } else if (indicators.length >= 2) {
+    if (areaCodes.length === 1 && areaCodes[0] === areaCodeForEngland) {
+      return <TwoOrMoreIndicatorsEnglandView searchState={searchState} />;
+    }
+    return <TwoOrMoreIndicatorsAreasView searchState={searchState} />;
+  }
```
gareth-allan-bjss, line 33: Just a suggestion, and I'm happy for you to decide whether you want to implement it or not, but I think this might be a bit more readable if you avoid nesting the if statements. Something like:
```suggestion
  if (indicators.length === 1 && areaCodes.length === 1) {
    return <OneIndicatorOneAreaView searchState={searchState} />;
  }

  if (indicators.length === 1 && areaCodes.length >= 2) {
    return <OneIndicatorTwoOrMoreAreasView searchState={searchState} />;
  }

  if (
    indicators.length >= 2 &&
    areaCodes.length === 1 &&
    areaCodes[0] === areaCodeForEngland
  ) {
    return <TwoOrMoreIndicatorsEnglandView searchState={searchState} />;
  }

  return <TwoOrMoreIndicatorsAreasView searchState={searchState} />;
```

I think it would lend itself better to the addition of more views in future as well.

frontend/fingertips-frontend/components/views/ViewsContext.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+  () =>
+    function fn() {
+      mockTwoOrMoreIndicatorsEnglandView();
+      return <div />;
+    }
+);
+
+async function generateSearchParams(value: SearchStateParams) {
+  return value;
+}
```
gareth-allan-bjss, line 48: This function is redundant.

frontend/fingertips-frontend/components/views/ViewsContext.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+describe('ViewsContext', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+  it.each([
+    [['foo'], ['zam'], mockOneIndicatorOneAreaView],
+    [['foo', 'bar'], ['zam'], mockOneIndicatorTwoOrMoreAreasView],
```
gareth-allan-bjss, line 56: We should also validate that we get the one indicator, two or more areas view for more than two areas:
```suggestion
    [['foo', 'bar'], ['zam'], mockOneIndicatorTwoOrMoreAreasView],
    [['foo', 'bar', 'baz'], ['zam'], mockOneIndicatorTwoOrMoreAreasView],
```

frontend/fingertips-frontend/components/views/TwoOrMoreIndicatorsAreasView/index.tsx
gareth-allan-bjss, line 1: Should this just be called `TwoOrMoreIndicatorsAreasView`? It doesn't actually matter how many areas are selected, does it? Just that it's not England.

frontend/fingertips-frontend/components/views/ViewsContext.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+    [['foo'], ['foo', 'bar', 'baz'], mockTwoOrMoreIndicatorsAreasView],
```
gareth-allan-bjss, line 57: Conversely I think we should have a test to validate this view is shown when there are just two indicators selected:
```suggestion
    [['foo'], ['foo', 'bar'], mockTwoOrMoreIndicatorsAreasView],
    [['foo'], ['foo', 'bar', 'baz'], mockTwoOrMoreIndicatorsAreasView],
```

frontend/fingertips-frontend/components/views/ViewsContext.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+    [[areaCodeForEngland], ['foo', 'bar'], mockTwoOrMoreIndicatorsEnglandView],
```
gareth-allan-bjss, line 58: Again, given you've got a nice parameterised test we may as well add a case to validate this works for more than 2 indicators:
```suggestion
    [[areaCodeForEngland], ['foo', 'bar'], mockTwoOrMoreIndicatorsEnglandView],
    [[areaCodeForEngland], ['foo', 'bar', 'baz'], mockTwoOrMoreIndicatorsEnglandView],
```

frontend/fingertips-frontend/components/views/ViewsContext.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+  ])(
+    'should return the expected view',
+    async (areaCodes, indicators, correctView) => {
```
gareth-allan-bjss, line 61: Given your views are named with indicators first and then areas, I think this test would be easier to follow if you swapped the order of the parameters here:
```suggestion
    async (indicators, areaCodes, correctView) => {
```

frontend/fingertips-frontend/components/views/OneIndicatorOneAreaView/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { areaCodeForEngland } from '@/lib/chartHelpers/constants';
+import {
+  SearchParams,
+  SearchStateManager,
+  SearchStateParams,
+} from '@/lib/searchStateManager';
+
+type OneIndicatorOneAreaViewProps = {
+  searchState: SearchStateParams;
+};
```
gareth-allan-bjss, line 11: Could we just have a single type for all view's props? Because of the way we want this to work, they'll only really ever need the searchState, right?

frontend/fingertips-frontend/components/views/OneIndicatorOneAreaView/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+    [SearchParams.AreasSelected]: areaCodes,
+    [SearchParams.GroupSelected]: selectedGroupCode,
+  } = stateManager.getSearchState();
+
+  const areasSelected = areaCodes ?? [];
+
+  const areaCodesToRequest =
+    selectedGroupCode && selectedGroupCode != areaCodeForEngland
+      ? [...areasSelected, areaCodeForEngland, selectedGroupCode]
+      : [...areasSelected, areaCodeForEngland];
```
gareth-allan-bjss, line 27: We know this should only be a single value, so we could do something like:
```suggestion
  if (areaCodes?.length !== 1) {
    throw new Error('Invalid parameters provided to view');
  }

  const areaSelected = areaCodes[0];

  const areaCodesToRequest = [areaSelected, areaCodeForEngland];
  if (selectedGroupCode && selectedGroupCode != areaCodeForEngland) {
    areaCodesToRequest.push(selectedGroupCode);
  }
```

Given it's fundamentally incorrect to create this component with the wrong number of area codes we should verify that up front and throw an error if we've not got data we should have. Then (because typescript is really nice about narrowing types) we can continue from that point safely knowing that the parameters are as we expect. We should also extend this to throwing an error when anything other than a single indicator is passed, when we need to read those.

frontend/fingertips-frontend/components/dashboards/OneIndicatorOneAreaDashboard/index.tsx
gareth-allan-bjss, line 1: I'm not sure I love the name "dashboard" for these, mainly because it's not clear to me that a View is a parent of a dashboard. I also can't think of anything better...

frontend/fingertips-frontend/components/dashboards/TwoOrMoreIndicatorsAreasDashboard/index.tsx
```diff
@@ -0,0 +1,14 @@
+'use client';
+
+import { H2, H3 } from 'govuk-react';
+
+export function TwoOrMoreIndicatorsAreasDashboard() {
+  return (
+    <>
+      <H2>View indicators data for the selected area(s)</H2>
+      <H3>Spine Chart</H3>
+      <H3>Heatmap</H3>
```
gareth-allan-bjss, line 10: For the record, the spine chart is only shown if <= 2 areas are selected, and the heat map is only shown when 2+ areas are selected (including all areas in a group).

frontend/fingertips-frontend/components/views/TwoOrMoreIndicatorsAreasView/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+  const areasSelected = areaCodes ?? [];
+
+  const areaCodesToRequest =
+    selectedGroupCode && selectedGroupCode != areaCodeForEngland
+      ? [...areasSelected, areaCodeForEngland, selectedGroupCode]
+      : [...areasSelected, areaCodeForEngland];
+
+  await connection();
+
+  console.log('TODO: fetch health data with inequalites');
```
gareth-allan-bjss, line 32: I'm sure this is just a copy and paste, but we won't want inequalities data in this scenario.

frontend/fingertips-frontend/components/views/TwoOrMoreIndicatorsEnglandView/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+    [SearchParams.AreasSelected]: areaCodes,
+    [SearchParams.GroupSelected]: selectedGroupCode,
+  } = stateManager.getSearchState();
+
+  const areasSelected = areaCodes ?? [];
+
+  const areaCodesToRequest =
+    selectedGroupCode && selectedGroupCode != areaCodeForEngland
+      ? [...areasSelected, areaCodeForEngland, selectedGroupCode]
+      : [...areasSelected, areaCodeForEngland];
```
gareth-allan-bjss, line 28: Again, we know there will only ever be one area code selected here, so we can simplify this logic as with the one indicator, one area view.

frontend/fingertips-frontend/components/views/ViewsContext.tsx
WolfieKnee: I've tried both versions, so happy to go with the more maintainable option.

frontend/fingertips-frontend/components/views/OneIndicatorOneAreaView/index.tsx
WolfieKnee: Agreed. Would the definition go in common types file?

frontend/fingertips-frontend/components/views/OneIndicatorOneAreaView/index.tsx
WolfieKnee: Thanks for the suggestion, that will pattern across several views.

frontend/fingertips-frontend/components/dashboards/TwoOrMoreIndicatorsAreasDashboard/index.tsx
gareth-allan-bjss: Just wanted to mention it as you've called out optional charts in some views and not others. Didn't want any confusion spreading!

frontend/fingertips-frontend/components/views/OneIndicatorOneAreaView/index.tsx
gareth-allan-bjss: I'm not sure we have a common types file, do we?

Perhaps we just rename `ViewsContextProps` in `frontend/fingertips-frontend/components/views/ViewsContext.tsx` to something like `ViewProps` and use it for the context and all its views?

frontend/fingertips-frontend/components/views/ViewsContext.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+  if (
+    indicators.length >= 2 &&
+    areaCodes.length === 1 &&
+    areaCodes[0] === areaCodeForEngland
+  ) {
+    return <TwoOrMoreIndicatorsEnglandView searchState={searchState} />;
+  }
+
+  return <TwoOrMoreIndicatorsAreasView searchState={searchState} />;
```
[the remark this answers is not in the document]
WolfieKnee: updated return type

frontend/fingertips-frontend/components/views/ViewsContext.test.tsx
WolfieKnee: good point

frontend/fingertips-frontend/components/views/ViewsContext.test.tsx
WolfieKnee: also switched the test data to match

frontend/fingertips-frontend/components/views/TwoOrMoreIndicatorsEnglandView/index.tsx
WolfieKnee: This works, but it feels like typescript should do this more nicely...

frontend/fingertips-frontend/components/dashboards/OneIndicatorOneAreaDashboard/index.tsx
WolfieKnee: I changed the naming to be equaly as unpleasent, but show the link to the views.

frontend/fingertips-frontend/components/views/TwoOrMoreIndicatorsAreasView/index.tsx
WolfieKnee: I think we have this now

frontend/fingertips-frontend/components/views/ViewsContext.test.tsx
```diff
@@ -43,27 +43,34 @@ jest.mock(
...
+    [
+      ['foo', 'bar', 'baz'],
+      [areaCodeForEngland],
+      mockTwoOrMoreIndicatorsEnglandView,
+    ],
+    [
+      ['foo', 'bar', 'baz'],
+      [areaCodeForEngland],
+      mockTwoOrMoreIndicatorsEnglandView,
+    ],
```
gareth-allan-bjss, line 64: Aren't these the same test case?

frontend/fingertips-frontend/components/views/ViewsContext.test.tsx
```diff
@@ -43,27 +43,34 @@ jest.mock(
...
+    [['foo', 'bar', 'baz'], ['foo'], mockTwoOrMoreIndicatorsAreasView],
+    [['foo', 'bar'], [areaCodeForEngland], mockTwoOrMoreIndicatorsEnglandView],
```
gareth-allan-bjss, line 66: I think this is a duplicate of the scenario on line 54,

frontend/fingertips-frontend/components/views/ViewsContext.test.tsx
gareth-allan-bjss: I still don't think you have a test case that covers one indicator and more than two areas shows `mockOneIndicatorTwoOrMoreAreasView`.

frontend/fingertips-frontend/components/dashboards/OneIndicatorOneAreaDashboard/index.tsx
gareth-allan-bjss: I think its an improvment, so happy to stick with it.

frontend/fingertips-frontend/components/views/OneIndicatorOneAreaView/index.tsx
```diff
@@ -1,33 +1,28 @@
...
   const {
-    [SearchParams.AreasSelected]: areaCodes,
+    [SearchParams.AreasSelected]: areasSelected,
     [SearchParams.GroupSelected]: selectedGroupCode,
   } = stateManager.getSearchState();
 
-  const areasSelected = areaCodes ?? [];
+  if (areasSelected?.length !== 1) {
+    throw new Error('Invalid parameters provided to view');
+  }
```
gareth-allan-bjss, line 17: I think now we're having the views throw an error when incorrectly configured, we should have a unit test that validates this behaviour. I'm happy if you want to do this as part of adding tests for the views once they're more fleshed out in a subsequent PR.

frontend/fingertips-frontend/components/views/OneIndicatorOneAreaView/index.tsx
```diff
@@ -1,33 +1,28 @@
...
 
-  const areaCodesToRequest =
-    selectedGroupCode && selectedGroupCode != areaCodeForEngland
-      ? [...areasSelected, areaCodeForEngland, selectedGroupCode]
-      : [...areasSelected, areaCodeForEngland];
+  const areaCodesToRequest = [areasSelected, areaCodeForEngland];
```
gareth-allan-bjss, line 19: `areasSelected` is an array here, so I think you may want to do:
```suggestion
  const areaCodesToRequest = [...areasSelected, areaCodeForEngland];
```

frontend/fingertips-frontend/components/views/TwoOrMoreIndicatorsAreasView/index.tsx
```diff
@@ -1,40 +1,35 @@
...
   const {
-    [SearchParams.AreasSelected]: areaCodes,
+    [SearchParams.AreasSelected]: areasSelected,
     [SearchParams.GroupSelected]: selectedGroupCode,
   } = stateManager.getSearchState();
 
-  const areasSelected = areaCodes ?? [];
+  if (!areasSelected || areasSelected.length < 2) {
+    throw new Error('Invalid parameters provided to view');
+  }
```
gareth-allan-bjss, line 18: This is the "2+ indicators, for specified areas (not England)" view from [the page](https://confluence.collab.test-and-trace.nhs.uk/display/FTN/Frontend+Application+-+Displaying+Charts), right? In which case, it needs to support 2+ indicators, 1 area, which this will prevent.

frontend/fingertips-frontend/components/views/TwoOrMoreIndicatorsEnglandView/index.tsx
```diff
@@ -1,31 +1,25 @@
...
-  const {
-    [SearchParams.AreasSelected]: areaCodes,
-    [SearchParams.GroupSelected]: selectedGroupCode,
-  } = stateManager.getSearchState();
+  const { [SearchParams.AreasSelected]: areasSelected } =
+    stateManager.getSearchState();
 
-  const areasSelected = areaCodes ?? [];
+  if (
+    areasSelected! &&
```
gareth-allan-bjss, line 15: I don't think this needs the non-null assertion here:
```suggestion
    areasSelected &&
```

frontend/fingertips-frontend/components/views/OneIndicatorOneAreaView/index.tsx
WolfieKnee: yes - working on this as part of 411 now

frontend/fingertips-frontend/components/views/TwoOrMoreIndicatorsEnglandView/index.tsx
WolfieKnee: I think I do in order to then make the next checks, but optional chainging seems better.

gareth-allan-bjss: Just having the `if(areasSelected ...` should be enough for Typescript to narrow the type of `areasSelected` so that it doesn't include undefined and you can then reference properties of it after that. I tried it quickly on my machine and it seemed happy enough with just removing the `!`. You're right though that optional chaining is neater.

In general, I'm always wary when I see the `!` operator being used. There are obviously some valid use cases, but it's too often used when not needed, or because some defensive coding that should probably be present hasn't been implemented.

PR #191, opened 2025-02-26
DHSCFT-332 Add trend dimension to API and DB

dlzhry2:
# Description

**DO NOT MERGE**: planning to allow https://github.com/dhsc-govuk/FingertipsNext/pull/180 to merge first and then align this PR with the same approach.

Jira ticket: [DHSCFT-332](https://bjss-enterprise.atlassian.net/browse/DHSCFT-332)

This ticket adds initial setup for the trend analysis work, so that the HealthMeasure data will now have a trend dimension.

## Changes
- Creates new TrendDimension table with a relationship to the HealthMeasure table.
- Updates entities and repository models in the API along with any relevant mapping.
- Updated relevant unit tests to ensure the new trend data is retrieved and returned correctly.
- Updated API specification to account for changes.

## Validation

Screenshot on ticket of some manual testing I did using the API. I also did a fair bit of database verification to check the data had loaded correctly.

PR #199, opened 2025-02-28
DHSCFT-366: Created trend analysis skeleton application

dlzhry2:
# Description

Jira ticket: [DHSCFT-366](https://bjss-enterprise.atlassian.net/browse/DHSCFT-366)

Creates a skeleton dotnet console application for analysing trends.

As per ticket acceptance criteria, it is just a simple standalone application for now that retrieves data and maps it to an entity. The follow on tickets will implement the actual calculation and then the data steps to perform the full process i.e. retrieve all required data, calculate trends and then update the relevant rows in the DB. Tests will also come in that later PR.

## Validation

Ran locally against the DB deployed as per https://github.com/dhsc-govuk/FingertipsNext/pull/191. Works fine. README included :)


TrendAnalysisApp/Mapper/HealthMeasureMapper.cs
```diff
@@ -0,0 +1,54 @@
+using TrendAnalysisApp.Entity;
+using TrendAnalysisApp.Repository.Models;
+
+namespace TrendAnalysisApp.Mapper;
+
+/// <summary>
+/// The Health Measure Mapper class.
+/// </summary>
+public class HealthMeasureMapper
+{
```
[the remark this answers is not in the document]
dlzhry2: The class or the docstring? I'm probably too used to the linting rules on my last Java project which mandated that all classes have a docstring - I suppose it doesn't really add value - can remove where it's just echoing what we can already read :D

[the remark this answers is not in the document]
dlzhry2: I have removed now. Please re-review as there has been a refactor now to use Entity Framework. But I have removed the mappers as requested.

TrendAnalysisApp/Program.cs
```diff
@@ -0,0 +1,30 @@
+﻿using Microsoft.EntityFrameworkCore;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+using TrendAnalysisApp.Repository;
+namespace TrendAnalysisApp;
+
+internal class Program
```
HowardMayBjss, line 7: Sonar is calling out that this should have a 'static' or 'protected' statement

dlzhry2: Good spot. I had left it aside as this is the same approach used in the DataCreator console application.

I think it is right that we change it to static though, so I'll do it (and retest); adding a protected constructor wouldn't make sense for this class.

TrendAnalysisApp/README.md
```diff
@@ -0,0 +1,14 @@
...
+This is a dotnet console application to be used during the data loading process. It will calculate the latest trend for each Health Measure with unique dimensions - i.e. a given indicator e.g `Under 75 mortality rate from all causes` associated with a given area, sex and age group - and then set this value in the database. This will mean that the frontend can simply retrieve trend data and will not have to perform any calculations.
+
+## How to run
+It is not envisaged that developers will need to run this locally, as it is intended to be invoked as an automated pipeline job. However, if ever required locally e.g. for debugging please do the following:
+
+```
+Ensure the fingertips-db is already running
+cd into this directory
+
+export SQLCONNSTR_FINGERTIPS_DB={YOUR_LOCAL_CONN_STRING}
```
HowardMayBjss, line 12: Other projects use individual params e.g. DB_USER, DB_PASSWORD, DB_NAME - if possible might be worth aligning with this to keep peoples .env files clean

TrendAnalysisApp/Repository/HealthMeasureDbContext.cs
```diff
@@ -0,0 +1,18 @@
...
+
+namespace TrendAnalysisApp.Repository;
+
+public class HealthMeasureDbContext : DbContext
+{
+    public HealthMeasureDbContext() {}
+
+    public HealthMeasureDbContext(DbContextOptions options) : base(options) {}
+
+    public DbSet<HealthMeasureModel> HealthMeasure { get; set; }
```
HowardMayBjss, line 12: These Models seem to, largely, be duplicates of those in the api HealthData.Repository package. If that is the case then can we set up a shared package for use by both projects? @steves-bjss

TrendAnalysisApp/README.md
dlzhry2: Yes, I think it comes down to what is easiest for when we work on the pipeline ticket: https://bjss-enterprise.atlassian.net/browse/DHSCFT-367 As per the README apart from during initial development, this should not be something that developers are running locally, manually.

I think it comes down to what we see being easier in the pipeline @epercy-bjss - happy to take your steer. We could construct the value for `SQLCONNSTR_FINGERTIPS_DB` from the environment variables we already have for USER, PWD, DB NAME, TRUST CERT and so on. Or if you prefer I can align.

It is worth noting that this will require more code - see `api/DHSC.FingertipsNext.Modules.HealthData/Module.cs` as an example. The neat thing with using `SQLCONNSTR` is that you can do it in a nice little one liner: `configuration.GetConnectionString(Constants.Database.FingertipsDbName)`

TrendAnalysisApp/Repository/HealthMeasureDbContext.cs
dlzhry2: If it's not too much hassle, that could be a good idea. You're right - the models will be the same as in the API project as it's interacting the same Tables. The only real difference is 1. the type of query it will need to make 2. the fact that this project will actually update the data whereas the API is read-only.

steves-bjss: Usually it would be good to create a shared for the model classes, as with the schemas.  In this scenario, I'm okay with them being separate - longer term, the trend calculator application would be removed / changed, so we would be potentially compromising the API design for a temporary process.

TrendAnalysisApp/Repository/HealthMeasureRepository.cs
```diff
@@ -0,0 +1,68 @@
...
+    /// <param name="sexId"></param>
+    /// <returns>A list of Health Measure Models</returns>
+    public async Task<IEnumerable<HealthMeasureModel>> GetForUniqueDimension(
+        int ageId,
+        string areaCode,
+        int indicatorId,
+        int sexId
+    ) {
+            return await _dbContext.HealthMeasure
+            .Where(hm => hm.AgeDimension == null || hm.AgeDimension.AgeID == ageId)
```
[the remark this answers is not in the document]
dlzhry2: Good point - I've updated to match indicator ID by the IndicatorKey.

TrendAnalysisApp/Repository/HealthMeasureRepository.cs
```diff
@@ -0,0 +1,68 @@
...
+            .Where(hm => hm.AreaDimension == null || hm.AreaDimension.Code == areaCode)
+            .Where(hm => hm.IndicatorDimension == null || hm.IndicatorDimension.IndicatorId == indicatorId)
```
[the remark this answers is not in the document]
dlzhry2: Based on the chat we had with Emma, I've updated the query to look for Age and Sex where HasValue == false, since we only need to calculate trends breaking down the indicators by area. For sex and age, we just want all Persons and all Ages.

It might turn out to be slightly more complex when it comes to the data processing ticket, but hopefully this is good enough for the scope of this ticket. I retested and could see the data was returned.

steves-bjss (approved):

PR #201, opened 2025-02-28
DHSCFT-221: inequalities bar chart table component

adawari-thomas-bjss:
# Description

Jira ticket: [DHSCFT-221](https://bjss-enterprise.atlassian.net/browse/DHSCFT-221)

## Changes

- Created the inequalities bar chart table component
- Moved inequalities components into tabs

## Validation

validated against API

image


frontend/fingertips-frontend/components/organisms/Inequalities/index.tsx
```diff
@@ -1,17 +1,142 @@
...
+  const latestDataIndex = tableData.rowData.length - 1;
+  const barchartTableData: InequalitiesBarChartTableData = {
+    areaName: healthIndicatorData.areaName,
+    data: tableData.rowData[latestDataIndex],
+  };
+
   return (
     <div data-testid="inequalities-component">
-      <InequalitiesTable healthIndicatorData={healthIndicatorData} />
+      <H4>Inequalities data for a single time period</H4>
```
gareth-allan-bjss, line 93: The title I can see for this section in the designs is `Inequalities data in a single time period`:
```suggestion
      <H4>Inequalities data in a single time period</H4>
```

frontend/fingertips-frontend/components/organisms/Inequalities/index.tsx
```diff
@@ -1,17 +1,142 @@
...
+      <TabContainer
+        id="inequalitiesBarChartAndTable"
+        items={[
+          {
+            id: 'inequalitiesBarChart',
+            title: 'Line chart',
```
gareth-allan-bjss, line 99: This tab should say `Bar chart`:
```suggestion
            title: 'Bar chart',
```

gareth-allan-bjss (commented):
A couple of early comments from a quick comparison of the page to the designs (it looks nice with the tabs by the way!).

frontend/fingertips-frontend/components/molecules/Inequalities/LineChart/Table/index.tsx
```diff
@@ -77,16 +53,10 @@ const getCellHeader = (heading: string, index: number): ReactNode => {
 };
 
 export function InequalitiesTable({
```
gareth-allan-bjss, line 55: This function name (and the name of the props type) should probably be updated to reference the fact it's a line chart table now too:
```suggestion
export function InequalitiesLineChartTable({
```

frontend/fingertips-frontend/components/organisms/Inequalities/index.tsx
```diff
@@ -1,17 +1,142 @@
...
+
+export interface RowDataFields {
+  count?: number;
+  value?: number;
+  lower?: number;
+  upper?: number;
+}
+
+export interface InequalitiesTableRowData {
+  [key: string]: number | RowDataFields | undefined;
```
gareth-allan-bjss, line 38: Can this just be:
```suggestion
  [key: string]: RowDataFields | undefined;
```

I'm not sure you're ever using the form where the value is a number, are you?

frontend/fingertips-frontend/components/organisms/Inequalities/index.tsx
```diff
@@ -1,17 +1,142 @@
...
+            }
+          : undefined;
+        return acc;
+      },
+      {}
+    );
+
+    return { period: Number(key), ...dynamicFields };
+  });
+};
```
gareth-allan-bjss, line 68: I don't feel particularly strongly about this, so I'll leave it up to you, but we do also have `inequalitiesHelpers.ts` in this directory, so it might be more consistent to put this function (maybe the above types?) in it.

frontend/fingertips-frontend/components/organisms/Inequalities/inequalities.test.tsx
```diff
@@ -9,7 +13,75 @@ describe('Inequalities suite', () => {
...
+        2004: {
+          Male: [MOCK_HEALTH_DATA[0].healthData[1]],
+          Female: [MOCK_HEALTH_DATA[0].healthData[2]],
+        },
+        2008: {
+          All: [MOCK_HEALTH_DATA[1].healthData[0]],
+          Male: [MOCK_HEALTH_DATA[1].healthData[1]],
+          Female: [MOCK_HEALTH_DATA[1].healthData[2]],
+        },
+      };
```
gareth-allan-bjss, line 41: Could this use `GROUPED_YEAR_DATA`?

frontend/fingertips-frontend/components/molecules/Inequalities/BarChart/Table/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+  areaName: string
+): ReactNode => {
+  if (header === InequalitiesBarChartTableHeaders.COMPARED_TO)
+    return (
+      <StyledAlignRightHeader
+        style={{ width: '16%' }}
+        key={`heading-${header}`}
+        data-testid={`heading-${header}`}
+      >
+        {header} <br /> {areaName}
```
gareth-allan-bjss, line 46: Does this actually need a line break, or is that just how the text is wrapped in the designs?

frontend/fingertips-frontend/components/molecules/Inequalities/BarChart/Table/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+      </StyledAlignRightHeader>
+    );
+  return header === InequalitiesBarChartTableHeaders.GROUPING_TYPES ? (
+    <StyledAlignLeftHeader
+      key={`heading-${header}`}
+      data-testid={`heading-${header}`}
+    >
+      {header}
+    </StyledAlignLeftHeader>
+  ) : (
```
gareth-allan-bjss, line 56: It seems a bit strange that the grouping types header is left-aligned while all the others are right-aligned. Do we definitely want to do this?

frontend/fingertips-frontend/components/molecules/Inequalities/BarChart/Table/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+                </StyledAlignRightTableCell>
+                <StyledAlignRightTableCell>
+                  {getDisplayedValue(
+                    (tableData.data[key] as RowDataFields).upper
+                  )}
+                </StyledAlignRightTableCell>
+              </Table.Row>
+            ))}
+          </>
+        }
```
gareth-allan-bjss, line 121: I don't think these rows should be in the `Table` component's `head` property. Shouldn't they be children of the `Table` element?

frontend/fingertips-frontend/components/organisms/Inequalities/index.tsx
gareth-allan-bjss: I think this is some typing strangeness. If you make this interface as I suggested above it should mean you don't need to cast the values to `RowDataFields` in the table components. It just needs a bit of reworking of the mapping function. I had a quick plan and think I have something that works, so let me know if you'd like some help with it.

adawari-thomas-bjss: the period key is a number, while `Persons`, `Male`, and `Female` are of type `RowDataFields`. So deleting the number type would cause the object mapping to break. Happy speak about your plan to change the mapping function

frontend/fingertips-frontend/components/molecules/Inequalities/BarChart/Table/index.tsx
adawari-thomas-bjss: the first column is left aligned in the designs, which is why there is no left padding on the first header in the table designs. It is similar for the linecharts table. So, I think we definitely want this

frontend/fingertips-frontend/components/molecules/Inequalities/BarChart/Table/index.tsx
adawari-thomas-bjss: good spot, I completely missed this

frontend/fingertips-frontend/components/molecules/Inequalities/BarChart/Table/index.tsx
gareth-allan-bjss: OK, we'll be playing it back to Emma/Leonie to make sure it's OK anyway.

frontend/fingertips-frontend/components/molecules/Inequalities/BarChart/Table/index.tsx
gareth-allan-bjss: I think we just need to be careful not to do something because of a specific example in the designs. I think it's one we should check with Leonie.

frontend/fingertips-frontend/components/organisms/Inequalities/index.tsx
gareth-allan-bjss: Could `period` be defined as part of `RowDataFields`?

adawari-thomas-bjss: I'd rather keep `period` apart from the `RowDataFields` because each period corresponds to a different row under the period column, while the `RowDataFields` defines the data that the inequalities hold. Having `period` as part of `RowDataFields` duplicates it across each inequality for each row and would require extra logic to extract its value. I've taken a different approach that's closer to what was there initially, let me know if you're not happy about it.

frontend/fingertips-frontend/components/molecules/Inequalities/BarChart/Table/index.tsx
gareth-allan-bjss: We're not longer doing this, so marking it as resolved.

frontend/fingertips-frontend/components/organisms/Inequalities/index.tsx
gareth-allan-bjss: I'm happy with the approach you've taken :)

gareth-allan-bjss (approved):

PR #202, opened 2025-02-28
Feature/dhscft 390 inequalities tag on search results2

MonkeyDate:
# Description

Jira ticket: DHSCFT-390](https://bjss-enterprise.atlassian.net/browse/DHSCFT-390)

Update frontend to show 'has inequalities' tag to any indicator known to include inequalities data.

## Changes

- Update `IndicatorsDocument` definition to expose `hasInequalities`, matching the data returned from AI Search
- If `hasInequalities` is `true` for an indicator add the appropriate tag
- Make tag row layout dynamic so there are 0-3 columns, depending on data (only 2 tags available currently)
- Small styling changes so that multiple tags layout as-per figma
- Sneaky change to correct casing of text spotted during BA review but unrelated to this actual ticket.

## Validation

Manual testing of the various tag possibilities during BA review session.


frontend/fingertips-frontend/components/molecules/result/index.tsx
```diff
@@ -123,17 +127,22 @@ export function SearchResult({
             <StyledParagraph>{`Data period: ${formatDataPeriod(result.earliestDataPeriod, result.latestDataPeriod)}`}</StyledParagraph>
             <FinalParagraph>{`Last updated: ${formatDate(result.lastUpdatedDate)}`}</FinalParagraph>
             <TagRow>
-              <GridCol>
```
Joe-Hendley, line 126: My understanding was that we were to have the tags in 3 columns - has this changed?

MonkeyDate: No. There will be up to 3 tags. I changed the code so the columns only exist if there is a tag to go inside them.

frontend/fingertips-frontend/.run/All Tests.run.xml
```diff
@@ -0,0 +1,10 @@
+<component name="ProjectRunConfigurationManager">
```
dlzhry2, line 1: This file plus the other 2 in the `frontend/fingertips-frontend/.run/` directory are appearing in the Git diff. Is this something to do with how you're running tests locally, or is this a change we want for everyone in the repo.

If it isn't could be worth putting the .run dir into git ignore.

dlzhry2 (approved):

frontend/fingertips-frontend/.run/All Tests.run.xml
MonkeyDate: I'm not sure if we have an official view on this...... They are for Jetbrains specific IDEs. I'll ask the higher-ups

PR #208, opened 2025-03-03
Feature/dhscft tech - Refactoring of Health Data module / query

steves-bjss:
# Description

Tech debt to improve query performance and structure of the Health Data module.

## Changes

- Update query to offload processing to the database
- Update query to only retrieve needed fields

## Validation

Unit tests have been updated and executed.  A manual inspection of the queries generated have been undertaken for the different scenarios.


api/DHSC.FingertipsNext.Modules.HealthData.UnitTests/Repository/HealthDataRepositoryTests.cs
```diff
@@ -55,14 +55,14 @@ public async Task Repository_ShouldFindByOnlyIndicatorId_IfQueryParamsAreEmpty()
...
-        var result = await _repository.GetIndicatorDataAsync(expectedIndicatorId, [], [], []);
+        var result = await _healthDataRepository.GetIndicatorDataAsync(expectedIndicatorId, [], [], []);
 
         // assert
         result.ShouldNotBeEmpty();
         result.Count().ShouldBe(1);
         result.ShouldBeEquivalentTo(new List<HealthMeasureModel>()
         {
-            expectedHealthMeasure
+            ResetKeys(expectedHealthMeasure)
```
gareth-allan-bjss, line 65: Given we're building `expectedHealthMeasure` using the `HealthMeasureModelHelper`, should we be updating the helper to set appropriate default values for the keys, rather than calling `ResetKeys` to modify the object produced further?

api/DHSC.FingertipsNext.Modules.HealthData.UnitTests/Repository/HealthDataRepositoryTests.cs
```diff
@@ -229,15 +229,12 @@ public async Task Repository_ShouldOnlyIncludeResultsWithoutASexDimensionValue()
...
+        var result = await _healthDataRepository.GetIndicatorDataAsync(500, [], [], []);
 
         // assert
         result.ShouldNotBeEmpty();
         result.Count().ShouldBe(1);
-        result.ShouldBeEquivalentTo(new List<HealthMeasureModel>()
-        {
-            expectedHealthMeasure
-        });
+        result.First().ShouldBeEquivalentTo(ResetKeys(expectedHealthMeasure));
```
gareth-allan-bjss, line 237: The test below (`Repository_ShouldOnlyIncludeResultsWithoutAnAgeDimensionValue`) does essentially the same check, but you've not changed that one to check only the `First()` result. I think we want to be consistent here.

Given we're checking the number of results in this test I'm not concerned this has changed the behaviour, although I'm not see a huge benefit in changing this. Is it just because we know there'll only ever be one result, or am I missing something?

steves-bjss: You're right it should be consistant and was a good spot - I was originally looking at changing the way we assert these but decided against it and left them alone.  This straggler has been returned to its original state.

api/DHSC.FingertipsNext.Modules.HealthData.UnitTests/Repository/HealthDataRepositoryTests.cs
steves-bjss: The helper is used in mutliple places and the keys are integral to some of the other tests.  I went with RestKeys, rather than do a selective assertion so that any new fields are asserted against, rather than risk missing them out.

gareth-allan-bjss: Discussed offline, and the fact that we're using the objects created by the helper to populate the DB as well as create the expected results means we can't omit setting the keys.

We agreed a refactor of how test data is set up can be looked at in a subsequent PR.

gareth-allan-bjss (approved):

dlzhry2 (approved):

PR #209, opened 2025-03-04
DHSCFT-216: Make HealthData table reference DeprivationData

gareth-allan-bjss:
# Description

Jira ticket: [DHSCFT-216](https://bjss-enterprise.atlassian.net/browse/DHSCFT-216)

This PR makes the database's `HealthData` table reference the `DeprivationDimension` table which was added to the DB in #198.
## Changes

- Updated the `HealthMeasure` table in the DB to add a foreign key reference to the `DeprivationDimension` table (
  - Note: this defaults to referencing the "All" value, as this essentially represents a "no value" 
- Added a `DeprivationDimensionModel` class to represent the `DeprivationDimension` table
- Updated the `HealthMeasureModel` class to add the reference to the deprivation dimension
- Updated the API's `HealthDataRepository` to filter out any data points with a value for the deprivation dimension (the logic around allowing the deprivation data to be returned will be added in a subsequent PR, this is just to prevent any changes to the data being returned)
- Update `.gitignore` to remove the override that allowed some VSCode settings to be committed (and removed `.vscode/settings.json` from the repo)

## Validation
Ran the Docker compose command to populate the database. Ran this twice without tearing down the containers to verify that it functioned as expected when the database was empty and when it had existing data:
Screenshot 2025-03-04 at 10 51 15
Screenshot 2025-03-04 at 10 52 04

Confirmed that the API response is unchanged from main.
This branch:
Screenshot 2025-03-04 at 13 37 25
Main:
Screenshot 2025-03-04 at 13 37 32

Confirmed using the output from the API when navigating to http://localhost:5144/indicators/108/data?area_codes=E12000001&area_codes=E92000001&inequalities=sex that the deprivation dimension is included in the SQL query:
```
info: Microsoft.EntityFrameworkCore.Database.Command[20101]
      Executed DbCommand (52ms) [Parameters=[@__indicatorId_0='?' (DbType = Int32), @__areaCodes_1='?' (Size = 4000)], CommandType='Text', CommandTimeout='30']
      SELECT [h].[Year], [h].[Value], [h].[Count], [h].[LowerCi], [h].[UpperCi], [a1].[Name], [a1].[HasValue], [s].[Name], [s].[HasValue], [i].[Name], [a].[Code], [a].[Name], [t].[Name], [d].[Name], [d].[Type], [d].[Sequence], [d].[HasValue]
      FROM [HealthMeasure] AS [h]
      INNER JOIN [IndicatorDimension] AS [i] ON [h].[IndicatorKey] = [i].[IndicatorKey]
      INNER JOIN [AreaDimension] AS [a] ON [h].[AreaKey] = [a].[AreaKey]
      INNER JOIN [AgeDimension] AS [a1] ON [h].[AgeKey] = [a1].[AgeKey]
      INNER JOIN [DeprivationDimension] AS [d] ON [h].[DeprivationKey] = [d].[DeprivationKey]
      INNER JOIN [SexDimension] AS [s] ON [h].[SexKey] = [s].[SexKey]
      INNER JOIN [TrendDimension] AS [t] ON [h].[TrendKey] = [t].[TrendKey]
      WHERE [i].[IndicatorId] = @__indicatorId_0 AND [a].[Code] IN (
          SELECT [a0].[value]
          FROM OPENJSON(@__areaCodes_1) WITH ([value] nvarchar(20) '$') AS [a0]
      ) AND [a1].[HasValue] = CAST(0 AS bit) AND [d].[HasValue] = CAST(0 AS bit)
      ORDER BY [h].[Year]
```

Verified the line chart/table has not been changed when compared to main.
This branch:
Screenshot 2025-03-04 at 14 05 12
Screenshot 2025-03-04 at 14 05 23

main:
Screenshot 2025-03-04 at 14 06 01
Screenshot 2025-03-04 at 14 06 19


api/DHSC.FingertipsNext.Modules.HealthData.UnitTests/Helpers/HealthMeasureModelHelper.cs
```diff
@@ -162,11 +194,13 @@ public HealthMeasureModel Build()
...
             AgeDimension = ageDimension,
             IndicatorDimension = indicatorDimension,
             SexDimension = sexDimension,
-            TrendDimension = trendDimension
+            TrendDimension = trendDimension,
+            DeprivationDimension = deprivationDimension
         };
     }
-}
+}
```
dlzhry2, line 206: Minor, but file should have new line.

dlzhry2 (approved):
Looks good to me - will ask questions in person to get a bit more understanding on how this will fit in with the new data loading work.

api/DHSC.FingertipsNext.Modules.HealthData.Repository/HealthDataRepository.cs
```diff
@@ -19,44 +19,54 @@ public async Task<IEnumerable<HealthMeasureModel>> GetIndicatorDataAsync(int ind
             .Where(hm => hm.IndicatorDimension.IndicatorId == indicatorId)
             .Where(hm => areaCodes.Length == 0 || areaCodes.Contains(hm.AreaDimension.Code))
             .Where(hm => years.Length == 0 || years.Contains(hm.Year))
-            .Where(hm => inequalities.Contains("sex") ? true: hm.SexDimension.HasValue == false)
-            .Where(hm => inequalities.Contains("age") ? true: hm.AgeDimension.HasValue == false)
+            .Where(hm => inequalities.Contains("sex") ? true : hm.SexDimension.HasValue == false)
+            .Where(hm => inequalities.Contains("age") ? true : hm.AgeDimension.HasValue == false)
+            .Where(hm => hm.DeprivationDimension.HasValue == false)
```
dlzhry2, line 24: Is it worth having a TODO with the ticket reference for why this is set to false for now? I know it'll trigger a SonarCloud issue, but in this case it seems potentially worthwhile.

api/DHSC.FingertipsNext.Modules.HealthData.Repository/Models/DeprivationDimensionModel.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace DHSC.FingertipsNext.Modules.HealthData.Repository.Models;
+
+[Serializable]
+public class DeprivationDimensionModel
+{
+    [Key]
+    public short DeprivationKey { get; set; }
+    [MaxLength(100)]
+    public string Name { get; set; }
```
HowardMayBjss, line 11: Is there any reason for not following sonarqubes suggestion and making this field required?

HowardMayBjss: And likewise for others

HowardMayBjss (approved):

api/DHSC.FingertipsNext.Modules.HealthData.Repository/Models/DeprivationDimensionModel.cs
gareth-allan-bjss: I believe @steves-bjss has found that doing so prevents us from omitting the fields from SQL queries, and we were getting a lot of data we didn't need.

I think he'd mentioned we should probably look at out .Net project settings to see whether this warning could be suppressed (I think this is actually Github reporting the .Net build warnings, rather than Sonarqube flagging it).

PR #213, opened 2025-03-05
DHSCFT-369: Setup trend calculator using legacy code

dlzhry2:
# Description

Jira ticket: [DHSCFT-369](https://bjss-enterprise.atlassian.net/browse/DHSCFT-369)

This PR moves the legacy code for the trend calculator into our new Trend Analysis application. A TrendCalculator has been created. Under the hood, it maps the entities to the required objects for the legacy calculator, but in future this mapping could be removed and the calculation re-implemented in the new calculator.

For now, the TrendCalculator is only invoked by the unit tests. In the next ticket (DHSCFT-374) we will plug this into the console application so that the trend calculator is actually used to calculate the trends during the data processing step.

## Changes

- Created a TrendCalculator class which will return the trend for a given indicator + the most recent health measure data points.
- Created a Legacy Mapper class for the trend calculator to use to convert between our models and the legacy code models.
- Create unit tests to validate the mapping functionality and the trend results returned from the calculation.
- Added pipeline logic so that unit tests in the new trend-analysis console application directory will be executed when required.

## Validation
This will need a second pair of eyes.

Currently, the Trend Calculator is only called in the unit tests. I have put together a suite a tests that use a variety of indicators to validate the trend result.

In each test you will see a link to the relevant indicator and a description of specifically what indicator + set of dimensions was used. I have noted the trend that is displayed on the FingerTips website and copied the relevant data from the underlying CSV file for the table.

I have then validated that the expected Trend is returned, as per the website. I have ensured that I have a test for each of the different trend types in addition to a good variety of Indicator value types such as 'Crude rate', 'Proportion', 'Indirectly standardised proportion' and 'Directly standardised proportion'.


trend-analysis/TrendAnalysisApp/Calculator/Legacy/Constants.cs
```diff
@@ -0,0 +1,44 @@
+namespace TrendAnalysisApp.Calculator.Legacy;
+
+public class ValueTypeId
```
dlzhry2, line 3: Unused code - could be removed from Legacy?

trend-analysis/TrendAnalysisApp/Calculator/Legacy/Models/CoreDataSet.cs
```diff
@@ -0,0 +1,32 @@
+﻿using Newtonsoft.Json;
+
+namespace TrendAnalysisApp.Calculator.Legacy.Models
+{
+  public class CoreDataSet : ValueWithCIsData, ICloneable
+    {
+        private double? countPerYear;
```
dlzhry2, line 7: Unused - remove?

trend-analysis/TrendAnalysisApp/Calculator/Trends.cs
```diff
@@ -0,0 +1,8 @@
+namespace TrendAnalysisApp.Calculator;
+
+public enum Trend {
```
dlzhry2, line 3: Currently planning to add polarity in the follow-on PR when data will be available and can test it out with real data. And as this PR is already quite large.

trend-analysis/TrendAnalysisApp/Calculator/Legacy/Constants.cs
steves-bjss: Lets remove anything we can safely can but we shouldn't refactor the calculator to pass SonarQube - we have no reasonable way of validating we haven't changed the behaviour of it.

trend-analysis/TrendAnalysisApp/Calculator/Legacy/Models/CoreDataSet.cs
steves-bjss: Should be safe to remove

dlzhry2: Removed.

trend-analysis/TrendAnalysisApp/Calculator/Legacy/Constants.cs
dlzhry2: I've updated what I could in the most recent commit. Please check it looks ok to you.

steves-bjss (approved):

PR #211, opened 2025-03-04
Feature/dhscft 411 one indicator one area view line chart

WolfieKnee:
# Description

Jira ticket: [DHSCFT-411](https://bjss-enterprise.atlassian.net/browse/DHSCFT-411)

move lineChart to OneIndicatorOneArea view
## Changes

- Add data fetching to View with corresponding tests
- Add data preparation and rendering of LineChart to ViewPlot with corresponding tests
- Remove LineChart from Chart and update unit tests accordingly
- Harmonise headings and remove skeleton text from other Views
- added 2 more of the core journeys into the e2e tests as discussed with @williamwillis11 

## Validation

All tests passing and behaviour as expected
Functionally will show no data on linechart due to changes in data and filtering for inequalities - this is to be addressed in other tickets e.g. [DHSCFT-442](https://bjss-enterprise.atlassian.net/browse/DHSCFT-442)

PR #214, opened 2025-03-05
Feature/dhscft 240 show value types

obarojohnson:
# Description

Jira ticket: [DHSCFT-240](https://bjss-enterprise.atlassian.net/browse/DHSCFT-240)

This ticket will primarily focus on the line charts and table then we can apply any logic going forward for further visualisations. The indicator data uses a particular unit of measurement e.g. 

## Changes
image

## LineChart
image

##Bar chart
image

- Add a new variable to the data object called "measurementUnit"
- Refactor the Linechart to make it easy to read. I believe this will have to change in the future.
- Linechart Y-Label added and also units on the values when hover has be added also.
- Barchart has now have a value unit to them.
- Snapshort now added


frontend/fingertips-frontend/components/pages/chart/index.tsx
```diff
@@ -105,6 +105,8 @@ export function Chart({
                     searchState={searchState}
                     groupIndicatorData={groupData}
                     xAxisTitle="Year"
+                    yAxisTitle="Unit"
```
gareth-allan-bjss, line 108: I don't think the y-axis title should be the literal text `Unit`. According to AC2 on the ticket, I think it should be:
```suggestion
                    yAxisTitle={indicatorMetadata?.unitLabel ? `Value: ${indicatorMetadata?.unitLabel}` : ''}
```

gareth-allan-bjss:
@obarojohnson: I think you need to merge main into this branch, to pick up the changes that moved the line chart into a view as part of #211.

PR #219, opened 2025-03-05
DHSCFT-417: tidy up area suggestions

alibombaye:
# Description

Jira ticket: [DHSCFT-417](https://bjss-enterprise.atlassian.net/browse/DHSCFT-417)

- Refactors of the area Search Suggestions.

Screenshot 2025-03-05 at 15 06 29

Screenshot 2025-03-05 at 15 06 18

Screenshot 2025-03-05 at 15 06 08

## Validation

Tests all passing, manual regression testing


frontend/fingertips-frontend/components/molecules/AreaSelectedPill/AreaSelectedPill.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+    name: 'Test Area',
+    typeName: gpsAreaType.name,
+    typeKey: gpsAreaType.key as AreaTypeKeys,
+  };
+
+  beforeEach(() => {
+    mockRemoveFilter.mockClear();
+  });
+
+  it('should render the AreaSearchPill correctly', () => {
```
williamwillis11, line 19: typo for AreaSelectedPill?

frontend/fingertips-frontend/components/molecules/AreaAutoCompleteInputField/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+  useEffect(() => {
+    if (firstSelectedArea) {
+      setCriteria(firstSelectedArea?.name ?? '');
+    }
+
+    const fetchSearchArea = async (criteria: string) => {
+      if (criteria) {
+        const areas = await getSearchSuggestions(criteria);
+        setSearchAreas(areas.slice(0, 20));
```
HowardMayBjss, line 43: while we are tidying this us would be good to pull '20' into a const definition

frontend/fingertips-frontend/components/molecules/AreaAutoCompleteInputField/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+      (criteria: string) => {
+        if (
+          firstSelectedArea ||
+          criteria == null ||
+          criteria.length < MIN_SEARCH_SIZE
+        ) {
+          setSearchAreas([]);
+          return;
+        }
+        void fetchSearchArea(criteria);
```
HowardMayBjss, line 65: is this trying to call fetchSearchArea? If so why is void in front of it?

frontend/fingertips-frontend/components/molecules/AreaAutoCompleteInputField/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+    const fetchCleanUp = () => {
+      if (timeoutRef.current != null) {
+        clearTimeout(timeoutRef.current);
+        timeoutRef.current = null;
+      }
+    };
+    fetchCleanUp();
+
+    timeoutRef.current = setTimeout(
```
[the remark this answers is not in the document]
alibombaye: Yes I believe that is correct. This is to prevent calling AI search for every key stroke in case a user is typing way too fast or trying to cause a DDOS

frontend/fingertips-frontend/components/molecules/AreaAutoCompleteInputField/index.tsx
alibombaye: that's a good spot. I'm not sure. I'll remove and see what affect that has.

frontend/fingertips-frontend/components/molecules/AreaAutoCompleteInputField/index.tsx
alibombaye: Add a const for this.

frontend/fingertips-frontend/components/molecules/AreaAutoCompleteInputField/index.tsx
alibombaye: Removed 👍

HowardMayBjss (approved):

frontend/fingertips-frontend/app/page.tsx
```diff
@@ -19,17 +22,40 @@ export default async function Page(
...
-      initialFormState={initialState}
-      searchState={stateManager.getSearchState()}
-    />
-  );
+    const areasApi = ApiClientFactory.getAreasApiClient();
+
+    const selectedAreasData =
+      areasSelected && areasSelected.length > 0
+        ? await Promise.all(
+            areasSelected.map((area) => areasApi.getArea({ areaCode: area }))
```
obarojohnson, line 33: This function ` areasApi.getArea({ areaCode: area })) `is pointing to a different data server than the auto search function; both should point to the same data source. For this reason, if you start the frontend with 'npm run dev', the autocomplete search works, but this function will throw an exception, causing unexpected behavior that is inconsistent with the auto search list.

frontend/fingertips-frontend/components/molecules/AreaAutoCompleteInputField/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+  const timeoutRef = useRef<number | null>(null);
+
+  useEffect(() => {
+    if (firstSelectedArea) {
+      setCriteria(firstSelectedArea?.name ?? '');
+    }
+
+    const fetchSearchArea = async (criteria: string) => {
+      if (criteria) {
+        const areas = await getSearchSuggestions(criteria);
```
obarojohnson, line 43: Should the `getSearchSuggestions` function point to the same data source as the `getArea` function called in the parent component? Otherwise, it causes inconsistency, making users wonder why they are getting a search result when the server is not running on the auto search panel, while the getArea function currently expects the API server to be running.

frontend/fingertips-frontend/components/forms/SearchForm/index.tsx
```diff
@@ -87,17 +55,24 @@ export const SearchForm = ({
...
+        value={selectedAreas || ''}
       />
       <AreaAutoCompleteInputField
-        onAreaSelected={(area: AreaDocument | undefined) => {
-          updateUrlWithSelectedArea(area?.areaCode ?? '');
-          setAreaCode(area?.areaCode ?? '');
-        }}
         inputFieldErrorStatus={!!formState.message}
-        defaultSelectedAreas={defaultAreas}
+        searchState={searchState}
```
obarojohnson, line 62: Select a search item from the suggestion list ,  and then delete the selected pill. This causes the SuggestionPanel to auto open again. 

Is this an expected behave  , see the snapshot 
image

frontend/fingertips-frontend/components/molecules/AreaSelectedPill/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+interface AreaSelectedPillProps {
+  areaName: string;
+  areaCode: string;
+  areaTypeKey: AreaTypeKeys;
+  areaTypeName: string;
+  onRemoveFilter: (filterID: string) => void;
+  inFilterPane?: boolean;
+}
+
+export const AreaSelectedPill = ({
```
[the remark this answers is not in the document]
alibombaye: Good question and I think this can be changed to pass a single selectedAreaData prop. The pill will use the the model `AreaWithRelations` rather than `AreaDocument`. 

Again this is just so we can re-use this component on other pages where we don't have the `AreaDocument`.

But a single prop can be provided here rather that a multi-fields prop 👍

frontend/fingertips-frontend/app/page.tsx
alibombaye: Yes once we have an areaSelected via the autoSearch that is the job of the AI area search complete. From then on we want to use the areas-api to get the area data. 

The AI search should be populated with the same area data that exists in the API and our database.

This was again to re-use the mechanism that has already been put in place for the area filtering on the results page.

frontend/fingertips-frontend/components/molecules/AreaAutoCompleteInputField/index.tsx
alibombaye: This is the correct data source here since we want to use the AI search data-source to fetch the suggested areas.

frontend/fingertips-frontend/components/forms/SearchForm/index.tsx
alibombaye: Good spot. I will run this by BA and design to see if this is expected behaviour or if we want to do something else in this situation and tackle this separately  👍 .

PR #220, opened 2025-03-05
DHSCFT-309: Inequalities line chart component

adawari-thomas-bjss:
# Description

Jira ticket: [DHSCFT-309](https://bjss-enterprise.atlassian.net/browse/DHSCFT-309)

## Changes

- Created the line chart component for inequalities

## Validation

validated against API

image


frontend/fingertips-frontend/lib/chartHelpers/pointFormatterHelper.ts
```diff
@@ -1,28 +1,29 @@
-enum SymbolsEnum {
+// any required to allow customisation of Highcharts tooltips
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+export enum SymbolsEnum {
```
gareth-allan-bjss, line 4: Do you need to export this enum? It's not actually used outside this file.

frontend/fingertips-frontend/lib/chartHelpers/pointFormatterHelper.ts
```diff
@@ -1,28 +1,29 @@
...
   Circle = '\u25CF',
   Square = '\u25a0',
   Diamond = '\u25c6',
   Triangle = '\u25b2',
   TriangleDown = '\u25bC',
 }
 
-const symbolEncoder: Record<string, string> = {
+export const symbolEncoder: Record<string, string> = {
```
gareth-allan-bjss, line 12: Again, do you need to export this, given it's only used within the file?

frontend/fingertips-frontend/lib/chartHelpers/pointFormatterHelper.ts
```diff
@@ -1,28 +1,29 @@
...
   'triangle': SymbolsEnum.Triangle,
   'triangle-down': SymbolsEnum.TriangleDown,
 };
 
-// any required to allow customisation of Highcharts tooltips
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export const pointFormatterHelper = (point: any) => {
+export const pointFormatterHelper = (
+  point: any,
+  generateTooltipList: (point: any, symbol?: string) => string[]
```
gareth-allan-bjss, line 22: Does symbol need to be optional here? It should always be set, as you default it to circle.

frontend/fingertips-frontend/lib/chartHelpers/pointFormatterHelper.test.ts
```diff
@@ -13,7 +14,10 @@ const mockPoint = {
 
 describe('pointFormatterHelper', () => {
   it('should return a string with the default symbol code for circle when symbolName is empty string', () => {
-    const actual = pointFormatterHelper(mockPoint);
+    const actual = pointFormatterHelper(
+      mockPoint,
+      generatePopPyramidTooltipStringList
```
gareth-allan-bjss, line 19: I think rather than using an actual implementation of `generatePopPyramidTooltipStringList` here, you'd be better passing a mock function in and verifying it was called with the correct symbol value, and that the value it returns is returned from the function. That way, if the implementation of `generatePopPyramidTooltipStringList` changes, you don't risk this test failing.

frontend/fingertips-frontend/components/molecules/Inequalities/LineChart/index.tsx
```diff
@@ -0,0 +1,88 @@
...
+  <span style="color: ${point.series.color}; font-weight: bold;">${symbol}</span></div>`,
+  `<div><span>${point.series.name}</br>Value: ${point.y}</span></div></div>`,
+];
+
+export function InequalitiesLineChart({
+  lineChartData,
+  dynamicKeys,
+  areasSelected,
+  type = InequalitiesTypes.Sex,
+  yAxisTitleText = 'Value',
```
gareth-allan-bjss, line 35: This Y-axis label should ultimately also include the unit label that @obarojohnson is working on as part of #232. We'll need to co-ordinate with him to make sure this doesn't get missed.

frontend/fingertips-frontend/components/molecules/Inequalities/LineChart/index.tsx
```diff
@@ -0,0 +1,88 @@
...
+  yAxisTitleText?: string;
+}
+
+const generateInequalitiesLineChartTooltipStringList = (
+  point: Highcharts.Point,
+  symbol?: string
+) => [
+  `<div style="display: flex; margin-top: 7px; align-items: center;"><div style="margin-right: 10px;">
+  <span style="color: ${point.series.color}; font-weight: bold;">${symbol}</span></div>`,
+  `<div><span>${point.series.name}</br>Value: ${point.y}</span></div></div>`,
```
gareth-allan-bjss, line 27: This tooltip should also show the unit label from #232.

frontend/fingertips-frontend/components/organisms/Inequalities/inequalitiesHelpers.ts
```diff
@@ -35,22 +39,50 @@ export enum Sex {
...
+  [InequalitiesTypes.Deprivation]: (keys: string[]) => keys,
+};
+
+export const inequalitiesBenchmarkColumnMapping: Record<
+  InequalitiesTypes,
+  string
+> = {
+  [InequalitiesTypes.Sex]: 'Persons',
+  [InequalitiesTypes.Deprivation]: 'England',
 };
```
gareth-allan-bjss, line 86: Is `inequalitiesBenchmarkColumnMapping` used? I can't find a reference to it.

frontend/fingertips-frontend/components/organisms/Inequalities/inequalitiesHelpers.ts
```diff
@@ -35,22 +39,50 @@ export enum Sex {
...
+  string[]
+> = {
+  [InequalitiesTypes.Sex]: [
+    ChartColours.Orange,
+    ChartColours.OtherLightBlue,
+    ChartColours.Purple,
+  ],
+  [InequalitiesTypes.Deprivation]: [],
+};
+
```
gareth-allan-bjss, line 66: I'm not sure either of these functions need to be exported, given they're not used outside this file.

gareth-allan-bjss, line 66: I think sex inequalities is something of the odd one out, in that it doesn't appear to follow the same order for line shapes/colours as the deprivation inequality and 'standard' line chart. Which makes it a bit awkward and means we need to do this separate colour/shape mapping. I think we should speak to design to see if we really need to do this.

frontend/fingertips-frontend/components/pages/chart/index.tsx
```diff
@@ -141,6 +141,7 @@ export function Chart({
               ? dataWithoutEngland[0]
               : healthIndicatorData[0][0]
           }
+          areasSelected={searchState[SearchParams.AreasSelected]}
```
gareth-allan-bjss, line 144: I wonder at this point whether we should be passing the whole `searchState` object to the `Inequalities` component, as we do with the `LineChart`?

frontend/fingertips-frontend/components/molecules/Inequalities/LineChart/index.tsx
obarojohnson: if you want to add the ability for the unit label , you can just add additional parameter to the  `generateInequalitiesLineChartTooltipStringLis`t  measurementUnit there and pass it via the component.

frontend/fingertips-frontend/components/molecules/Inequalities/LineChart/index.tsx
obarojohnson: InequalitiesLineChart  you can add another variable here to be consistence and this variable should be feed to your above function called 'measurementUnit" as string type.

frontend/fingertips-frontend/components/molecules/Inequalities/LineChart/index.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+  const lineChartOptions: Highcharts.Options = {
+    ...lineChartDefaultOptions,
+    yAxis: {
+      title: { text: yAxisTitleText, margin: 20 },
+      minorTickInterval: 'auto',
+      minorTicksPerMajor: 2,
+    },
+    xAxis: {
+      title: { text: 'Year', margin: 20 },
```
obarojohnson, line 52: This might be already discussed and agreed upon , so forgive me . Would it be a nice Idea to make the xAxis **text** a props with a default value to "Year" instead of the fixed hardcoded one to provide the ability to configured it in the future if a different text is needed just like the yAxis. Unless of course this will not changed.

frontend/fingertips-frontend/components/molecules/Inequalities/LineChart/inequalitiesLineChart.test.tsx
```diff
@@ -0,0 +1,23 @@
...
+      />
+    );
+
+    expect(
+      screen.getByTestId('inequalitiesLineChart-component')
+    ).toBeInTheDocument();
+    expect(
+      screen.getByTestId('highcharts-react-component-inequalitiesLineChart')
+    ).toBeInTheDocument();
+  });
```
obarojohnson, line 22: Can you add another test to check what happen when no areaSelected has be selected  and also one that check what if dynamicKeys  is empty.

obarojohnson (changes requested):

frontend/fingertips-frontend/components/molecules/Inequalities/LineChart/inequalitiesLineChart.test.tsx
adawari-thomas-bjss: I agree that this is a good test scenario, but it can't be done on this component as we do not visually test Highcharts or its functionality in unit tests. So, I've added extra test cases to the inequalities tables and the helper method to generate the inequalities line chart data to show the effect of having empty dynamicKeys and areaSelected lists

frontend/fingertips-frontend/components/molecules/Inequalities/LineChart/index.tsx
adawari-thomas-bjss: done

frontend/fingertips-frontend/components/molecules/Inequalities/LineChart/index.tsx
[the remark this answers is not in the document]
adawari-thomas-bjss: They are also different in the way that they handle data. We could create a common interface to cause the line chart data to be generic and move all the computation logic up a level, but I believe this might be time-consuming. So, I've opted for a separate component

frontend/fingertips-frontend/components/molecules/Inequalities/LineChart/index.tsx
gareth-allan-bjss: #232 has now gone in.

frontend/fingertips-frontend/components/molecules/Inequalities/LineChart/index.tsx
```diff
@@ -66,9 +73,11 @@ export function InequalitiesLineChart({
...
         '<span>Year {point.x}</span><br/>',
       pointFormatter: function (this: Highcharts.Point) {
-        return pointFormatterHelper(
-          this,
-          generateInequalitiesLineChartTooltipStringList
+        return (
+          pointFormatterHelper(
+            this,
+            generateInequalitiesLineChartTooltipStringList
+          ) + ` ${measurementUnit}</span></div></div>`
```
gareth-allan-bjss, line 80: I think we should probably have a ternary here to make sure there isn't a trailing space after the value if no measurement unit is provided.

frontend/fingertips-frontend/components/molecules/Inequalities/LineChart/index.tsx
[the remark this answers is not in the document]
gareth-allan-bjss: I don't think we need a new interface. The data is processed into `Highcharts.SeriesOptionsType[]` in both cases, so that could be passed to a common line chart component.

I'll approve this PR, as we don't have to do it here, but I think we should bear this in mind as we expand the functionality of the inequalities line chart (e.g. to add support for showing CI error bars, etc.). I think we'll find they're doing very similar things.

gareth-allan-bjss (approved):

gareth-allan-bjss (approved):

obarojohnson (approved):

PR #221, opened 2025-03-05
DHSCFT-221: BA review

adawari-thomas-bjss:
# Description

Jira ticket: [DHSCFT-221](https://bjss-enterprise.atlassian.net/browse/DHSCFT-221)

## Changes

- Made suggested changes following BA review

## Validation

image

### DIFF
--- a/frontend/fingertips-frontend/components/organisms/Inequalities/index.tsx
+++ b/frontend/fingertips-frontend/components/organisms/Inequalities/index.tsx
@@ -50,7 +50,7 @@ export function Inequalities({
           },
           {
             id: 'inequalitiesBarChartTable',
-            title: 'Tabular data',
+            title: 'Table',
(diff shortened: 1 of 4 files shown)


frontend/fingertips-frontend/components/organisms/Inequalities/index.tsx
```diff
@@ -50,7 +50,7 @@ export function Inequalities({
           },
           {
             id: 'inequalitiesBarChartTable',
-            title: 'Tabular data',
+            title: 'Table',
```
gareth-allan-bjss, line 53: I think the line chart's tab still says `Tabular data`, so we should presumably update that to `Table` as well, in order to keep things consistent.

gareth-allan-bjss (approved):

PR #222, opened 2025-03-05
DHSCFT-438: fix for real api data

alibombaye:
# Description

Jira ticket: [DHSCFT-438](https://bjss-enterprise.atlassian.net/browse/DHSCFT-438)

- Align Area Type data with the API
- Fix groupType to be autopicked from the first availableGroupTypes

Screenshot 2025-03-05 at 17 30 46

## Validation

All tests passing. Tested against the real api. Area filtering working as expected

### DIFF
--- a/frontend/fingertips-frontend/lib/areaFilterHelpers/areaType.ts
+++ b/frontend/fingertips-frontend/lib/areaFilterHelpers/areaType.ts
@@ -3,21 +3,21 @@ import { AreaType } from '@/generated-sources/ft-api-client';
 export const combinedAuthoritiesAreaType = {
   key: 'combined-authorities',
   name: 'Combined Authorities',
-  hierarchyName: 'Admin',
+  hierarchyName: 'Administrative',
   level: 3,
 };
 
 export const countiesAndUnitaryAuthoritiesAreaType = {
   key: 'counties-and-unitary-authorities',
   name: 'Counties and Unitary Authorities',
-  hierarchyName: 'Admin',
+  hierarchyName: 'Administrative',
   level: 4,
 };
 
 export const districtAndUnitaryAuthoritiesAreaType = {
   key: 'districts-and-unitary-authorities',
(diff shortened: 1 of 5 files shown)


frontend/fingertips-frontend/lib/areaFilterHelpers/determineSelectedGroupType.test.ts
```diff
@@ -9,7 +14,17 @@ describe('DetermineSelectedGroupType', () => {
     expect(selectedGroupType).toEqual('nhs-integrated-care-boards');
   });
 
-  it('should return "england" as the default group type when selectedGroupType is not provided', () => {
+  it('should return the first areaType from the availableGroupTypes provided', () => {
```
williamwillis11, line 17: do we need a looping test for each area type? ie check that it picks the first group type for each area type?

alibombaye: No we don't the Group type is a subset of all areaTypes. We already provide this function with a sorted list of available groupTypes that can be picked. We just want to pick the first on in the array provided as the default option.

frontend/fingertips-frontend/lib/areaFilterHelpers/areaType.ts
```diff
@@ -59,14 +59,14 @@ export const nhsSubIntegratedCareBoardsAreaType = {
...
-  hierarchyName: 'Admin',
+  hierarchyName: 'Administrative',
   level: 2,
 };
 
 export const englandAreaType = {
   key: 'england',
   name: 'England',
-  hierarchyName: 'All',
+  hierarchyName: 'Both',
```
MonkeyDate, line 69: Have you performed a search to check that we've not missed any of these name changes we might have hard coded?

MonkeyDate (approved):

PR #203, opened 2025-02-28
DHSCFT-287 add chart component screenshot testing

williamwillis11:
# Description

Jira ticket: [DHSCFT-287](https://bjss-enterprise.atlassian.net/browse/DHSCFT-287)

Added visual screenshot comparison testing of the chart components for E2E tests when they are executed in CI against docker. Note that this testing is not performed when running the e2e tests locally or in CD mode against azure, and that the screenshots are stored in cache, not in the repo.

Currently the assertion is in a try catch as agreed in the demo.

## Changes

- added screenshot comparison of chart components current against expected, stored in github cache - currently warning only
- added manual workflow job to allow updating of the base screenshots against a branch
- this workflow also creates the base snapshots for a branch when the branch is opened
- removed double cache of the node dependencies in the ui and e2e yml templates
- changes to playwright config to support screenshot testing
- README updates to describe screenshot testing
- fixed a bug where we were not checking the line chart table for visibility

## Validation

E2E tests pass locally and in CI
UI tests pass locally and in CI

PR #223, opened 2025-03-06
DHSCFT-287 fix to not run the create screenshot job in CD

williamwillis11:
# Description

Jira ticket: [DHSCFT-287](https://bjss-enterprise.atlassian.net/browse/DHSCFT-287)
By design we arnt running the screenshot comparison checks in the e2e test run against azure (in CD) but i forgot in https://github.com/dhsc-govuk/FingertipsNext/pull/203 to add a rule to that step to stop that job from running. This PR adds that.

kicked off a pipeline here (as yml changes dont) - https://github.com/dhsc-govuk/FingertipsNext/actions/runs/13696786879

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -72,11 +72,12 @@ jobs:
         if: inputs.e2e-test-container-image == null
         run: npx playwright install --with-deps chromium webkit
 
-      - name: Create base snapshots if none in cache, or during manual snapshot update or when merging to main
+      - name: Create base snapshots if none in cache, or during manual snapshot update or when merging to main - for CI test execution
         if: >
-          steps.cache.outputs.cache-hit != 'true' ||
+          inputs.e2e-test-container-image == null &&
+          (steps.cache.outputs.cache-hit != 'true' ||
           inputs.update-screenshots == 'true' ||
-          github.ref == 'refs/heads/main'
+          github.ref == 'refs/heads/main')
         run: npm run test-e2e-ci-update-snapshots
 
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e-tests.yml
```diff
@@ -72,11 +72,12 @@ jobs:
         if: inputs.e2e-test-container-image == null
         run: npx playwright install --with-deps chromium webkit
 
-      - name: Create base snapshots if none in cache, or during manual snapshot update or when merging to main
+      - name: Create base snapshots if none in cache, or during manual snapshot update or when merging to main - for CI test execution
         if: >
-          steps.cache.outputs.cache-hit != 'true' ||
+          inputs.e2e-test-container-image == null &&
```
epercy-bjss, line 77: Please can you indent the conditions, so it reads like:

```
inputs.e2e-test-container-image == null &&
(
  steps.cache.outputs.cache-hit != 'true' ||
  inputs.update-screenshots == 'true' ||
  github.ref == 'refs/heads/main'
)
```

.github/workflows/e2e-tests.yml
```diff
@@ -72,11 +72,12 @@ jobs:
...
+          (steps.cache.outputs.cache-hit != 'true' ||
           inputs.update-screenshots == 'true' ||
-          github.ref == 'refs/heads/main'
+          github.ref == 'refs/heads/main')
```
epercy-bjss: This will not run for PR's due to the github ref being different for PR runs. Is that what you are intending?

epercy-bjss (approved):

PR #231, opened 2025-03-07
Feature/dhscft 385 indicator details page

Joe-Hendley:
https://bjss-enterprise.atlassian.net/browse/DHSCFT-385

adds an indicator definition page at .../indicator/{indicatorId}

most fields are placeholder

validated ones we don't expect to be placeholder are tested for

back / 404 currently redirect to the home page. There's no user flow implemented that gets the user here at the mo so dealing with these is future work

image
image
image
image


frontend/fingertips-frontend/components/pages/indicator/index.tsx
```diff
@@ -0,0 +1,270 @@
...
+      <H4>Notes</H4>
+      <Paragraph>{indicatorDefinitionProps.notes}</Paragraph>
+      <H4>Caveats</H4>
+      <Paragraph>{indicatorDefinitionProps.caveats}</Paragraph>
+      <H4>Impact of Covid-19</H4>
+      <Paragraph>{indicatorDefinitionProps.covidImpact}</Paragraph>
+
+      <H3 id="links">Public health profile usage and related content</H3>
+      <H4>This indicator is used in the following health profiles</H4>
+      <UnorderedList listStyleType="none">
```
MonkeyDate, line 248: do we have designs for the scenario where this list is empty?

frontend/fingertips-frontend/components/pages/indicator/index.tsx
```diff
@@ -0,0 +1,270 @@
...
+            <ListItem key={`profiles-${idx}`}>
+              <Link href="#">{profile}</Link>
+            </ListItem>
+          );
+        })}
+      </UnorderedList>
+
+      <H4>Related Content</H4>
+      <UnorderedList listStyleType="none">
+        {indicatorDefinitionProps.links.map((link, idx) => {
```
MonkeyDate, line 260: same here, what if the list is empty

frontend/fingertips-frontend/components/pages/indicator/index.tsx
Joe-Hendley: No designs yet. It falls under the placeholder part of the data.

As it stands adding keying to the list to make react/sonarqube happy would imply things we don't really know about the underlying data in a way that isn't quite the same for the rest of the fields.

So, binned it and replaced it with hardcoded placeholders for now.

frontend/fingertips-frontend/components/pages/indicator/index.tsx
Joe-Hendley: https://github.com/dhsc-govuk/FingertipsNext/pull/231#discussion_r1987298871

PR #232, opened 2025-03-07
Feature/dhscft 240 show value types 1

obarojohnson:
# Description

Jira ticket: [DHSCFT-240](https://bjss-enterprise.atlassian.net/browse/DHSCFT-240)

This ticket will primarily focus on the line charts and table then we can apply any logic going forward for further visualisations. The indicator data uses a particular unit of measurement e.g. 

## Changes
image

## LineChart
image

##Bar chart
image

- Add a new variable to the data object called "measurementUnit"
- Refactor the Linechart to make it easy to read. I believe this will have to change in the future.
- Linechart Y-Label added and also units on the values when hover has be added also.
- Barchart has now have a value unit to them.
- Snapshort now added


frontend/fingertips-frontend/components/pages/chart/index.tsx
```diff
@@ -68,6 +68,7 @@ export function Chart({
         yAxisTitle="Value"
         benchmarkLabel="England"
         benchmarkValue={800}
+        measurementUnit=""
```
gareth-allan-bjss, line 71: I think we'd be better making `measurementUnit` an optional property, rather than having to pass it an empty string when we don't want it displayed. It makes it clearer that it's valid for no measurement unit to be supplied.

frontend/fingertips-frontend/components/viewPlots/OneIndicatorOneAreaViewPlots/index.tsx
```diff
@@ -80,6 +86,7 @@ export function OneIndicatorOneAreaViewPlots({
                     healthIndicatorData={dataWithoutEngland}
                     englandBenchmarkData={englandBenchmarkData}
                     groupIndicatorData={groupData}
+                    measurementUnit={indicatorMetadata?.unitLabel ?? ''}
```
gareth-allan-bjss, line 89: Again, I think `measurementUnit` should just be an optional property of `LineChartTable`.

frontend/fingertips-frontend/components/viewPlots/OneIndicatorOneAreaViewPlots/index.tsx
```diff
@@ -68,6 +68,12 @@ export function OneIndicatorOneAreaViewPlots({
                     searchState={searchState}
                     groupIndicatorData={groupData}
                     xAxisTitle="Year"
+                    yAxisTitle={
+                      indicatorMetadata?.unitLabel
+                        ? `Value: ${indicatorMetadata?.unitLabel}`
+                        : ''
+                    }
+                    measurementUnit={indicatorMetadata?.unitLabel ?? ''}
```
gareth-allan-bjss, line 76: As with the others, I think `measurementUnit` and in this case `yAxisTitle` should be optional properties:
```suggestion
                    yAxisTitle={
                      indicatorMetadata?.unitLabel
                        ? `Value: ${indicatorMetadata?.unitLabel}`
                        : undefined
                    }
                    measurementUnit=indicatorMetadata?.unitLabel
```

frontend/fingertips-frontend/components/organisms/LineChart/lineChart.test.tsx
```diff
@@ -28,13 +28,16 @@ const state: SearchStateParams = {
...
 it('should render the Highcharts react component with passed parameters within the LineChart component', async () => {
   const xAxisPropsTitle = 'DifferentXTitle';
+  const yAxisPropsTitle = 'DifferentYTitle';
 
   render(
     <LineChart
       healthIndicatorData={mockHealthData[1]}
-      xAxisTitle={`${xAxisPropsTitle}`}
+      xAxisTitle={xAxisPropsTitle}
+      yAxisTitle={yAxisPropsTitle}
```
gareth-allan-bjss, line 37: We have an assertion in this test that the x-axis title is present. We should do the same for the y-axis title.

gareth-allan-bjss: It might be nice to see if we can verify the measurement unit is present too, but probably not something to sink time into if it's not straightforward.

frontend/fingertips-frontend/components/organisms/LineChartTable/index.tsx
```diff
@@ -137,7 +138,10 @@ const getCellHeader = (
       data-testid={`header-${heading}-${index}`}
       key={`header-${heading}`}
     >
-      {heading} <StyledSpan>(%)</StyledSpan>
+      {heading}
+      <StyledSpan>
+        {heading === LineChartTableHeadingEnum.AreaValue ? `(${units})` : null}
```
gareth-allan-bjss, line 143: We already have logic in this function that decides whether to show the units or not. I think we could just update the `return heading === LineChartTableHeadingEnum.AreaCount ? (` line to be `return heading !== LineChartTableHeadingEnum.AreaValue ? (` and you wouldn't need the ternary statement here.

frontend/fingertips-frontend/components/organisms/LineChartTable/index.tsx
```diff
@@ -236,14 +244,16 @@ export function LineChartTable({
...
+            {/* The header rendering is here */}
             <Table.Row>
               <StyledAlignLeftHeader
                 data-testid={`header-${LineChartTableHeadingEnum.AreaPeriod}-${0}`}
                 key={`header-${LineChartTableHeadingEnum.AreaPeriod}`}
               >
                 {LineChartTableHeadingEnum.AreaPeriod}
               </StyledAlignLeftHeader>
-              {healthIndicatorData.map(() =>
+              {healthIndicatorData.map((_) =>
```
gareth-allan-bjss, line 256: I'm not sure of the value of adding this?

frontend/fingertips-frontend/components/organisms/LineChart/lineChart.test.tsx
obarojohnson: I can try and add a check if the value is presented , since we are going to have more than one instance of this. We can use check if the value is render atleast on the UI.

frontend/fingertips-frontend/components/viewPlots/OneIndicatorOneAreaViewPlots/index.tsx
gareth-allan-bjss: I think this one is still outstanding.

frontend/fingertips-frontend/components/organisms/LineChartTable/__snapshots__/LineChartTable.test.tsx.snap
gareth-allan-bjss, line 1: Why delete this?

gareth-allan-bjss: I think this is why the unit tests are failing.

gareth-allan-bjss (approved):

PR #237, opened 2025-03-10
Feature/dhscft 449 add value units to inequalities table

obarojohnson:
# Description

Jira ticket: [DHSCFT-449](https://bjss-enterprise.atlassian.net/browse/DHSCFT-449)

This ticket is add the unit in the table for Sex inequalities for a single time period

image

image

## Changes

-  New measurement unit variable add to the inequalities component
-  Unit test to take the snapshot since it will be more difficult to trying to check for the measurement unit.


frontend/fingertips-frontend/components/organisms/Inequalities/inequalities.test.tsx
```diff
@@ -21,4 +21,24 @@ describe('Inequalities suite', () => {
...
+    render(
+      <Inequalities
+        healthIndicatorData={MOCK_HEALTH_DATA[1]}
+        measurementUnit="kg"
+      />
+    );
+    expect(screen.getByText('(kg)')).toBeInTheDocument();
+  });
+
+  it('take a snapshot', () => {
```
adawari-thomas-bjss, line 35: I think this snapshot test might cause issues when the barchart and linechart are added to this component, so I don't think we need it here

obarojohnson: Yes you can update it , if new changes is added I think so. That is fine right to do so I think, but assuming you have faced some issues before, can you explain what issues thanks?

adawari-thomas-bjss: Highcharts dynamically renders ids in its chart components. This might affect the line chart and bar chart when they're added to this component, and the snapshot test might keep failing even when we haven't made any changes. This was an issue at the start of the project, so snapshot testing was removed from components containing charts. See https://github.com/highcharts/highcharts-react/issues/157

frontend/fingertips-frontend/components/organisms/Inequalities/index.tsx
```diff
@@ -1,3 +1,5 @@
+'use client';
+
```
gareth-allan-bjss, line 2: Do we need to explicitly make this a client component?

frontend/fingertips-frontend/components/organisms/Inequalities/index.tsx
gareth-allan-bjss, line 1: You're not actually passing a value for your new `measurementUnit` prop into this component. So when running as part of the application no unit values are being shown. I don't think we can call this ticket complete if you're not actually showing unit values in these tables.

frontend/fingertips-frontend/components/molecules/Inequalities/BarChart/Table/index.tsx
```diff
@@ -44,13 +48,23 @@ const getCellHeader = (header: InequalitiesBarChartTableHeaders): ReactNode =>
       style={{ width: '16%' }}
       data-testid={`heading-${header}`}
     >
-      {header}
+      {header === InequalitiesBarChartTableHeaders.VALUE && measurementUnit ? (
+        <>
+          {header}{' '}
+          <span
+            style={{ display: 'block', margin: 'auto' }}
+          >{`(${measurementUnit})`}</span>
```
gareth-allan-bjss, line 56: Is this rendering the measurement unit wrapped in curly braces? E.g. `{kg}`? I don't believe that's the desired display for the inequalities bar chart table.

frontend/fingertips-frontend/components/molecules/Inequalities/BarChart/Table/inequalitiesBarChartTable.test.tsx
```diff
@@ -100,6 +100,18 @@ describe('Inequalities bar chart table suite', () => {
...
+          tableData={tableData}
+          yearlyHealthDataGroupedByInequalities={GROUPED_YEAR_DATA}
+          type={Inequalities.Sex}
+          measurementUnit="kg"
+        />
+      );
+      expect(screen.getByText('(kg)')).toBeInTheDocument();
+    });
+
     it('snapshot test - should match snapshot', () => {
```
gareth-allan-bjss, line 115: For the inequalities line chart table you updated the snapshot test to pass the measurement unit. I think we should do the same here.

frontend/fingertips-frontend/components/molecules/Inequalities/LineChart/Table/inequalitiesLineChartTable.test.tsx
```diff
@@ -107,12 +107,35 @@ describe('Inequalities table suite', () => {
...
+    it("Value: 'unit measurement' should not be in document when the unit measurement is not provided", () => {
+      render(
+        <InequalitiesLineChartTable
+          tableData={tableData}
+          yearlyHealthDataGroupedByInequalities={GROUPED_YEAR_DATA}
+          type={Inequalities.Sex}
+        />
+      );
+      expect(screen.queryByText('Value: per 100,000')).not.toBeInTheDocument();
+    });
```
gareth-allan-bjss, line 131: This might be better done by giving the measurement unit span a `data-testid` property and then asserting that no element with that property is in the document. As it is, this test is really just validating that some text isn't present.

frontend/fingertips-frontend/components/molecules/Inequalities/BarChart/Table/inequalitiesBarChartTable.test.tsx
```diff
@@ -100,6 +100,18 @@ describe('Inequalities bar chart table suite', () => {
         .forEach((id) => expect(id).toHaveTextContent('X'));
     });
 
+    it('check if the measurementUnit value "kg" is rendered correctly with braces', () => {
```
gareth-allan-bjss, line 103: You had a test for the inequalities line chart table that verified the measurement unit wasn't shown when it wasn't passed. I think the same test would be good here.

frontend/fingertips-frontend/components/organisms/Inequalities/inequalities.test.tsx
gareth-allan-bjss: Ah, I missed that, good thinking @adawari-thomas-bjss!

frontend/fingertips-frontend/components/molecules/Inequalities/BarChart/Table/index.tsx
obarojohnson: no

obarojohnson: linechart

frontend/fingertips-frontend/components/pages/chart/index.tsx
```diff
@@ -33,20 +36,39 @@ export function Chart({
...
+          await SearchServiceFactory.getIndicatorSearchService().getIndicator(
+            indicator
+          );
+        setIndicatorMetaData(document);
+      } catch (e) {
+        console.error(e);
+      }
+    };
+    fetchIndicatorMeta(
+      indicatorsSelected?.length > 0 ? indicatorsSelected[0] : ''
```
gareth-allan-bjss, line 68: I don't think we need to worry about whether more than one indicator has been selected. We know we're only going to be using this data for inequalities (which is displayed only when 1 indicator/1 area is selected). Plus, we plan on refactoring this soon. I think we should just grab the code from `frontend/fingertips-frontend/components/views/OneIndicatorOneAreaView/index.tsx` that makes this query and re-use that (including the try-catch, as that lets us show the page even if we can't get the unit labels to display).

frontend/fingertips-frontend/components/pages/chart/index.tsx
```diff
@@ -69,6 +92,7 @@ export function Chart({
         benchmarkLabel="England"
         benchmarkValue={800}
         accessibilityLabel="A bar chart showing healthcare data"
+        measurementUnit={indicatorMetadata?.unitLabel}
```
gareth-allan-bjss, line 95: Not something you need to change now, but for the record this bar chart was an early version of the inequalities view and should be removed as part of Ada's work, so I wouldn't be worried if it didn't have unit labels. It doesn't hurt to add them though!

frontend/fingertips-frontend/components/organisms/Inequalities/index.tsx
gareth-allan-bjss: Discussed offline, but yes I'm saying we should query the indicator metadata in the chart page and pass it down to this component.

frontend/fingertips-frontend/components/molecules/Inequalities/BarChart/Table/inequalitiesBarChartTable.test.tsx
```diff
@@ -112,12 +112,26 @@ describe('Inequalities bar chart table suite', () => {
       expect(screen.getByText('(kg)')).toBeInTheDocument();
     });
 
+    it('check if that measurementUnit value does shown when its not passed', () => {
```
gareth-allan-bjss, line 115:
```suggestion
    it('check if that measurementUnit value in not shown when its not passed', () => {
```

frontend/fingertips-frontend/components/pages/chart/index.tsx
```diff
@@ -33,20 +36,39 @@ export function Chart({
...
+        setIndicatorMetaData(document);
+      } catch (e) {
+        console.error(e);
+      }
+    };
+    fetchIndicatorMeta(
+      indicatorsSelected?.length > 0 ? indicatorsSelected[0] : ''
+    );
+  }, [indicatorsSelected]);
+
```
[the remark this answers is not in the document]
obarojohnson: Got it! That’s what I was talking about—moving the query to the top. Doing this in this ticket means I’m already refactoring the code, which will then involve refactoring the chart component as discussed. Should [DHSCFT-465](https://bjss-enterprise.atlassian.net/browse/DHSCFT-465) be used for that task?

frontend/fingertips-frontend/components/molecules/Inequalities/BarChart/Table/index.tsx
obarojohnson: the image of it is shown above, its at the top of the page. Yea my bad provided the wrong one :)

frontend/fingertips-frontend/components/molecules/Inequalities/BarChart/Table/inequalitiesBarChartTable.test.tsx
gareth-allan-bjss: Ah, I made a typo in my suggestion. Can this say `is not` instead of `in not` please?

frontend/fingertips-frontend/components/pages/chart/index.tsx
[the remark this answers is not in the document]
gareth-allan-bjss: I think there's more to DHSCFT-465 than just moving the component, as you'll have to update the query made in the view to include inequalities data and implement some additional filtering. I think there's also the risk of causing conflicts because Ada is still working on some inequalities components.

I think it still makes sense to just look up the indicator metadata in `frontend/fingertips-frontend/app/chart/page.tsx` and pass it into the `Chart` component for now.

frontend/fingertips-frontend/components/pages/chart/index.tsx
gareth-allan-bjss: I think you could just have a check to see if `indicatorsSelected` is undefined to avoid that.

frontend/fingertips-frontend/components/pages/chart/index.tsx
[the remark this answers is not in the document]
gareth-allan-bjss: This looks good now.

frontend/fingertips-frontend/components/molecules/Inequalities/BarChart/Table/index.tsx
gareth-allan-bjss: I think this is also true for the (non-inequalities) line chart table as well, so it might be good to update that as part of this PR as well.

gareth-allan-bjss (approved):

PR #247, opened 2025-03-11
DHSCFT-374: Extended console app to process trend data

dlzhry2:
# Description

Jira ticket: [DHSCFT-374](https://bjss-enterprise.atlassian.net/browse/DHSCFT-374)

This now extends the TrendAnalysisApp console application so that it will set the trend value for each of the PoC Indicators (for each of their unique dimensions).

This is runnable locally using the instructions in the README. The follow on ticket will look to create an optional pipeline job to update the values in the actual deployed DB in Azure.

## Changes

- Updated the names of the trends that are added to the database. Updated API and frontend API client and tests to reflect those changes
- Added polarity logic to the TrendAnalysisApp, so that the value that is set will be considering polarity. This minimises any work needed to be done either in the frontend or API code.
- Etc.

## Validation

If you had to update the chart component screenshots then please state that here and also link to the `Artifact download URL:` from the `Upload playwright artefacts` step of the `e2e-tests-local-docker` failure job.


trend-analysis/TrendAnalysisApp/Calculator/Trend.cs
```diff
@@ -0,0 +1,13 @@
+namespace TrendAnalysisApp.Calculator;
+
+public enum Trend {
```
dlzhry2, line 3: In my initial implementation I had a repository to retrieve the trends data from the DB right at the start and then each time calculate a trend, look up the key by the name. However, I thought this might be better because.
A. Simpler, more efficient as it does not need to make the call. Plus all the trend data loaded into the DB is within our control so I don't believe it's an issue that my app contains references to the keys/order of that data.
B. Looked like I'd have to use consts for the Trend names from the DB as they have spaces in them, so thought an enum might be nicer for the application code.

dlzhry2: Let me know if agree or any thoughts on an alternative approach.

trend-analysis/TrendAnalysisApp/Constants.cs
```diff
@@ -8,4 +8,14 @@ public static class Database {
         // The name for the Fingertips DB: used in the environment variable for the connection string.
         public const string FingertipsDbName = "FINGERTIPS_DB";
```
MonkeyDate, line 9: Shame we don't use the same name across the entire FingertipsNext solution.

trend-analysis/TrendAnalysisApp/Mapper/LegacyMapper.cs
```diff
@@ -60,7 +60,7 @@ private static CoreDataSet HealthMeasureToLegacyDataset(HealthMeasureModel healt
         {
             Count = healthMeasure.Count,
             Value = healthMeasure.Value,
-            Denominator = healthMeasure.Denominator,
+            Denominator = healthMeasure.Denominator ?? 0,
```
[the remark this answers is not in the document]
MonkeyDate: Perhaps we should add a comment for future reference

trend-analysis/TrendAnalysisApp/Calculator/TrendCalculator.cs
```diff
@@ -1,13 +1,15 @@
...
 using TrendAnalysisApp.Repository.Models;
+using static TrendAnalysisApp.Constants;
 
 namespace TrendAnalysisApp.Calculator;
 
 public class TrendCalculator(TrendMarkerCalculator legacyCalculator, LegacyMapper legacyMapper) : ITrendCalculator
 {
     private readonly TrendMarkerCalculator legacyCalculator = legacyCalculator;
     private readonly LegacyMapper legacyMapper = legacyMapper;
+    public readonly int RequiredNumberOfDataPoints = 5;
```
MonkeyDate, line 12: 'const' ?

api/definition/swagger.yaml
```diff
@@ -407,8 +407,17 @@ components:
           description: The statistical trend that applies to the data point,
             given the preceding data. Will only be calculated if there are at
             least 5 data points to use.
-          enum: [Increasing, Decreasing, NoChange, CannotBeCalculated,
-            NotYetCalculated]
+          enum: [
```
steves-bjss, line 410: In the scenario we don't have 5 data points, are we leaving the status as 'Not yet calculated', or do we need a new 'Not enough data' status?

trend-analysis/TrendAnalysisApp/Constants.cs
steves-bjss: > Shame we don't use the same name across the entire FingertipsNext solution.

Its a fair comment but for various reasons, we have chosen to keep this as a standalone project, separate to the API.

trend-analysis/TrendAnalysisApp/Program.cs
```diff
@@ -24,7 +27,11 @@ private static ServiceProvider CreateServices()
             .AddSingleton<IConfiguration>(configuration)
             .AddDbContext<HealthMeasureDbContext>(options => options.UseSqlServer(connString))
             .AddSingleton<TrendDataProcessor>()
+            .AddSingleton<IndicatorRepository>()
             .AddSingleton<HealthMeasureRepository>()
+            .AddSingleton<TrendCalculator>()
```
steves-bjss, line 32: Are these singletons all thread safe if we were to start processing in parallel?

trend-analysis/TrendAnalysisApp/Calculator/TrendCalculator.cs
dlzhry2: I've updated now.

MonkeyDate (approved):

trend-analysis/TrendAnalysisApp/Mapper/LegacyMapper.cs
```diff
@@ -60,7 +60,8 @@ private static CoreDataSet HealthMeasureToLegacyDataset(HealthMeasureModel healt
         {
```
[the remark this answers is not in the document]
dlzhry2: Thanks, very keen-eyed indeed :)

trend-analysis/TrendAnalysisApp/Calculator/Trend.cs
steves-bjss: I'm happy with this approach to have an enum that matches the database.

api/definition/swagger.yaml
dlzhry2: Discussed with Emma. In terms of the error scenarios for the PoC the requirement is a catch all string to simply state could not be calculated.

Could definitely extend in future - either with our own implementation or retrieving the messages from the legacy calculator - as there's a few different reasons that can cause that result.

trend-analysis/TrendAnalysisApp/Program.cs
dlzhry2: That's a good question. I think everything would be safe except for the HealthMeasureRepository. It could be used safely but would need a warning for proper usage.

The bit that is unsafe would be to do with A. updates are saved to the DB context and B. on calling save, the ChangeTracker is cleared. So with improper usage you could have a situation where one thread removes the queued changes for another.

 I had developed it with purely sync code in mind i.e. await the changes being saved for an indicator and then move onto the next, given the kick off we had. I can look into refactoring it, if we're wanting to move to parallel e.g. create an instance of the TrendCalculator per thread. Although I'm quite new to .NET so might need a pointer. Or is the current approach good enough for PoC/we create a backlog ticket to parallelize the indicator updates?

dlzhry2: Thinking about it a little more: you could parallelize the CPU-bound calculation and store the changes. The save would have to be called at the end after all the updates have been added to the DB context though.

PR #270, opened 2025-03-13
DHSCFT-468: cache api request on FE

alibombaye:
# Description

Jira ticket: [DHSCFT-468](https://bjss-enterprise.atlassian.net/browse/DHSCFT-468)

Adds caching config to all api requests for 1 hour before revalidating

## Validation

Tested locally - first requests to an endpoint will fetch from the API. Subsequent requests to the same endpoint will fetch from cache

### DIFF
--- a/frontend/fingertips-frontend/lib/apiClient/apiClientFactory.ts
+++ b/frontend/fingertips-frontend/lib/apiClient/apiClientFactory.ts
@@ -16,6 +16,7 @@ export class ApiClientFactory {
       const apiUrl = readEnvVar('FINGERTIPS_API_URL');
       const config: Configuration = new Configuration({
         basePath: apiUrl,
+        fetchApi: fetch,
       });
 
       this.areasApiInstance = new AreasApi(config);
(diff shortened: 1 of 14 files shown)


frontend/fingertips-frontend/lib/apiClient/apiClientFactory.ts
```diff
@@ -6,6 +6,8 @@ import {
 } from '@/generated-sources/ft-api-client';
 import { readEnvVar } from '../envUtils';
 
+export const API_CACHE_CONFIG = { next: { revalidate: 3600 } };
```
williamwillis11, line 9: worth adding a test to frontend/fingertips-frontend/lib/apiClient/apiClientFactory.test.ts to prove this config is having the right impact?

alibombaye: Can't really unit test the caching. The tests where we make api calls have been updated to check that this is being passed across. I think that's as much as we can do here.

gareth-allan-bjss (approved):

PR #276, opened 2025-03-14
DHSCFT-434: Make the API handle indicators with a single inequality value correctly

gareth-allan-bjss:
# Description

Jira ticket: [DHSCFT-434](https://bjss-enterprise.atlassian.net/browse/DHSCFT-434)

This PR updates the API so that it will treat an indicator which has only data with a single value for an inequality dimension (i.e. sex, age or deprivation) as if it does not have data broken down by that inequality (since it doesn't!).

## Changes

- Updated the API's `HealthDataRepository` to use a separate query count the number of inequality values for the indicator being requested.
- Updated `HealthDataRepository` to use the count on inequality values to treat an indicator with a single value for an inequality dimension as if it didn't have a value
- Updated API configuration/code to resolve some build warnings
- Various dependency updates in the API

## Validation

Captured the SQL generated for a health data request with no inequalities parameters for an indicator with data broken down by age and sex, but not deprivation.

Request 1:
```sql
SELECT COUNT(DISTINCT ([h].[SexKey])) AS [SexCount], COUNT(DISTINCT ([h].[AgeKey])) AS [AgeCount], COUNT(DISTINCT ([h].[DeprivationKey])) AS [DeprivationCount]
FROM [HealthMeasure] AS [h]
INNER JOIN [IndicatorDimension] AS [i] ON [h].[IndicatorKey] = [i].[IndicatorKey]
WHERE [i].[IndicatorId] = @__indicatorId_0
GROUP BY [h].[IndicatorKey]
```

Request 2:
```sql
FROM [HealthMeasure] AS [h]
INNER JOIN [IndicatorDimension] AS [i] ON [h].[IndicatorKey] = [i].[IndicatorKey]
INNER JOIN [AreaDimension] AS [a] ON [h].[AreaKey] = [a].[AreaKey]
INNER JOIN [SexDimension] AS [s] ON [h].[SexKey] = [s].[SexKey]
INNER JOIN [AgeDimension] AS [a0] ON [h].[AgeKey] = [a0].[AgeKey]
INNER JOIN [TrendDimension] AS [t] ON [h].[TrendKey] = [t].[TrendKey]
INNER JOIN [DeprivationDimension] AS [d] ON [h].[DeprivationKey] = [d].[DeprivationKey]
WHERE [i].[IndicatorId] = @__indicatorId_0 AND [a].[Code] IN (N'E06000008', N'E92000001') AND [s].[HasValue] = CAST(0 AS bit) AND [a0].[HasValue] = CAST(0 AS bit)
ORDER BY [h].[Year]
```

Verified that data was being returned as expected.

For an indicator with `Persons` sex data, `All ages` age data and aggregated deprivation data (41101):
Screenshot 2025-03-14 at 16 00 09

For an indicator with a single value for the age dimension, `Persons` sex data and aggregated deprivation data (383):
Screenshot 2025-03-14 at 16 11 17
(Note: the duplication of values for the same year appears to be present in the DB itself and is being investigated separately.)

For an indicator with a single value for the sex dimension, `All ages` age data and aggregated deprivation data (93085):
Screenshot 2025-03-14 at 16 12 35

For an indicator with a single value for the sex dimension, a single value for the age dimension and aggregated deprivation data (92266):
Screenshot 2025-03-14 at 16 13 59

Checked the performance of the health data query with these changes (for indicator 337, which has in the region of 400,000 rows of data): 
Request without inequalities parameters:
First query: `Executed DbCommand (1,194ms)`
Second query: `Executed DbCommand (34ms)`

Request with age and sex inequality parameters:
First query: `Executed DbCommand (69ms)`
Second query: `Executed DbCommand (30ms)`

Confirmed that with these changes I am able to view the chart page for an indicator which currently works on main (41101) and the data is unchanged.
With these changes:
Screenshot 2025-03-14 at 16 35 28
Screenshot 2025-03-14 at 16 36 03

On main:
Screenshot 2025-03-14 at 16 27 29
Screenshot 2025-03-14 at 16 27 51

Also that I can view data for an indicator which does not currently work on main (383).
With changes:
Screenshot 2025-03-14 at 16 43 11

Main:
Screenshot 2025-03-14 at 16 43 33


api/DHSC.FingertipsNext.Modules.HealthData.Repository/HealthDataRepository.cs
```diff
@@ -1,22 +1,38 @@
...
 public class HealthDataRepository(HealthDataDbContext healthDataDbContext) : IHealthDataRepository
 {
     private readonly HealthDataDbContext _dbContext = healthDataDbContext ?? throw new ArgumentNullException(nameof(healthDataDbContext));
 
     public async Task<IEnumerable<HealthMeasureModel>> GetIndicatorDataAsync(int indicatorId, string[] areaCodes, int[] years, string[] inequalities)
     {
+        var countInequalityDimensionsQuery = _dbContext.HealthMeasure
+            .Where(hm => hm.IndicatorDimension.IndicatorId == indicatorId)
+            .Include(hm => hm.IndicatorDimension)
+            .GroupBy(x => x.IndicatorKey)
```
HowardMayBjss, line 17: Isn't this redundant because we only have data for one indicatorId?

gareth-allan-bjss: The group by clause? Yes, I'll take a look at removing it (unfortunately it doesn't seem as simple as just deleting it).

frontend/fingertips-frontend/components/organisms/LineChart/index.tsx
```diff
@@ -118,7 +118,7 @@ export function LineChart({
     tooltip: {
       format:
         '<b>{point.series.name}</b><br/>Year: {point.x}<br/><br/><span style="color:{color}">\u25CF</span> Value {point.y}' +
-        `${measurementUnit}`,
+        ` ${measurementUnit}`,
```
gareth-allan-bjss, line 121: This is to add a space between the value and measurement unit in the line chart's tooltip. At the moment, for an indicator with a unit that isn't `%` it displays without a space (e.g. `24per 100,000`). This aligns with how the inequalities line chart will display too.

steves-bjss (approved):

api/DHSC.FingertipsNext.Modules.HealthData.Repository/HealthDataRepository.cs
gareth-allan-bjss: I took a look at this with @steves-bjss and we couldn't find a nice way of removing the group by clause and keeping the select section the same. The only other version of this query we came up with was messier and ended up running multiple queries and doing the filtering client-side.

I'm going to leave this here for now.

api/DHSC.FingertipsNext.Modules.HealthData.Repository/HealthDataRepository.cs
```diff
@@ -1,22 +1,38 @@
...
 namespace DHSC.FingertipsNext.Modules.HealthData.Repository;
 
+[SuppressMessage("ReSharper", "SimplifyConditionalTernaryExpression")]
 public class HealthDataRepository(HealthDataDbContext healthDataDbContext) : IHealthDataRepository
 {
     private readonly HealthDataDbContext _dbContext = healthDataDbContext ?? throw new ArgumentNullException(nameof(healthDataDbContext));
 
     public async Task<IEnumerable<HealthMeasureModel>> GetIndicatorDataAsync(int indicatorId, string[] areaCodes, int[] years, string[] inequalities)
     {
+        var countInequalityDimensionsQuery = _dbContext.HealthMeasure
```
HowardMayBjss, line 14: I presume because this doesn't have an await it is executed when something tries to access the results?

HowardMayBjss, line 14: Does this result in two trips to the database or is it clever enough to do it in one?

api/DHSC.FingertipsNext.Modules.HealthData.Repository/HealthDataRepository.cs
```diff
@@ -1,22 +1,38 @@
...
+            .Where(hm => hm.IndicatorDimension.IndicatorId == indicatorId)
+            .Include(hm => hm.IndicatorDimension)
+            .GroupBy(x => x.IndicatorKey)
+            .Select(x => new { 
+                SexKeyCount = x.Select(hm => hm.SexKey).Distinct().Count(),
```
HowardMayBjss, line 19: Would it make the intent more clear to refactor this as a var/attribute indicatorMetaData.hasInherentSexInequality : Boolean

api/DHSC.FingertipsNext.Modules.HealthData.Repository/HealthDataRepository.cs
gareth-allan-bjss: Yes, my understanding is that's what makes this run as a nested query (see the SQL I've now included in the PR description).

gareth-allan-bjss: It's clever enough to do it in one SQL query (again, I've now included the generated SQL in the PR description).

api/DHSC.FingertipsNext.Modules.HealthData.Repository/HealthDataRepository.cs
gareth-allan-bjss: Do you mean replace `SexKeyCount` here (and the other counts) with something like this:
`SexKeyInherent = x.Select(hm => hm.SexKey).Distinct().Count() == 1`

Or am I misunderstanding?

api/DHSC.FingertipsNext.Modules.HealthData.Repository/HealthDataRepository.cs
```diff
@@ -1,22 +1,86 @@
...
+    }
+
+    private static bool ShouldIncludeSexDimension(string[] inequalities, List<InequalitiesCount> inequalityDimensionsCount)
+    {
+        return InequalitiesParamsContainSex(inequalities) || inequalityDimensionsCount[0].SexCount == 1;
+    }
+
+    private static bool InequalitiesParamsContainAge(string[] inequalities)
+    {
+        return inequalities.Contains("age");
```
HowardMayBjss, line 31: I think it would be better to factor these functions out.

api/DHSC.FingertipsNext.Modules.HealthData.Repository/HealthDataRepository.cs
```diff
@@ -1,22 +1,86 @@
...
+        public int? AgeCount { get; init; }
+        public int? DeprivationCount { get; init; }
+    }
+
+    private static bool InequalitiesParamsContainSex(string[] inequalities)
+    {
+        return inequalities.Contains("sex");
+    }
+
+    private static bool ShouldIncludeSexDimension(string[] inequalities, List<InequalitiesCount> inequalityDimensionsCount)
```
HowardMayBjss, line 24: I think it would be better to renmae inequalities to inequalitiesParams and replace InequalitiesParamsContainsSex with inequalitiesParams.Contains("sex");

api/DHSC.FingertipsNext.Modules.HealthData.Repository/HealthDataRepository.cs
```diff
@@ -1,22 +1,86 @@
...
+    {
+        return inequalities.Contains("deprivation");
+    }
+
+    private static bool ShouldIncludeDeprivationDimension(List<InequalitiesCount> inequalityDimensionsCount)
+    {
+        return inequalityDimensionsCount[0].DeprivationCount == 1;
+    }
+
+    private async Task<List<InequalitiesCount>> CountInequalityDimensions(int indicatorId, string[] inequalities)
```
HowardMayBjss, line 49: This will need extending when new inequalities are added - worth a comment somewhere

api/DHSC.FingertipsNext.Modules.HealthData.Repository/HealthDataRepository.cs
```diff
@@ -1,22 +1,86 @@
...
             .Where(hm => years.Length == 0 || years.Contains(hm.Year))
-            .Where(hm => inequalities.Contains("sex") ? true : hm.SexDimension.HasValue == false)
-            .Where(hm => inequalities.Contains("age") ? true : hm.AgeDimension.HasValue == false)
+            .Where(hm => ShouldIncludeSexDimension(inequalities, inequalityDimensionsCount)
+                ? true
+                : hm.SexDimension.HasValue == false)
+            .Where(hm => ShouldIncludeAgeDimension(inequalities, inequalityDimensionsCount)
+                ? true
+                : hm.AgeDimension.HasValue == false)
             // TODO: Will be expanded to allow the deprivation dimension to be retrieved based on a query param in DHSCFT-396
```
HowardMayBjss, line 82: I think this comment needs to move to the ShouldIncludeDeprivationDimension function above

HowardMayBjss:
This data point is dupolicated - this doesn't look right?
image

steves-bjss (approved):

api/DHSC.FingertipsNext.Modules.HealthData.Repository/HealthDataRepository.cs
HowardMayBjss, line 49: This is counting the number of distinct values each dimension has. I think naming the function CountInequalityDimensionsDistinctValues or CountInequalityDimensionsDistinctKeys is more correct

gareth-allan-bjss:
> This data point is dupolicated - this doesn't look right? image

Yes, I've just updated the description. If you use the following query you can see that this is actually due to duplication of data in the DB, and not these changes:
```sql
SELECT TOP (1000) [HealthMeasureKey]
      ,[AgeKey]
      ,a.[Code]
      ,a.Name
      ,i.[IndicatorId]
      ,[SexKey]
      ,[TrendKey]
      ,[DeprivationKey]
      ,[Count]
      ,[Denominator]
      ,[Value]
      ,[LowerCI]
      ,[UpperCI]
      ,[Year]
FROM [fingertips-db].[dbo].[HealthMeasure] hm
JOIN IndicatorDimension i ON i.IndicatorKey = hm.IndicatorKey
JOIN AreaDimension a ON a.AreaKey = hm.AreaKey
WHERE i.IndicatorId = 383
AND a.Code = 'E38000201'
ORDER BY [Year]
```

This results in the following (note the duplicated rows):
Screenshot 2025-03-18 at 13 49 54

As mentioned in the description, this is being investigated separately, so I'll not deal with it in this PR.

HowardMayBjss (approved):

api/DHSC.FingertipsNext.Modules.HealthData.Repository/HealthDataRepository.cs
HowardMayBjss: This is now redundant

PR #285, opened 2025-03-17
DHSCFT-214: Add trend marker to line chart table

dlzhry2:
# Description

Jira ticket: [DHSCFT-214](https://bjss-enterprise.atlassian.net/browse/DHSCFT-214)
I believe this also resolves [DHSCFT-371](https://bjss-enterprise.atlassian.net/browse/DHSCFT-371)

On a line chart table with displaying 1-2 areas for a given indicator, the trend will now be shown for the relevant areas

## Changes

- Slight modification to existing TrendTag component to use the actual data that will be returned by the API
- Removed the demo tags (hope this is ok - they did not seem to be referenced anywhere
- Extended line chart table to render the relevant trend tag and updated all relevant tests
- The Indicator polarity strings in the database had been changed, so needed to update the trend analysis application accordingly

## Validation

Tested locally (you need to wait for the trend-analysis container to finish running locally (about 30 seconds).

Indicator used: Alcohol-related mortality

All of England:
image
And matches the Fingertips site:
image

2 Areas - Blackpool and Bolton (I flagged the rendering issue - see Engineering channel - it's an existing issue with a ticket to look into it.)
image
This also matches Fingertips website.

Indicator with no polarity: Deaths in hospital (%)

East Midlands region:
image


frontend/fingertips-frontend/package-lock.json
williamwillis11, line 1: why did this file change? looks like just formatting but it hard to be sure

compose.yaml
```diff
@@ -19,6 +19,8 @@ services:
     depends_on:
       db-setup:
         condition: service_completed_successfully
+      db-apply-trends:
```
dlzhry2, line 22: This ensures that the API container does not start until the db has been seeded AND trend data has been calculated.

This ensures consistency for e2e tests and, now that we have trend data in the screenshots, tests are only executed once the DB is fully prepared.

Discussed with @williamwillis11 - this will add around 30 seconds to the running time, but I am looking into parallelising the trend-analysis app so it should get faster.

frontend/fingertips-frontend/lib/tableHelpers/mocks.ts
```diff
@@ -151,3 +151,98 @@ export const GROUPED_YEAR_DATA = {
...
+    areaName: 'Manchester - Counties and Unitary Authorities',
+    healthData: [
+      {
+        year: 2019,
+        count: 1400,
+        value: 222.305692,
+        lowerCi: 220.69151,
+        upperCi: 224.32766,
+        ageBand: 'All',
+        sex: 'All',
```
adawari-thomas-bjss, line 167: Could all these be changed to `Persons` instead of `All` as the api no longer returns the `All` value for the sex field
```suggestion
        sex: 'Persons',
```

compose.yaml
dlzhry2: This has been parallelised now. Resolving.

frontend/fingertips-frontend/lib/tableHelpers/mocks.ts
dlzhry2: Thanks - good point. I've updated. Please check you're happy with the changes.

williamwillis11 (approved):

PR #300, opened 2025-03-18
don't add ccb data to the healthdata.csv

andyevans-bjss:
# Description

Jira ticket: [DHSCFT-500](https://bjss-enterprise.atlassian.net/browse/DHSCFT-500)

Raw data has identical rows for CCGs that are now ICBs - so there was 2 identical rows of data for some areas. Spoke to Doris who said to ignore the CCG area types.

## Validation
Run the import and check the extra rows in the HealthMeasure table are no longer there


DataCreator/DataCreator/DataFileManager.cs
```diff
@@ -76,7 +76,11 @@ public static List<HealthMeasureEntity> GetHealthDataForIndicator(int indicatorI
                 var split=Regex.Split(lines[count], "[,]{1}(?=(?:[^\"]*\"[^\"]*\")*(?![^\"]*\"))");
                 if (split.Length != 27)
                     continue; //avoid bad data
+              
+                if (split[6].Equals("CCG", StringComparison.CurrentCultureIgnoreCase)) //CCGs and ICBs share the same area code so do this to avoid doubling up data
```
HowardMayBjss, line 80: Can you a const with a descriptive name to replace the magic 6 here. Also I think that there is value in giving a more full description as to what is going on here. Future devs will thank us.

HowardMayBjss (approved):

PR #301, opened 2025-03-18
DHSCFT-489 - Add e2e test for clearing indicator on search change

Nayira-BJSS:
# Description

Jira ticket: [DHSCFT-489](https://bjss-enterprise.atlassian.net/browse/DHSCFT-489)

Added new e2e test to cater for the changes made as part of DHSCFT-358

## Changes

- Added new test to ensure the URL is updated correctly to remove any selected indicators when the search changes.

## Validation

New test added and passing

### DIFF
--- a/frontend/fingertips-frontend/playwright/tests/isolated_ui_tests/navigation_and_validation.spec.ts
+++ b/frontend/fingertips-frontend/playwright/tests/isolated_ui_tests/navigation_and_validation.spec.ts
@@ -220,10 +220,10 @@ test.describe(`Navigation, accessibility and validation tests`, () => {
     });
 
     await test.step('Clear search bar and search, verify url is updated to remove indicator', async () => {
-      await resultsPage.clearIndicatorSearchBox()
-      await resultsPage.clickIndicatorSearchButton()
-      await test.expect(resultsPage.page).not.toHaveURL(/&is=/)
-    })
+      await resultsPage.clearIndicatorSearchBox();
+      await resultsPage.clickIndicatorSearchButton();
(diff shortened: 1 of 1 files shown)


frontend/fingertips-frontend/playwright/tests/isolated_ui_tests/navigation_and_validation.spec.ts
```diff
@@ -198,6 +198,33 @@ test.describe(`Navigation, accessibility and validation tests`, () => {
...
+      ]);
+    });
+
+    await test.step('Select single indicator, and verify url is updated to include indicator', async () => {
+      await resultsPage.selectIndicatorCheckboxes(
+        filteredIndicatorIds,
+        indicatorMode
+      );
+
+      await test.expect(resultsPage.page).toHaveURL(/&is=/);
```
williamwillis11, line 219: can we be more specific about which indicator id is added to the url? we should have it in selectIndicatorCheckboxes

williamwillis11: in fact the last thing that act method does is an asset on

await this.waitForURLToContain(indicatorID);

which should not really be in there as its mixing act and assert in the same function. my bad

williamwillis11: but im thinking we either get selectIndicatorCheckboxes to return out the indicator selector id and so the assertion outside that function, where its visible in this test step

or we we remove this line as its essentially dupe

PR #318, opened 2025-03-20
DHSCFT-396: Update deprivation data in API Response

gareth-allan-bjss:
# Description

Jira ticket: [DHSCFT-396](https://bjss-enterprise.atlassian.net/browse/DHSCFT-396)

This PR updates the API to return an object for the `deprivation` field instead of a single string. This object contains the deprivation value currently returned, as well as the deprivation type (which will be used to categorise deprivation values in the frontend) and its sequence value (which will be used to sort the deprivation data when displayed).

## Changes

- Updated the API's Swagger definition to represent a health data point's `deprivation` field as an object
- Added a `Deprivation` schema model, to represent the new deprivation object in the API response
- Expanded the `HealthDataRepository` unit tests to validate that the `deprivation` inequalities parameter works as expected
- Added an example query for the deprivation inequality to the `get-indicator-data.http` file
- Regenerated the frontend's API client to match the updated Swagger spec
- Updated frontend test data to include an example deprivation object

## Validation

Example of a query for an indicator without deprivation data:

Example of a query for an indicator with deprivation data, no inequalities param passed:
Screenshot 2025-03-24 at 17 01 48

Example of a query for an indicator with deprivation data, `deprivation` inequalities param passed:
Screenshot 2025-03-24 at 17 17 00

Example of a query for an indicator without deprivation data, `deprivation` inequalities param passed:
Screenshot 2025-03-24 at 17 24 20


api/DHSC.FingertipsNext.Modules.HealthData/Mappings/AutoMapperProfiles.cs
```diff
@@ -8,12 +8,17 @@ public class AutoMapperProfiles : Profile
 {
     public AutoMapperProfiles()
     {
+        CreateMap<DeprivationDimensionModel, Deprivation>()
+            .ForMember(dest => dest.Sequence, options => options.MapFrom(src => src.Sequence))
```
MonkeyDate, line 12: I think the first 2 mappings aren't needed as they are the same name/shape

frontend/fingertips-frontend/components/organisms/Heatmap/index.tsx
```diff
@@ -22,6 +23,11 @@ interface HeatmapChartProps {
 
 // Temporary test data for Heatmap
```
MonkeyDate, line 24: do you know if there's a ticket that will sort this?

gareth-allan-bjss: [DHSCFT-230](https://bjss-enterprise.atlassian.net/browse/DHSCFT-230) covers implementing the heatmap properly.

api/DHSC.FingertipsNext.Modules.HealthData/Mappings/AutoMapperProfiles.cs
gareth-allan-bjss: Thanks, I have to confess I didn't really think closely about which fields needed mapping!

api/definition/swagger.yaml
```diff
@@ -449,6 +450,29 @@ components:
               type: number
               example: 123
               description: The value used in the comparison
+        deprivation:
+          $ref: "#/components/schemas/DeprivationData"
+    DeprivationData:
```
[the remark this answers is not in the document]
gareth-allan-bjss: Nice spot, thanks!

api/DHSC.FingertipsNext.Modules.HealthData/HttpClientScripts/get-indicator-data.http
```diff
@@ -31,3 +31,5 @@ GET http://localhost:5144/Indicators/355/data
 GET http://localhost:5144/indicators/247/data?area_codes=N81096&comparison_method=Rag
 ###
 
+### Example which requests data broken down by deprivation
+GET http://localhost:5144/indicators/91894/data?area_codes=E92000001&inequalities=deprivation
```
HowardMayBjss, line 35: Is this validating the response in any way? If the request stops working does it fail? Why not add some validation to the results?

gareth-allan-bjss: I don't think any of the other examples in this file validate results. I was just including it as an example of making the query.

I appreciate what you're saying about having tests at this level, but given we don't run these .http files as part of any pipeline I'm not sure of the value of them.

PR #320, opened 2025-03-21
Resolve issues with inequalities, specifically resolve performance issues

andyevans-bjss:
# Description

Jira ticket: [DHSCFT-434](https://bjss-enterprise.atlassian.net/browse/DHSCFT-434)

Add extra columns to the HealthMeasure table of data type bit that define if the row is aggregated for sex, age or deprivation. These are set to true if the health measure is for the Persons or there is only 1 sex for the related indicator. Similar logic for age and deprivation.
The GetIndicatorDataAsync method in the HealthDataRepository is changed to use these columns instead of calculting aggregate data. This should remove performance issues.
The DataCreator application is changed to calculate these values and add them to the healthdata.csv file.
The database is changed to add the 3 new columns to HealthMeasures. The popuatefingertips.sql is changed to populate the daabase with the new data.

## Validation

Ensure the API unit tests work.
Look at the SQL output from EF and ensure it looks correct.
Deploy the database locally to ensure it deploys and the new data is created
Run the API locally and ensure it gives sensible results.

### DIFF
--- a/database/fingertips-db/Tables/HealthMeasure.sql
+++ b/database/fingertips-db/Tables/HealthMeasure.sql
@@ -101,3 +101,13 @@ CREATE NONCLUSTERED INDEX [YearIndex] ON [dbo].[HealthMeasure]
 	[Year] ASC
 )
 GO
+
+CREATE NONCLUSTERED INDEX InequalitiesAggregatedIndex
+ON [dbo].[HealthMeasure] ([IsSexAggregatedOrSingle],[IsAgeAggregatedOrSingle],[IsDeprivationAggregatedOrSingle])
+INCLUDE ([AgeKey],[AreaKey],[IndicatorKey],[SexKey],[TrendKey],[DeprivationKey],[Count],[Value],[LowerCI],[UpperCI],[Year])
+
+GO
+CREATE NONCLUSTERED INDEX InequalitiesAndIndicatorAggregatedIndex
+ON [dbo].[HealthMeasure] ([IndicatorKey],[IsSexAggregatedOrSingle],[IsAgeAggregatedOrSingle],[IsDeprivationAggregatedOrSingle])
+INCLUDE ([AgeKey],[AreaKey],[SexKey],[TrendKey],[DeprivationKey],[Count],[Value],[LowerCI],[UpperCI],[Year])
+GO


database/fingertips-db/PostDeployment/PopulateFingertips.sql
```diff
@@ -317,7 +317,10 @@ CREATE TABLE #TempHealthData
     Age NVARCHAR(255),
     CategoryType NVARCHAR(MAX),
     Category NVARCHAR(MAX),
-    AgeID INT
+    AgeID INT,
+    IsSexAggregatedOrSingle NVARCHAR(255),
```
stuc-bjss, line 321: is this not a boolean?

andyevans-bjss: nope

andyevans-bjss: it is the string 'true' or 'false' at this point  - it is cast to a bit data type later

HowardMayBjss: I think this is because SQL Server doesn't support a boolean type?

andyevans-bjss: no, it is because BULK INSERT just reads the data from a CSV file and writes it directly to a table. It cannot cast a varchar to a bit automatically.

DataCreator/DataCreator/DataManager.cs
```diff
@@ -6,7 +6,8 @@ namespace DataCreator
     public class DataManager(PholioDataFetcher pholioDataFetcher)
     {
         private readonly PholioDataFetcher _pholioDataFetcher = pholioDataFetcher;
-        private const int MAXNUMBERGPS = 5;
+        private const int MAXNUMBERGPS = 4;
```
HowardMayBjss, line 9: Why the change? Where possible I would favour stability of the data at this point.

andyevans-bjss: 5 mmeans the healthdata.csv file more than 100mb, which is too large for GitHub. Adding in the 3 extra columns in the csv has increased the file size. We can go back to 5 by changing the data type to an init (which will take less space) in another PR

andyevans-bjss: I wanted to get this PR asap as it was delaying Gareth

HowardMayBjss (approved):

PR #324, opened 2025-03-21
DHSCFT-TECH: scroll to top on pages

alibombaye:
# Description

Scrolls to top of the page when navigating between pages. 
When navigating between pages it kept the focus of the screen at the same position as the previous page.
Interaction within the same page do not lose scroll focus

Screenshot 2025-03-21 at 14 13 10
Clicking on View Data now results being on top of the chart page.

Screenshot 2025-03-21 at 14 13 19

where as before it would be somewhere lot further down.

## Validation

Test manually navigate between pages.


frontend/fingertips-frontend/components/pages/results/index.tsx
```diff
@@ -44,6 +44,10 @@ export function SearchResults({
   searchState,
   currentDate,
 }: Readonly<SearchResultsProps>) {
+  useEffect(() => {
```
MonkeyDate, line 47: would it work if this was in the layout rather than duplicated on all pages?

stuc-bjss:
Isn't that what {scroll: false} is for when using the next router thingy `replace(searchStateManager.generatePath(pathname), { scroll: false });` - perhaps investigate why it's not working like it should and remove if if going for the above.

gareth-allan-bjss:
> Isn't that what {scroll: false} is for when using the next router thingy `replace(searchStateManager.generatePath(pathname), { scroll: false });` - perhaps investigate why it's not working like it should and remove if if going for the above.

From [the docs](https://nextjs.org/docs/pages/api-reference/functions/use-router#routerpush), it looks like we're specifically telling it not to scroll to the top of the page if we're passing `scroll: false`. Should we just not be passing this?

frontend/fingertips-frontend/components/pages/results/index.tsx
alibombaye: That would have been nice. But unfortunately that doesn't work 😢

frontend/fingertips-frontend/components/pages/indicator/index.tsx
gareth-allan-bjss, line 1: I'm not sure you need this here. The indicator page already seems to scroll to the top on display, and scrolls the results page to the top when you click the back link.

gareth-allan-bjss: Discussed offline, and the plan is to add it to all pages for consistency.

gareth-allan-bjss (approved):

PR #325, opened 2025-03-21
Feature/dhscft 452 polarity and method from indicator dimension

stuc-bjss:
# Description

Jira ticket: [DHSCFT-452](https://bjss-enterprise.atlassian.net/browse/DHSCFT-452)

## Changes

- Get indicator polarity, confidence level and comparison method from db
- Refactor response to allow indicator data to be returned at the root of the response object
- Updated FE to consume new format response


api/DHSC.FingertipsNext.Modules.HealthData.Service/IndicatorService.cs
```diff
@@ -33,18 +30,51 @@ public class IndicatorService(IHealthDataRepository healthDataRepository, IMappe
...
+    ///     <c>IndicatorWithHealthDataForArea</c> matching the criteria
     /// </returns>
-    public async Task<IEnumerable<HealthDataForArea>> GetIndicatorDataAsync
+    public async Task<IndicatorWithHealthDataForArea?> GetIndicatorDataAsync(
+        int indicatorId,
+        IEnumerable<string> areaCodes,
+        IEnumerable<int> years,
+        IEnumerable<string> inequalities)
+    {
+        var indicatorData = await healthDataRepository.GetIndicatorDimensionAsync(indicatorId );
```
dlzhry2, line 42:
```suggestion
        var indicatorData = await healthDataRepository.GetIndicatorDimensionAsync(indicatorId);
```

Minor change.

api/DHSC.FingertipsNext.Modules.HealthData.Schemas/IndicatorPolarity.cs
```diff
@@ -2,7 +2,19 @@ namespace DHSC.FingertipsNext.Modules.HealthData.Schemas;
...
     NoJudgement,
     LowIsGood,
     HighIsGood
+}
+
+public class IndicatorPolarityConvertor
+{
+    public static IndicatorPolarity Convert(string polarityFromDb)
+    {
+        if(polarityFromDb.Contains("High", StringComparison.OrdinalIgnoreCase)) return IndicatorPolarity.HighIsGood;
```
dlzhry2, line 15: Would we want an consts and to check for full equality? Or could perhaps just use something like StartsWith to be a bit more robust.

Feel free to dismiss though, cos appreciate time constraints.

api/DHSC.FingertipsNext.Modules.HealthData.UnitTests/Services/IndicatorServiceTests.cs
```diff
@@ -31,15 +41,15 @@ public IndicatorServiceTests()
...
-            new object[] { 9, 8, 7, IndicatorPolarity.HighIsGood, BenchmarkOutcome.Better },
-            new object[] { 1, 2, 3, IndicatorPolarity.LowIsGood, BenchmarkOutcome.Better },
-            new object[] { 4, 6, 5, IndicatorPolarity.LowIsGood, BenchmarkOutcome.Similar },
-            new object[] { 9, 8, 7, IndicatorPolarity.LowIsGood, BenchmarkOutcome.Worse },
-            new object[] { 1, 2, 3, IndicatorPolarity.NoJudgement, BenchmarkOutcome.Lower },
-            new object[] { 4, 6, 5, IndicatorPolarity.NoJudgement, BenchmarkOutcome.Similar },
-            new object[] { 9, 8, 7, IndicatorPolarity.NoJudgement, BenchmarkOutcome.Higher }
+            new object[] { 1, 2, 3, "High is good", BenchmarkOutcome.Worse },
+            new object[] { 4, 6, 5, "with high", BenchmarkOutcome.Similar },
+            new object[] { 9, 8, 7, "ci hIgH", BenchmarkOutcome.Better },
```
dlzhry2, line 46: What is the purpose of these tests? Do we anticipate some changes to the polarity descriptors?

api/DHSC.FingertipsNext.Modules.HealthData.UnitTests/Services/IndicatorServiceTests.cs
```diff
@@ -73,13 +75,14 @@ public async Task GetIndicatorData_ShouldReturnExpectedResult()
...
+        _healthDataRepository.GetIndicatorDimensionAsync(1).Returns(testIndicator);
         _healthDataRepository.GetIndicatorDataAsync(1, Arg.Any<string[]>(), [], []).Returns([healthMeasure]);
 
-        var result = await _indicatorService.GetIndicatorDataAsync(1, [], [], [], BenchmarkComparisonMethod.None);
-
-        result.ShouldNotBeEmpty();
-        result.Count().ShouldBe(1);
-        result.ElementAt(0).ShouldBeEquivalentTo(expected);
+        var result = await _indicatorService.GetIndicatorDataAsync(1, [], [], []);
+
```
dlzhry2, line 82:
```suggestion

```

dlzhry2: Might be easier to do in the IDE. I'm not sure what auto-commiting this will do. But basically just hinting to remove the whitespace there.

api/DHSC.FingertipsNext.Modules.HealthData.Service/IndicatorService.cs
andyevans-bjss, line 42: we don't need to make a seperate database query to get the polarity and comparison method - it should be done in GetIndicatorDataAsync

andyevans-bjss (changes requested):
we don't need to make a seperate database query to get the polarity and comparison method - it should be done in GetIndicatorDataAsync.

dlzhry2 (commented):
Looks good. Just a couple of minor comments added.

api/DHSC.FingertipsNext.Modules.HealthData.UnitTests/Services/IndicatorServiceTests.cs
stuc-bjss: I asked about linting/style settings and was told to turn them off

api/DHSC.FingertipsNext.Modules.HealthData.Service/IndicatorService.cs
dlzhry2: Was having a think about this when I looked at this new method. If we were to add it to GetIndicatorDataAsync this might mean adding some additional fields to grab from the `IndicatorDimensionModel` that we are already including in that query.

However this would mean adding it to each health data point when we only really need it once at the top level of the object. Maybe there's a nice way to do that in entity framework i.e. selecting certain items just once.

api/DHSC.FingertipsNext.Modules.HealthData.Schemas/IndicatorPolarity.cs
stuc-bjss: polarity descriptions in the database are verbose and do not match the enums we've defined in the code, I don't know the db contents well enough to know how much variance there is in them so I just went for picking out keywords

dlzhry2: Ok, worth having @andyevans-bjss confirm. Appreciate they aren't in a separate lookup table but I think there's a consistent set of values that they can take:

image

api/DHSC.FingertipsNext.Modules.HealthData.Service/IndicatorService.cs
stuc-bjss: Not a .NET dev so I know nothing about entity framework, is what Dan is suggesting what also what you're suggesting @andyevans-bjss which if I understand correctly is duplicate all the data for the indicator onto every result from `GetIndicatorDataAsync` and then stripping it all off again in the next step? Is it more or less effort for the DB to handle such a trivial call separately or to tag it onto all the results of another really complicated one? Please advise.

api/DHSC.FingertipsNext.Modules.HealthData.Schemas/IndicatorPolarity.cs
andyevans-bjss: there are 3 distinct polarity values in the IndicatorsDimension table

api/DHSC.FingertipsNext.Modules.HealthData.Service/IndicatorService.cs
andyevans-bjss: yeah basically

andyevans-bjss: lets discuss next week

README.md
HowardMayBjss, line 1: I think you have some auto-formatting options coming into play here. Can you look to identify what is causing these formatting changes please.

api/DHSC.FingertipsNext.Modules.HealthData.Schemas/BenchmarkComparisonMethod.cs
```diff
@@ -1,7 +1,37 @@
...
+    Rag,
+    Quintiles
+}
+
+public static partial class BenchmarkComparisonMethodConvertor
+{
+    [GeneratedRegex(@"\d+(\.\d+)?")]
+    private static partial Regex NumberRegex();
+    
+    public static BenchmarkComparisonMethod Convert(string method)
```
HowardMayBjss, line 18: We can use an AutoMapper to do this conversion - I have examples on my branch. 386_Benchmarking-quintiles

api/DHSC.FingertipsNext.Modules.HealthData.Schemas/IndicatorWithHealthDataForArea.cs
```diff
@@ -0,0 +1,29 @@
+using System.Text.Json.Serialization;
+
+namespace DHSC.FingertipsNext.Modules.HealthData.Schemas;
+
+public class IndicatorWithHealthDataForArea
```
HowardMayBjss, line 5: This holds data for multiple Areas? If so it is better named IndicatorWithHealthDataForAreas

api/DHSC.FingertipsNext.Modules.HealthData.Schemas/IndicatorWithHealthDataForArea.cs
```diff
@@ -0,0 +1,29 @@
...
+    [JsonPropertyName("polarity")]
+    [JsonConverter(typeof(JsonStringEnumConverter))]
+    public IndicatorPolarity Polarity { get; set; }
+    
+    [JsonPropertyName("benchmarkMethod")]
+    [JsonConverter(typeof(JsonStringEnumConverter))]
+    public BenchmarkComparisonMethod BenchmarkMethod { get; set; }
+    
+    [JsonPropertyName("benchmarkConfidenceLevel")]
+    public float BenchmarkConfidenceLevel { get; set; }
```
HowardMayBjss, line 25: I'm looking again at this and think we might need to combine BenchmarkComparisonMethod and ConfidenceLevel - lets  discuss this.

api/DHSC.FingertipsNext.Modules.HealthData.Service/IndicatorService.cs
HowardMayBjss: Stuart made the point that Quintiles and CI benchmarking are separate and distinct implementations. This means we either need to have an additional call to the DB to know which BenchmarkMethod to use OR we need to somehow combine the implementations into a single DB call. It is not clear to me how this will work or if it is easily possible. I suggest we leave the second DB call in at this point and when Quintiles is in we can take a view as to what work is required to reduce this to a single DB call.

api/DHSC.FingertipsNext.Modules.HealthData.Service/IndicatorService.cs
```diff
@@ -33,25 +30,57 @@ public class IndicatorService(IHealthDataRepository healthDataRepository, IMappe
...
+        IEnumerable<string> areaCodes,
+        IEnumerable<int> years,
+        IEnumerable<string> inequalities)
+    {
+        var indicatorData = await healthDataRepository.GetIndicatorDimensionAsync(indicatorId);
+        if (indicatorData == null) return null;
+        var method = _mapper.Map<BenchmarkComparisonMethod>(indicatorData.BenchmarkComparisonMethod);
+        var polarity = _mapper.Map<IndicatorPolarity>(indicatorData.Polarity); 
+        
+        var indicator = new IndicatorWithHealthDataForAreas()
```
HowardMayBjss, line 47: I should just return this directly without the redundant variable.

HowardMayBjss:
The UI tests failed on the last run - so best check that out as well

andyevans-bjss (approved):

PR #327, opened 2025-03-21
Feature/dhscft 144 compare areas confidence intervals

priya-thorpe-bjss:
# Description

Jira ticket: [DHSCFT-144](https://bjss-enterprise.atlassian.net/browse/DHSCFT-144)

## Changes

- Add confidence interval checkbox to bar chart embedded table component
- Confidence intervals display when check box is clicked in bar chart embedded table component
- Remove CI from search state manager and its tests
- Update method of displaying confidence intervals in line chart component to useState
- Remove search state from confidence interval checkbox component and refactor component 

## Validation

The error bars should be on either side of the value bar like the lower bars, but for the data we have in testing, the error bars are smaller for the other value bars.

Screenshot 2025-03-25 at 09 56 23

Screenshots have been updated to include the CIs and checkbox


frontend/fingertips-frontend/components/molecules/ConfidenceIntervalCheckbox/confidenceIntervalCheckbox.test.tsx
```diff
@@ -1,73 +1,19 @@
...
-        searchState={state}
+        setShowConfidenceIntervalsData={jest.fn()}
       />
     );
     await userEvent.click(screen.getByRole('checkbox'));
 
     expect(screen.getByRole('checkbox')).toBeChecked();
   });
-
-  it('should update the url with the chart name when the checkbox is clicked', async () => {
```
gareth-allan-bjss, line 39: I think you could probably still have a unit test that verifies `setShowConfidenceIntervalsData` is called when the checkbox is clicked.

frontend/fingertips-frontend/components/organisms/SparklineChart/index.tsx
```diff
@@ -2,16 +2,37 @@
...
+  value: (number | undefined)[];
   maxValue: number;
+  errorBarValues: (number | undefined)[];
+  showConfidenceIntervalsData: boolean;
 }
 export function SparklineChart({
   value,
   maxValue,
+  errorBarValues,
+  showConfidenceIntervalsData,
```
gareth-allan-bjss, line 17: `showConfidenceIntervalsData` might be better named `showErrorBars`, as it's then consistent with the name of the `errorBarValues` prop.

frontend/fingertips-frontend/components/organisms/SparklineChart/index.tsx
```diff
@@ -2,16 +2,37 @@
...
+  const series: Highcharts.SeriesOptionsType[] = [
+    { type: 'bar', data: [value] },
+  ];
+
+  if (showConfidenceIntervalsData) {
+    series.push({
+      type: 'errorbar',
+      data: [errorBarValues],
+    });
+  }
```
gareth-allan-bjss, line 33: Do we need to be careful here to avoid clicking the CI checkbox resulting in the bars changing length? We had to make the error series hidden on the line chart and just show/hide it based on the checkbox to avoid this issue.

frontend/fingertips-frontend/components/organisms/SparklineChart/index.tsx
```diff
@@ -2,16 +2,37 @@
...
   value,
   maxValue,
+  errorBarValues,
+  showConfidenceIntervalsData,
 }: Readonly<SparklineChartProps>) {
-  const sparklineOptions = {
+  const [options, setOptions] = useState<Highcharts.Options>();
+  const loadHighchartsModules = async (callback: () => void) => {
+    await import('highcharts/highcharts-more').then(callback);
+  };
```
gareth-allan-bjss, line 22: This is the same function as in the line chart. It might be worth extracting it to a helper somewhere to avoid duplication.

frontend/fingertips-frontend/components/organisms/SparklineChart/SparklineChart.test.tsx
```diff
@@ -3,13 +3,22 @@ import { SparklineChart } from '@/components/organisms/SparklineChart/index';
...
+  it('Should render the highcharts react component', async () => {
+    const mockValue = [48];
     const maxValue = 100;
-    render(<SparklineChart value={mockValue} maxValue={maxValue} />);
+    render(
+      <SparklineChart
+        value={mockValue}
+        maxValue={maxValue}
+        errorBarValues={[5, 10]}
+        showConfidenceIntervalsData={false}
```
gareth-allan-bjss, line 14: If this test is to validate that we can successfully render the chart when appropriate options are passed, then we should probably set `showConfidenceIntervalsData` to `true`, as it'll do more rendering in that case.

frontend/fingertips-frontend/.test/spec/snaps/webkit/fully_integrated_e2e_tests/core_journeys.spec.ts/BOTH-SUBJECT-AND-AREA-then-select-ONE-INDICATOR-and-ONE-AREA-then-check-the-charts-page-barChart-component.png
gareth-allan-bjss, line 1: Why are these screenshots disappearing? Is this expected @williamwillis11?

frontend/fingertips-frontend/components/organisms/SparklineChart/index.tsx
```diff
@@ -2,16 +2,37 @@
...
+
+  const series: Highcharts.SeriesOptionsType[] = [
+    { type: 'bar', data: [value] },
+  ];
+
+  if (showConfidenceIntervalsData) {
+    series.push({
+      type: 'errorbar',
+      data: [errorBarValues],
+    });
```
gareth-allan-bjss, line 32: I think, unless you've already agreed otherwise, we need to update the formatting of these error bars to match the designs. They're black, instead of the grey on the designs, and the lines look thinner too.

frontend/fingertips-frontend/components/organisms/SparklineChart/index.tsx
priya-thorpe-bjss: I have changed to confidenceIntervalValues, to remain consistent everywhere.

frontend/fingertips-frontend/components/organisms/SparklineChart/index.tsx
priya-thorpe-bjss: I have changed to match the line chart bars

frontend/fingertips-frontend/components/organisms/SparklineChart/index.tsx
priya-thorpe-bjss: I have extracted to chart helpers and replaced in line chart and bar chart embedded table components

frontend/fingertips-frontend/components/organisms/SparklineChart/index.tsx
priya-thorpe-bjss: Its doesn't seem to be updating on here. I have tried pushing this change twice now.

frontend/fingertips-frontend/components/organisms/SparklineChart/index.tsx
gareth-allan-bjss: Thanks. I did see that they were stable in my own testing. Does the max value account for the possibility an upper CI will be bigger than the largest max value though? Perhaps we should be calculating it using the upper CI field, rather than the value field?

priya-thorpe-bjss: No it only takes into account the largest value. Do we want the bar to extend all the way to meet the upper CI value though?

frontend/fingertips-frontend/.test/spec/snaps/webkit/fully_integrated_e2e_tests/core_journeys.spec.ts/BOTH-SUBJECT-AND-AREA-then-select-ONE-INDICATOR-and-ONE-AREA-then-check-the-charts-page-barChart-component.png
williamwillis11: not expected. its a facet on the fact that the screenshot comparisons are only set to warn

so if someone makes changes that should also require screenshot changes, otherwise it fails - but they dont update the screenshots at that point in time and they merge into main, the next time someone whos up to date with main updates the screenshots they also mop up this mess left behind 

for this file, it looks like in the past the barChart was incorrectly assigned to the One indicator + One area scenario, but that wasnt right according to the confluence page so someone corrected the scenario config but didnt update the screenshots - ill see if i can find the commit/PR

gareth-allan-bjss: OK, yeah, makes sense. I think the bar chart component referred to was the placeholder one that got removed when we implemented the inequalities bar chart.

frontend/fingertips-frontend/components/organisms/SparklineChart/index.tsx
gareth-allan-bjss: Also, I think the whiskers need to be a bit longer. They look a lot smaller to me than the designs. The line probably needs to be a bit thicker too. Sorry to nitpick, but I think they're quite hard to see as they are.

frontend/fingertips-frontend/components/organisms/SparklineChart/index.tsx
gareth-allan-bjss: I just wondered if we needed to worry about leaving enough space to the right of the maximum bar to draw the upper CI value, but it seems to be OK from my testing. I'm happy for you to resolve this if you think there's nothing to do here.

frontend/fingertips-frontend/components/organisms/SparklineChart/index.tsx
priya-thorpe-bjss: I think the grey whiskers are difficult to see on the blue bars. The length of the whiskers is determined by the data we have. I have made the line thicker and the whisker length longer.

priya-thorpe-bjss: I have noticed changing the colour of the whiskers to grey means it now blends into the background of the bench mark row. I will check with BA on what to do with that.

gareth-allan-bjss: Yes, I saw that the grey is difficult to see on the blue bars, but we will be making these bars different colours depending on the benchmark. I think the England bar will stay blue though, so we should definitely discuss with design.

gareth-allan-bjss (approved):

PR #328, opened 2025-03-24
Refactor deployments to use self-hosted runner

epercy-bjss:
# Description

Jira ticket: [DHSCFT-246](https://bjss-enterprise.atlassian.net/browse/DHSCFT-246)

Refactor GitHub Actions workflow to use the GitHub Self-Hosted Runner, provisioned through Azure Container App Jobs. This runner will scale automatically (maximum of 10 instances) to run all deployment jobs from within the Azure Virtual Network. NSG rules have been added to allow secure private routing to the required subnets.

## Changes

- Build and pushing the container images is now performed through AZ ACR CLI to allow access to the private container registry without requiring the Docker CLI
- Database deployment now running through the Linux self-hosted runner using SQLPackage for Linux
- Small change added to the post deployment script to improve performance on the HealthMeasure import
- Trend analysis running on self-hosted runner
- Search Setup running on self-hosted runner
- Container App deployments running through Terraform on the self-hosted runner
- Playwright dependencies added to the self-hosted runner and networking modified to permit access to the App Gateway for the frontend
- Workflow conditions modified to allow deployments from any branch if the are run manually (workflow_dispatch)

## Validation
Deployment has been tested from branch and all steps have now passed


.github/workflows/fingertips-workflow.yml
```diff
@@ -352,23 +350,31 @@ jobs:
...
-            --account-name '${{ secrets.SQL_DATA_STORAGE_ACCOUNT }}' `
-            --auth-mode 'login' `
+          az storage blob upload-batch \
+            --destination "${{ vars.SQL_DATA_BLOB_CONTAINER }}" \
+            --source "DataCreator/DataCreator/data/out" \
+            --account-name "${{ secrets.SQL_DATA_STORAGE_ACCOUNT }}" \
+            --auth-mode "login" \
             --overwrite true
-
+
```
dlzhry2, line 362: Optional - but could remove this whitespace.

.github/workflows/e2e-tests.yml
```diff
@@ -9,9 +9,9 @@ env:
 on:
   workflow_call:
     inputs:
-      e2e-test-container-image:
-        description: 'container image to run the e2e tests in - for CD test executions'
-        required: false
+      playwright-mode:
```
dlzhry2, line 12: Just to check my understanding, is this called from other workflows rather than manual dispatch, so we don't need to specifically call out the valid options?

dlzhry2 (approved):
Looks good to me. One question and one minor format comment.

.github/workflows/fingertips-workflow.yml
epercy-bjss: Done

.github/workflows/e2e-tests.yml
epercy-bjss: It is called only from the main `fingertips-workflow.yml`, the `playwright-mode` was added in order to conditionally run the relevant job within the `e2e-tests.yml` reusable workflow. It is used for running playwright either against the docker stack, or from the locally installed playwright tool on the self-hosted runner. Breaking it into two jobs allows us to use either hosted or self-hosted runner depending on if it's docker or not.

PR #329, opened 2025-03-24
Feature/dhscft tech api refactor

andyevans-bjss:
# Description

General code tidy

## Validation

Unit tests run

### DIFF
--- a/api/DHSC.FingertipsNext.Modules.Area.Repository/AreaRepository.cs
+++ b/api/DHSC.FingertipsNext.Modules.Area.Repository/AreaRepository.cs
@@ -63,7 +63,8 @@ public class AreaRepository : IAreaRepository
     /// <param name="includeSiblings"></param>
     /// <param name="childAreaType"></param>
     /// <returns></returns>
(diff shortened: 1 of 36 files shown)


api/DHSC.FingertipsNext.Modules.Area.Repository/AreaRepository.cs
```diff
@@ -86,25 +73,25 @@ string childAreaType
...
 
-                .Include(a => a.Children)
-                .ThenInclude(x => x.Parents)
+                .Include(area => area.Children)
+                .ThenInclude(area => area.Parents)
 
-                .Include(a => a.Children)
-                .ThenInclude(x => x.AreaType)
+                .Include(area => area.Children)
+                .ThenInclude(area => area.AreaType)
```
steves-bjss, line 84: I think this makes things a little less readable here.  Could we replace x with a different name to 'area' to distinguish that its working on either the child or parent.

steves-bjss (approved):

PR #330, opened 2025-03-24
set default ages for indicators 93088 and 93015

andyevans-bjss:
# Description

2 PoC indicators do not hve 'default' ages currently, so no data is returned f. This PR fixes that by setting the appropriate age as the default.

## Validation

Import the data to the database, run a dataase query and check if the appopriate data is returned


search-setup/assets/indicators.json
```diff
@@ -1 +0,0 @@
-../../trend-analysis/TrendAnalysisApp/SearchData/assets/indicators.json
```
dlzhry2, line 1: This symlink needs to remain in place.

For full visibility (as I haven't had time to document in Confluence), when there is an adjustment to the `indicators.json` file output by the DataCreator, that now needs to go to the symlinked location.

Then, when the pipeline is run, the trend-analysis app will ingest and update that file passing the result to the search-setup so that accurate trend data is included in the AI Search index.

Additionally, I can run it locally to keep the file in sync so it shows what the trend data would look like, but strictly speaking you only need the first step.

DataCreator/DataCreator/DataManager.cs
```diff
@@ -174,6 +174,13 @@ private static void SetAggregateFlags(List<IndicatorWithAreasAndLatestUpdate> in
                 else //the associated indicator has more than 1 age, the health measure could be for 'All ages', or others. If it is 'All ages' set the flag to true, otherwise false
                     healthMeasure.IsAgeAggregatedOrSingle = healthMeasure.Age == "All ages";
 
+                //there are 2 indicators that have multiple ages but no aggregate age
+                //using the same behaviour as current Fingertips set the default age specifically
+                if (healthMeasure.IndicatorId == 93015 && healthMeasure.Age == "19+ yrs")
```
dlzhry2, line 179: Now that we have default constraints for these 2 indicators, I need to make a change to the trend-analysis app so it treats these 2 indicators accordingly.

I'm happy to pick this up once you've made the below change regarding the symlinked file.

search-setup/assets/indicators.json
dlzhry2: Actioned

dlzhry2 (approved):

PR #331, opened 2025-03-24
DHSCFT-372: Add trends to the bar chart embedded table

dlzhry2:
# Description

Jira ticket: [DHSCFT-372](https://bjss-enterprise.atlassian.net/browse/DHSCFT-372)

Don't be intimidated by the number of files - it's mostly the screenshots.

There is a compare areas bar chart table which appears if more than one area is selected for an indicator. This adds a 'Recent trend' column to the table in addition to the relevant trend marker.

Please refer to the screenshots to see how this now looks (there are 4 examples I believe where this is rendered, although because it is local testing all examples result in 'No trend available'). I tested locally however with the DB and trend analysis having run and those with arrows and colours looked fine.


frontend/fingertips-frontend/components/organisms/BarChartEmbeddedTable/index.tsx
```diff
@@ -100,6 +103,9 @@ export function BarChartEmbeddedTable({
               <Table.CellHeader>
                 {BarChartEmbeddedTableHeadingEnum.AreaName}
               </Table.CellHeader>
+              <Table.CellHeader>
+                {BarChartEmbeddedTableHeadingEnum.RecentTrend}
+              </Table.CellHeader>
```
gareth-allan-bjss, line 108: The designs have the `Recent trend` column to the right of the `Period` column. Is this deliberately in a different place, or is it an omission?

frontend/fingertips-frontend/lib/chartHelpers/chartHelpers.ts
```diff
@@ -72,6 +72,7 @@ export function getMostRecentData(data: HealthDataPoint[]) {
   return data.length > 0
     ? {
         count: data[0].count,
+        recentTrend: data[0].trend,
```
gareth-allan-bjss, line 75: Could this function just return `data[0]`? It seems odd to map all these fields through manually.

frontend/fingertips-frontend/components/organisms/BarChartEmbeddedTable/index.tsx
dlzhry2: Very good point - I had been looking at the Figma link which did not even seem to have a period comment. Can see from the screenshot on the ticket that you're correct - and I'm sure there's places in Figma with more up to date designs.

Have adjusted and rerun screenshots + snapshots.

frontend/fingertips-frontend/lib/chartHelpers/chartHelpers.ts
dlzhry2: Yes, I've been trying to adjust as little code as possible given the timeframes but this would make sense - no need to remap/introduce a new type.

I had a look to see if we had a common lib in use, as this is just a basic array utility rather than anything specific to charts. In the end just did a minor adjustment.

frontend/fingertips-frontend/components/organisms/BarChartEmbeddedTable/index.tsx
gareth-allan-bjss: For the record, I think this is the appropriate design link: https://www.figma.com/design/QXpgnbZvVOtP1u6fqrN6Aw/FTSN-Ready-for-Dev?node-id=4116-13286&t=kEbatdrxal06Ts4m-0

frontend/fingertips-frontend/lib/chartHelpers/chartHelpers.test.ts
```diff
@@ -432,27 +431,6 @@ describe('isEnglandSoleSelectedArea', () => {
...
-    expect(result).toEqual(expected);
-  });
-
-  it('should return undefined when there is no data passed', () => {
-    const result = getMostRecentData([]);
-
-    expect(result).toEqual(undefined);
-  });
-});
-
```
gareth-allan-bjss, line 455: Why remove the unit tests for this function? I think there's still value in verifying it returns undefined if an empty array is passed.

dlzhry2: My bad, I had thought to remove the function entirely before deciding to keep as it was used in two places. I'll put it back now.

dlzhry2: Added it back. Got a conflict in this file so will look just after standup.

gareth-allan-bjss (approved):

PR #344, opened 2025-03-26
Feature/dhscft 469 compare areas hover over bars

priya-thorpe-bjss:
# Description

Jira ticket: [DHSCFT-469](https://bjss-enterprise.atlassian.net/browse/DHSCFT-469)

## Changes
- Customised hover with benchmark/ group prefix's and CI percentages .
- Updated sparkline component to change display of error bars to appear and disappear.

## Validation

Screenshot 2025-03-26 at 12 07 47


frontend/fingertips-frontend/components/organisms/SparklineChart/index.tsx
```diff
@@ -25,31 +32,43 @@ export function SparklineChart({
...
-      whiskerLength: '50%',
-      lineWidth: 3,
-    });
-  }
+    if (label === SparklineLabelEnum.Benchmark) {
+      category = 'Benchmark: ';
+    }
+    if (label === SparklineLabelEnum.Group) {
+      category = 'Group: ';
+    }
```
gareth-allan-bjss, line 51: Instead of passing in a label and deciding what text to show based on its value, why not make label an optional string, and just show it if it's present?

frontend/fingertips-frontend/components/organisms/SparklineChart/index.tsx
```diff
@@ -25,31 +32,43 @@ export function SparklineChart({
...
+
+    return [
+      `<b>${category}${area}</b><br/>${year}<br/><br/><span style="color:${point.color}">${symbol}</span> ${value} ${measurementUnit}`,
```
gareth-allan-bjss, line 54: All three ACs mention including the "Similar to England", "Worse than England", etc. text in the hover, which I don't think you have. Is that a deliberate omission to be implemented later? We need to make sure it'll be covered at some point.

frontend/fingertips-frontend/components/organisms/SparklineChart/index.tsx
```diff
@@ -25,31 +32,43 @@ export function SparklineChart({
...
+  year,
+  measurementUnit,
 }: Readonly<SparklineChartProps>) {
   const color = getBenchmarkColour(benchmarkOutcome as BenchmarkLabelType);
   const [options, setOptions] = useState<Highcharts.Options>();
 
-  const series: Highcharts.SeriesOptionsType[] = [
-    { type: 'bar', data: [value], color },
-  ];
+  const formatSparklineTooltips = (point: Highcharts.Point, symbol: string) => {
```
gareth-allan-bjss, line 43: I agree it's probably not valuable to test Highcharts is showing a tooltip when we expect (as we're just testing Highcharts at that point), but it would probably be worth having some unit tests that validate `formatSparklineTooltips` returns the expected value. Then we can validate it's doing the right thing with the `Benchmark: `/`Group: ` prefixes, etc.

frontend/fingertips-frontend/components/organisms/SparklineChart/index.tsx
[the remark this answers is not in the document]
priya-thorpe-bjss: The confidence interval bars are black and it would get lost when applied.

priya-thorpe-bjss: The benchmark work has now gone in, so I will add it to this PR.

frontend/fingertips-frontend/components/organisms/SparklineChart/index.tsx
```diff
@@ -46,19 +54,30 @@ export function SparklineChart({
...
     }
 
+    if (benchmarkOutcome === BenchmarkOutcome.Similar) {
+      benchmarkLabel = `${outcome} to England`;
+    } else if (
+      benchmarkOutcome &&
+      benchmarkOutcome !== BenchmarkOutcome.NotCompared
+    ) {
+      benchmarkLabel = `${outcome} than England`;
+    }
```
gareth-allan-bjss, line 75: We're going to need to do this in several places, so it's probably worth extracting it as a utility somewhere and just passing the resulting string as a prop of this component. I think it will also need to deal with quintiles (we don't want it saying `Middle than England`). It's probably worth checking in with @stuc-bjss to see if he's considered how this should work as part of the benchmarking work.

Also, I think ideally we should pass the benchmark area name into this logic too, rather than hardcoding it to `England`. We've tried to avoid hardcoding anything that assumes the benchmark is England.

frontend/fingertips-frontend/components/organisms/SparklineChart/index.tsx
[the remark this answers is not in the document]
gareth-allan-bjss: I think we need to speak to design about this. As it is, we're not matching the designs, but there are obviously some questions to be answered.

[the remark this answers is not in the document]
gareth-allan-bjss: It doesn't need to hold up this PR though.

frontend/fingertips-frontend/components/organisms/SparklineChart/SparklineChart.test.tsx
```diff
@@ -25,4 +34,186 @@ describe('SparklineChart', () => {
...
+          value={mockValue}
+          maxValue={maxValue}
+          confidenceIntervalValues={[5, 10]}
+          showConfidenceIntervalsData={true}
+          label={SparklineLabelEnum.Benchmark}
+          area={'mockArea'}
+          year={2000}
+          measurementUnit={''}
+        />
+      );
```
gareth-allan-bjss, line 55: Why is the `SparklineChart` component here if this is a test for the `sparklineTooltipContent` function?

frontend/fingertips-frontend/components/organisms/SparklineChart/SparklineChart.test.tsx
```diff
@@ -25,4 +34,186 @@ describe('SparklineChart', () => {
...
+            maxValue={maxValue}
+            confidenceIntervalValues={[5, 10]}
+            showConfidenceIntervalsData={true}
+            label={SparklineLabelEnum.Benchmark}
+            area={'mockArea'}
+            year={2000}
+            measurementUnit={''}
+          />
+        );
+      });
```
gareth-allan-bjss, line 76: Again, why render the `SparklineChart` if this is a test for `sparklineTooltipContent`? I don't think it'll affect the assertion you're actually making in this test.

frontend/fingertips-frontend/components/organisms/SparklineChart/index.tsx
```diff
@@ -22,8 +27,46 @@ interface SparklineChartProps {
...
+
+  if (benchmarkOutcome === BenchmarkOutcome.Similar) {
+    benchmarkLabel = `${outcome} to England`;
+    comparisonLabel = `(${comparison}${measurementUnit})`;
+  } else if (
+    benchmarkOutcome &&
+    benchmarkOutcome !== BenchmarkOutcome.NotCompared
+  ) {
+    benchmarkLabel = `${outcome} than England`;
+    comparisonLabel = `(${comparison}${measurementUnit})`;
```
gareth-allan-bjss, line 64: I don't think we should be using the `measurementUnit` in the comparison label here. Assuming this is the measurement unit we display on the chart for the indicator, it could be `per 100,000` instead of `%` and I don't think it makes sense to display the CIs as `95per 100,000`. I think this can probably be a hardcoded `%` symbol?

frontend/fingertips-frontend/components/organisms/SparklineChart/SparklineChart.test.tsx
```diff
@@ -25,4 +34,186 @@ describe('SparklineChart', () => {
       )
     ).toBeInTheDocument();
   });
+
+  describe('sparklineTooltipContent', () => {
```
gareth-allan-bjss, line 38: I think we should have a specific test to verify that if we pass a comparison method of "Not Compared" that we don't get a benchmark label or comparison label returned.

gareth-allan-bjss (approved):

PR #349, opened 2025-03-26
DHSCFT-512 chart page filter UI test

williamwillis11:
# Description

Jira ticket: [DHSCFT-512](https://bjss-enterprise.atlassian.net/browse/DHSCFT-512)

Add in testing at the isolated UI level to cover changing the area filters on the chart page and seeing different chart components therefore appear.

## Changes

- Added in step to the main isolated ui test to change area filter on chart page
- added in check to assert a new component has been displayed
- refactored areafilter into its own page which is now inherited from the pages that need it
- refactored getScenarioConfig so that it returns a config object which makes checkChart simpler
- deleted locally all the screenshots and ran the update (as i also upped the allowed ratio difference threshold)

## Validation

ui and e2e tests passing


frontend/fingertips-frontend/package.json
```diff
@@ -20,7 +20,7 @@
     "test": "jest",
     "test-ci": "jest --ci --silent",
     "test-ui-local-mocks": "playwright test isolated_ui_tests --ignore-snapshots",
-    "test-e2e-local-docker": "cross-env MOCK_SERVER=false npm run start:all && playwright test fully_integrated_e2e_tests || true && npm run stop:all",
+    "test-e2e-local-docker": "cross-env MOCK_SERVER=false npm run start:all && playwright test fully_integrated_e2e_tests --ui || true && npm run stop:all",
```
Nayira-BJSS, line 23: Left the ui flag here by error :)

Nayira-BJSS (approved):

PR #350, opened 2025-03-26
DHSCFT-386: Perform Quintile Benchmark comparison calculation in the API

HowardMayBjss:
# Description

Jira ticket: [DHSCFT-386](https://bjss-enterprise.atlassian.net/jira/software/projects/DHSCFT/boards/823?selectedIssue=DHSCFT-386)

This adds the ability for the backend to perform Quintile benchmark comparison calculations. This complements the existing CI based benchmarking. 
 
## Changes

- Add a stored procedure to return selected healthdata with Quintile value as a benchmark comparison output.
- Call stored procedure from API if the Requested Indicator has a benchmark Comparison method of Quintile
- Entity Framework code to marshal results into correct format
- API tests to validate response
- Miscellaneous updates to tests to correct

## Validation

New API tests to exercise Quintile calculation
The results have been manually compared with existing site where possible
- this suggests some unexpected behaviour on the existing site which
Pipeline Green

Some observations are to be resolved:
* Some indicators have a large proportion of 'zero count' data which impacts the Quintile calculation - how such data should be handled needs to be resolved
* The existing site returns unexpected values which are different to ours. This also needs to be discussed and resolved.


api/DHSC.FingertipsNext.Modules.HealthData.Service/IndicatorService.cs
```diff
@@ -61,26 +74,57 @@ private async Task<IEnumerable<HealthDataForArea>> GetIndicatorAreaDataAsync
...
+        {
+            // get the data from the database
+            healthMeasureData = await healthDataRepository.GetIndicatorDataWithQuintileBenchmarkComparisonAsync(
+                indicatorId,
+                areaCodesForSearch.ToArray(),
+                years.Distinct().ToArray(),
+                areaType
+                );
+
+            var retVal= healthMeasureData
```
stuc-bjss, line 97:
```suggestion
            return healthMeasureData
```

api/DHSC.FingertipsNext.Modules.HealthData/HttpClientScripts/test-indicators-benchmark-quintiles-response.json
```diff
@@ -0,0 +1,301 @@
+{
+  "name": "People reporting Alzheimer's disease or dementia",
+  "startDate": "2015-03-26T14:11:02.2333333",
```
HowardMayBjss, line 3: These dates shouldn't be here and are non-deterministic.

database/fingertips-db/StoredProcedures/GetIndicatorDataWithQuintileBenchmarkComparison.sql
```diff
@@ -0,0 +1,156 @@
...
+		    sex.Name as SexDimensionName,
+		    sex.HasValue as SexDimensionHasValue,
+		    trendDim.Name as TrendDimensionName,
+		    ageDim.Name as AgeDimensionName,
+		    ageDim.HasValue as AgeDimensionHasValue,
+		    imd.Name as DeprivationDimensionName,
+		    imd.[Type] as DeprivationDimensionType,
+		    imd.[Sequence] as DeprivationDimensionSequence,
+		    imd.HasValue as DeprivationDimensionHasValue,
+		    Count,
```
HowardMayBjss, line 41: We may need to exclude data with Count set to zero or null.

api/DHSC.FingertipsNext.Modules.HealthData.Repository/HealthDataRepository.cs
```diff
@@ -93,4 +96,45 @@ public async Task<IEnumerable<HealthMeasureModel>> GetIndicatorDataAsync(int ind
...
+        var yearsOfInterest = new SqlParameter("@RequestedYears", YearsTable)
+        {
+            SqlDbType = SqlDbType.Structured,
+            TypeName = "YearList"
+        };
+
+        var areaTypeOfInterest = new SqlParameter("@RequestedAreaType", areaTypeKey);
+        var requestedIndicatorId = new SqlParameter("@RequestedIndicatorId", indicatorId);
+
+        var denormalisedHealthData = await _dbContext.DenormalisedHealthMeasure.FromSql
```
andyevans-bjss, line 130: use @$"sql goes here " rather than """

andyevans-bjss (changes requested):
looks really good - a couple of simple changes

api/DHSC.FingertipsNext.Modules.HealthData.Service/IIndicatorsService.cs
```diff
@@ -13,8 +13,9 @@ public interface IIndicatorsService
     ///     supplying one or more area codes and one or more years in the query string.
     /// </summary>
     /// <param name="indicatorId">The unique identifier of the indicator.</param>
-    /// <param name="areaCodes">A list of area codes.</param>
-    /// <param name="years">A list of years.</param>
+    /// <param name="areaCodes">A list of area codes. Up to 10 distinct area codes can be requested.</param>
```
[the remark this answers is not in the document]
HowardMayBjss: some rebase issue

database/fingertips-db/StoredProcedures/GetIndicatorDataWithQuintileBenchmarkComparison.sql
```diff
@@ -0,0 +1,158 @@
...
+            hm.TrendKey = trendDim.TrendKey
+	WHERE
+		(
+		--- This ensures we are only dealing with Aggregate data
+	        hm.IsSexAggregatedOrSingle=1 AND hm.IsAgeAggregatedOrSingle=1 AND hm.IsDeprivationAggregatedOrSingle=1
+	    )
+        AND
+        (
+            areaDim.AreaType = @RequestedAreaType
+        --- This is special case handling for data which has a dual identify as both district and county.
```
dlzhry2, line 80:
```suggestion
        --- This is special case handling for data which has a dual identity as both district and county.
```

database/fingertips-db/StoredProcedures/GetIndicatorDataWithQuintileBenchmarkComparison.sql
dlzhry2: From reading through, I understand the stored proc returns the comparison outcome rather than finding the first value in each quintile. Unless we ask someone, or legacy fingertips stores the quintile values for indicators so we can perform some sort of comparison, then it can only be a guess as to the desired behaviour.

Could exclude null but might we want 0s? e.g. if a GP practice had a 0 count for a particular indicator then that is valid data and would affect a quintile outcome? Either way, some testing at least of the outcomes should give us a reasonable confidence and could always put a comment in here/ticket to come back if we're worried.

dlzhry2 (commented):
Minor comment and thought on inclusion of Count added. This looks like great work and must have taken some careful thought.

database/fingertips-db/StoredProcedures/GetIndicatorDataWithQuintileBenchmarkComparison.sql
```diff
@@ -0,0 +1,156 @@
...
+            OR (
+               @RequestedAreaType = 'districts-and-unitary-authorities'
+               AND areaDim.Code LIKE 'E09%' OR areaDim.Code LIKE 'E08%' OR areaDim.Code LIKE 'E06%'
```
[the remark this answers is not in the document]
HowardMayBjss: Done!

database/fingertips-db/StoredProcedures/GetIndicatorDataWithQuintileBenchmarkComparison.sql
HowardMayBjss: With regards to testing our implementation results by comparison with the existing site, this has proved to be difficult as there is limited quintile output from the current site for the indicators used in POC. Comparison has been performed against these and this has lead to some ongoing discussions about the expected behaviour. It should also be noted that we are expecting differences in the Quintile calculation because of how the NTILE function in SQL Server handles groups of areas not divisible by 5 compared to the existing site. These differences have been discussed and agreed.

andyevans-bjss (approved):

PR #353, opened 2025-03-27

dlzhry2:
# Description

## Changes
Next Backend changes:
- Updated the mockIndicator.json file so the mock search service will have the latest trend data.

Frontend changes:

Trend Analysis:
- Very minor change added to update the indexes in the config data. I thin I had made a mistake in a recent PR that updated the keys. I also reran the trend analysis to ensure it worked as expected, and now the generated indicators.json file looks correct again.

## Validation

- Regenerated screenshots and compared before an after.

image


frontend/fingertips-frontend/components/pages/results/index.tsx
```diff
@@ -106,6 +108,10 @@ export function SearchResults({
               key={JSON.stringify(searchState)}
               searchResults={searchResults}
               searchState={searchState}
+              showTrends={
+                selectedAreasData?.length === 1 ||
+                (selectedAreasData?.length === 0 && isEnglandSelectedAsGroup)
```
dlzhry2, line 113: Because the trend analysis calculates the trend for every single area code for each indicator, we could conceivably change the logic to be if no areas are selected but a group has been e.g. all combined authorities within a parent, then we could show the trend for indicators in that parent location.

By the letter of the ticket it's just England. Is this worth raising to designers so late in the day or one for the back pocket?

frontend/fingertips-frontend/lib/search/indicatorSearchService.ts
```diff
@@ -41,32 +46,42 @@ export class IndicatorSearchService implements IIndicatorSearchService {
...
+    );
   }
 
   async getIndicator(
     indicatorId: string
   ): Promise<IndicatorDocument | undefined> {
-    const indicatorDocument = await this.searchClient
-      .getDocument(indicatorId)
-      .catch(() => {
-        return undefined;
```
dlzhry2, line 67: Catching an exception in this way does not cause the function to return early. Therefore modified to follow what I think the intent was.

frontend/fingertips-frontend/playwright/page-objects/pages/resultsPage.ts
```diff
@@ -49,6 +50,17 @@ export default class ResultsPage extends AreaFilter {
     ).toBeVisible();
   }
 
+  async checkRecentTrends(areaMode: AreaMode) {
+    if (areaMode !== AreaMode.TWO_PLUS_AREAS) {
+      // Raise comment in PR and agree e2e test approach
```
dlzhry2, line 55: One to potentially discuss @williamwillis11. This is quite a lightweight check where I am verifying: if 1 area selected or no/all areas selected with England as the group, then Recent trends is show as a heading above the results.

However, I am not doing any testing to check the actual trend label assigned. We could do this but I think this would be quite a bit of work i.e. we would need to have some config data to say what we are expecting the trend to be for each of the indicators.

It also could be a bit brittle, as we'd need to keep the mock search data up to date with what gets loaded in live site, as this test suite gets run against both local docker and dev.

Finally, there is a further complication with the isolate UI tests. Because it uses the local mocks, the mocked search needs to return particular data. I found when I replaced the mock search indicator data with the latest file, those tests fail because it expects certain data for the scenarios. I can explain in more detail if required and I'll also flag the issues in the channel as Stuart is looking to get rid of the mock. But hopefully this is sufficient for now.

dlzhry2: https://bjss.slack.com/archives/C08335HN952/p1743156028656909 Had a good discussion in thread. Think it would be sensible to not introduce these assertions until if/when we throw out the mock AI search for everything except the isolated e2e tests.

dlzhry2:
Closing as something weird happening with the Github Actions stuff.

PR #356, opened 2025-03-28

dlzhry2:
# Description

## Changes
Next Backend changes:
- Updated the mockIndicator.json file so the mock search service will have the latest trend data.

Frontend changes:

Trend Analysis:
- Very minor change added to update the indexes in the config data. I thin I had made a mistake in a recent PR that updated the keys. I also reran the trend analysis to ensure it worked as expected, and now the generated indicators.json file looks correct again.

## Validation

- Regenerated screenshots and compared before an after.

image


dlzhry2:
Something is happening here, but you're not quite sure what it is.

PR #358, opened 2025-03-28

dlzhry2:
# Description

## Changes
Next Backend changes:
- Updated the mockIndicator.json file so the mock search service will have a value for the latestTrend data, although this defaults to empty. As discussed in the comment on local mock e2e testing; it's quite brittle so replacing with real data caused issues.

Frontend changes:

Trend Analysis:
- Very minor change to bump the timeout for the queries. Occasionally still seeing this timeout in the pipeline job, and this can have implications for the indicators.json file.

## Validation

- Regenerated screenshots and compared before and after.

image


frontend/fingertips-frontend/components/pages/results/index.tsx
```diff
@@ -106,6 +108,10 @@ export function SearchResults({
               key={JSON.stringify(searchState)}
               searchResults={searchResults}
               searchState={searchState}
+              showTrends={
+                selectedAreasData?.length === 1 ||
+                (selectedAreasData?.length === 0 && isEnglandSelectedAsGroup)
```
dlzhry2, line 113: Because the trend analysis calculates the trend for every single area code for each indicator, we could conceivably change the logic to be if no areas are selected but a group has been e.g. all combined authorities within a parent, then we could show the trend for indicators in that parent location.

By the letter of the ticket it's just England. Is this worth raising to designers so late in the day or one for the back pocket?

frontend/fingertips-frontend/app/results/page.tsx
```diff
@@ -13,6 +13,7 @@ import {
   API_CACHE_CONFIG,
   ApiClientFactory,
 } from '@/lib/apiClient/apiClientFactory';
+import { areaCodeForEngland } from '@/lib/chartHelpers/constants';
```
dlzhry2, line 16: Should I move this into a common constants area? I'm finding this useful in places beyond charts. I initially did not want to move because it's plugged in everywhere so will massively up my files changes which can be off putting.

But it's a trivial change so very happy to action.

frontend/fingertips-frontend/lib/search/indicatorSearchService.ts
```diff
@@ -41,32 +46,42 @@ export class IndicatorSearchService implements IIndicatorSearchService {
...
+      isEnglandSelectedAsGroup
+    );
   }
 
   async getIndicator(
     indicatorId: string
   ): Promise<IndicatorDocument | undefined> {
-    const indicatorDocument = await this.searchClient
-      .getDocument(indicatorId)
-      .catch(() => {
```
dlzhry2, line 66: Catching an exception in this way does not cause the function to return early. Therefore modified to follow what I think the intent was.

frontend/fingertips-frontend/playwright/page-objects/pages/resultsPage.ts
```diff
@@ -49,6 +50,17 @@ export default class ResultsPage extends AreaFilter {
     ).toBeVisible();
   }
 
+  async checkRecentTrends(areaMode: AreaMode) {
+    if (areaMode !== AreaMode.TWO_PLUS_AREAS) {
+      // Raise comment in PR and agree e2e test approach
```
dlzhry2, line 55: Raised comment earlier here (on the PR I closed due to some initial issues with Docker). One for a quick discussion. Will remove comment once we're happy: https://github.com/dhsc-govuk/FingertipsNext/pull/353#discussion_r2018260130

frontend/fingertips-frontend/lib/search/searchTypes.ts
```diff
@@ -3,13 +3,21 @@
...
 export const AREA_TYPE_GP = 'GPs';
 
-export type IndicatorDocument = {
+export interface AreaCodeWithTrend {
+  areaCode: string;
+  trend: HealthDataPointTrendEnum;
+}
+
+// Indicator data as returned by the search backend
+export type RawIndicatorDocument = {
```
williamwillis11, line 20: there are hundreds of imports of IndicatorDocument across the codebase, i was just wondering if you have reviewed each to see if it is correct to continue using that type or should now use this new Raw IndicatorDocument?

frontend/fingertips-frontend/playwright/page-objects/pages/resultsPage.ts
dlzhry2: Discussed - closing.

frontend/fingertips-frontend/lib/search/searchTypes.ts
dlzhry2: Discussed on call - both happy with outcome and the refactoring.

frontend/fingertips-frontend/playwright/page-objects/pages/resultsPage.ts
```diff
@@ -52,6 +53,13 @@ export default class ResultsPage extends AreaFilter {
     ).toBeVisible();
   }
 
+  async checkRecentTrends(areaMode: AreaMode) {
+    const trendsShouldBeVisible = areaMode !== AreaMode.TWO_PLUS_AREAS;
```
williamwillis11, line 57: nice :)

frontend/fingertips-frontend/app/results/page.tsx
alibombaye: Yeah agree with this. Maybe worth doing as a quick little follow up tech-debt PR to move this to a common place.

frontend/fingertips-frontend/components/forms/IndicatorSelectionForm/index.tsx
```diff
@@ -138,14 +152,18 @@ export function IndicatorSelectionForm({
...
+            <Checkbox
+              key={`select-all-indicator-${areAllIndicatorsSelected}`}
+              data-testid="select-all-checkbox"
+              defaultChecked={areAllIndicatorsSelected}
+              onChange={(e) => handleSelectAll(e.target.checked)}
+            >
+              Select all
+            </Checkbox>
+
+            {showTrends ? 'Recent trend for selected area' : null}
```
alibombaye, line 165: What we can do here is use the `Paragraph` component for this. That would remove the need for certain Font stying in the `ResultLabelsContainer` that we should then be getting for free.

frontend/fingertips-frontend/lib/search/indicatorMapper.test.ts
```diff
@@ -0,0 +1,243 @@
...
+          dataSource: 'Mega City 1',
+          lastUpdatedDate: new Date('November 5, 2023'),
+          trend: 'Decreasing',
+          unitLabel: '',
+          hasInequalities: true,
+        },
+      ]);
+    });
+
+    it('maps trend to undefined when a user has requested more an indicator in more than one area', () => {
```
alibombaye, line 91: change `maps trend to undefined when a user has requested more an indicator in more than one area` -> `maps trend to undefined when a user has requested an indicator in more than one area`?

frontend/fingertips-frontend/lib/search/indicatorMapper.test.ts
```diff
@@ -0,0 +1,243 @@
...
+        {
+          indicatorID: '1',
+          indicatorName: 'Red faced',
+          indicatorDefinition:
+            'Count of people who did something they are embarrassed by',
+          earliestDataPeriod: '1938',
+          latestDataPeriod: '2023',
+          dataSource: 'The Beano',
+          lastUpdatedDate: new Date('December 6, 2024'),
+          trend: undefined,
```
alibombaye, line 135: I'm a little confused by this. The test above for indicator 2 has more than one area but doesn't return undefined there. What makes that behave differently?

alibombaye: ah I see it's line 121 where we are requesting 2 areas.

frontend/fingertips-frontend/app/results/page.tsx
dlzhry2: Yup, good shout. Will create ticket after merging.

frontend/fingertips-frontend/lib/search/indicatorMapper.test.ts
```diff
@@ -0,0 +1,243 @@
...
+            {
+              areaCode: 'E92000001',
+              trend: HealthDataPointTrendEnum.DecreasingAndGettingBetter,
+            },
+          ],
+          unitLabel: '',
+          hasInequalities: false,
+          usedInPoc: true,
+        },
+      ];
```
alibombaye, line 222: what might be nice is we had a base mock consts for Indicator 1 and Indicator 2.
Then in the tests we spread the base and change the things are different for the tests. That will remove some duplication but also highlights as part of the test the bits of the object that affects the outcome of the test.

Can do the same with the expected results too.

frontend/fingertips-frontend/lib/search/indicatorMapper.ts
```diff
@@ -0,0 +1,79 @@
...
+    const areaCodeToMatch =
+      isEnglandSelectedAsGroup && requestedAreaCodes.length === 0
+        ? areaCodeForEngland
+        : requestedAreaCodes[0];
+
+    const matchedElement = trendsWithAreas.find(
+      (trendWithArea) => trendWithArea.areaCode === areaCodeToMatch
+    );
+
+    return !matchedElement ? undefined : matchedElement.trend;
```
alibombaye, line 54: Can we flip change this to 
`return matchedElement ? matchedElement.trend : undefined;`
Should make it little easier to read.

frontend/fingertips-frontend/lib/search/indicatorMapper.ts
```diff
@@ -0,0 +1,79 @@
...
+        rawIndicator.trendsByArea,
+        isEnglandSelectedAsGroup
+      ),
+      dataSource: rawIndicator.dataSource,
+      earliestDataPeriod: rawIndicator.earliestDataPeriod,
+      latestDataPeriod: rawIndicator.latestDataPeriod,
+      lastUpdatedDate: rawIndicator.lastUpdatedDate,
+      hasInequalities: rawIndicator.hasInequalities,
+      unitLabel: rawIndicator.unitLabel,
+    } as IndicatorDocument;
```
alibombaye, line 77: I am wondering if we can do this;
```
return {
  ... rawIndicator,
  trend: this.getTrendForIndicatorInArea(
        requestedAreaCodes,
        rawIndicator.trendsByArea,
        isEnglandSelectedAsGroup
      ),
}
```

frontend/fingertips-frontend/lib/search/indicatorMapper.test.ts
dlzhry2: Sounds good, will do that.

frontend/fingertips-frontend/lib/search/indicatorMapper.ts
dlzhry2: Could do, but we'd have to do a delete afterwards as the mapper removes (does not map) 2 very large arrays which appear in the RawIndicatorDocument: `associatedAreaCodes` and `trendsByArea`.

frontend/fingertips-frontend/lib/search/indicatorSearchServiceMock.test.ts
```diff
@@ -26,66 +37,130 @@ describe('IndicatorSearchServiceMock', () => {
...
   ];
+  let mockMappedData: IndicatorDocument[];
   let indicatorSearchMock: IIndicatorSearchService;
+
   beforeAll(() => {
-    indicatorSearchMock = new IndicatorSearchServiceMock(mockData);
+    indicatorSearchMock = new IndicatorSearchServiceMock(mockRawData);
+  });
+
+  beforeEach(() => {
```
alibombaye, line 62: not sure why we need the beforeEach here. Doesn't look like we change mockMappedData in any of the tests.

frontend/fingertips-frontend/components/molecules/Result/index.tsx
```diff
@@ -109,47 +118,59 @@ export function SearchResult({
...
+                    <GreyTag data-testid="tag-has-inequalities">
+                      Contains inequality data
+                    </GreyTag>
+                  </TagsColumn>
+                ) : (
+                  <></>
+                )}
+              </TagRow>
+            </Checkbox>
+            {showTrends ? (
```
dlzhry2, line 166: Had feedback from the designers that they would like the width of the TrendTag to be consistent no matter the length of the link text. I've tried a couple of things and can make it a little better but not perfect. Any idea on how best to do this styling?

frontend/fingertips-frontend/lib/search/indicatorSearchServiceMock.test.ts
```diff
@@ -26,66 +37,130 @@ describe('IndicatorSearchServiceMock', () => {
...
+      ).toEqual(mockMappedData);
     });
 
     it('should filter out any not in Area2', async () => {
-      expect(await indicatorSearchMock.searchWith('Red', ['Area2'])).toEqual([
-        mockData[1],
-      ]);
+      const mockResult = mockMappedData[1];
+      // This is the trend for this indicator in Area2
+      mockResult.trend = 'No significant change';
```
alibombaye, line 125:
```
cont expectedResult = {
  ...mockMappedData[1],
  trend: 'No significant change'
}
```

frontend/fingertips-frontend/lib/search/searchTypes.ts
alibombaye, line 20: Don't think the comment is needed here.

frontend/fingertips-frontend/components/pages/results/index.tsx
```diff
@@ -43,6 +44,7 @@ export function SearchResults({
   searchResults,
   areaFilterData,
   selectedAreasData,
+  isEnglandSelectedAsGroup,
```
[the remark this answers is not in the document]
dlzhry2: It's a fair point. I was thinking to pass it down, as we need to calculate it in `frontend/fingertips-frontend/app/results/page.tsx` since it is a parameter to pass to the SearchService. Since we have it there, then we can just pass it down into the component. Appreciate it gets passed down a couple of layers, but was thinking it might be best to just work it out once.

frontend/fingertips-frontend/lib/search/indicatorMapper.ts
alibombaye: I think the casting may remove those for us anyway.

frontend/fingertips-frontend/components/molecules/Result/index.tsx
alibombaye: Will catch up with you about this.

frontend/fingertips-frontend/lib/search/indicatorMapper.ts
dlzhry2: Gave it a test, unfortunately the additional keys from the raw object are included when using ... destructuring. Only other approach would be to do this and then delete the keys, but I prefer specifically calling out the keys we are mapping.

frontend/fingertips-frontend/lib/search/indicatorSearchServiceMock.test.ts
dlzhry2: Removed

frontend/fingertips-frontend/lib/search/indicatorSearchServiceMock.test.ts
dlzhry2: Refactored - please check.

frontend/fingertips-frontend/lib/search/indicatorSearchService.test.ts
```diff
@@ -143,27 +186,48 @@ describe('IndicatorSearchService', () => {
...
+              trend: HealthDataPointTrendEnum.Decreasing,
+            },
+            {
+              areaCode: 'Area2',
+              trend: HealthDataPointTrendEnum.NoSignificantChange,
+            },
+          ],
+          unitLabel: '',
+          hasInequalities: true,
+          usedInPoc: true,
```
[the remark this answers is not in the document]
dlzhry2: Thanks, this was useful. I've used where appropriate now within my PR.

frontend/fingertips-frontend/lib/search/indicatorMapper.ts
dlzhry2: Yes, this has helped to massively reduce the boilerplate and duplicate test code. Please check you're happy with resolution.

frontend/fingertips-frontend/lib/search/indicatorMapper.test.ts
dlzhry2: Completed - please check you're happy with resolution.

steves-bjss (approved):

frontend/fingertips-frontend/lib/search/indicatorMapper.test.ts
alibombaye: Ah nice. I'm liking that 👍

alibombaye (approved):

PR #363, opened 2025-03-28
DHSCFT-541 - Add navigation to indicators page to e2e test

Nayira-BJSS:
# Description

Jira ticket: [DHSCFT-541](https://bjss-enterprise.atlassian.net/browse/DHSCFT-541)

Added step to existing test to navigate to the selected indicator background information page, verify the title displayed and return back to the results page

## Changes

- Added new step to existing e2e test
- Added helper to get indicator name from provided indicator id
- Added new methods to indicator and results page
- Updated index files to include new data-testid for elements

## Validation

Step added to e2e


frontend/fingertips-frontend/playwright/page-objects/pages/indicatorPage.ts
```diff
@@ -1,7 +1,35 @@
...
+
   public async navigateToIndicatorPage(indicatorId: string) {
     await this.navigateTo(`/indicator/${indicatorId}`);
   }
+
+  async checkIndicatorNameTitle(
+    indicatorId: string,
+    typedIndicatorData: IndicatorDocument[]
+  ) {
+    const indicatorName = getIndicatorNameById(indicatorId, typedIndicatorData);
```
williamwillis11, line 18: as discussed i think we should get a fuller indicator object back from the call done in the e2e test file itself and pass that single object around, rather than passing around the id then re-searching for different things for that ID

but also as discussed you said you would revisit that once this was in working so just posting for visibility :)

frontend/fingertips-frontend/playwright/page-objects/pages/resultsPage.ts
```diff
@@ -245,4 +248,21 @@ export default class ResultsPage extends AreaFilter {
       new RegExp(`&is=${deselectedIndicator}`)
     );
   }
+
+  async clickViewBackgroundInformationLinkForIndicator(
+    indicatorId: string,
+    typedIndicatorData: IndicatorDocument[]
+  ) {
+    const indicatorName = getIndicatorNameById(indicatorId, typedIndicatorData);
```
williamwillis11, line 256: as discussed i think we should get a fuller indicator object back from the call done in the e2e test file itself and pass that single object around, rather than passing around the id then re-searching for different things for that ID

but also as discussed you said you would revisit that once this was in working so just posting for visibility :)

PR #366, opened 2025-03-31
DHSCFT-520: Display Deprivation Data

gareth-allan-bjss:
# Description

Jira ticket: [DHSCFT-520](https://bjss-enterprise.atlassian.net/browse/DHSCFT-520)

This PR adds the capability to display deprivation inequality data as well as sex. Currently this is done by manually specifying a query parameter to show deprivation instead of sex. Future tickets ([DHSCFT-224](https://bjss-enterprise.atlassian.net/browse/DHSCFT-224) & [DHSCFT-382](https://bjss-enterprise.atlassian.net/browse/DHSCFT-382)) will add a dropdown above the chart to allow you to select which type of inequality to show.

## Changes

- Updated the API to return the `sex` and `ageBand` fields of the health data response as objects, so that an `isAggregate` flag can be included for each field. This allows us to filter data in the frontend to retrieve the data points that are disaggregated for only the inequality we want to show
- Updated the API to include an `isAggregate` flag in the `deprivation` object (previously added in #318)
- Updated the health data API call in the `OneIndicatorOneAreaView` to request data broken down by sex and deprivation
- Updated the frontend `Inequalities` component to filter the health data retrieved depending on the inequality type being shown. This allows us to get only the data for sex, or only the data for deprivation, in order to show it on the inequalities line chart/bar chart.
- Updated several function in `inequalitiesHelpers.ts` to handle deprivation data as well as sex data. This was accomplished by passing in function to select the relevant fields for data display or sorting, rather than adding specific logic for each case. The idea is that to support future inequalities you can add new selector functions to the appropriate maps, rather than having to update each individual function to support them.
- Updated the `SearchParams` utility to add a query parameter for the inequality type selected. This is currently used to allow you to display deprivation inequalities, but will be extended to allow a dropdown to be used in future.
- Updated the sequence value of the `All` deprivation dimension entry to be `10`, rather than `1`, as we show these values in descending order.
- Updated lots of frontend test data to reflect the API changes

## Validation

Ran the .http file tests to verify they pass

Re-ran the "update screenshots" command to verify the existing chart displays were not changed.

### Sex
Verified that "one indicator, one area" view has not changed between this branch and main when showing the sex inequality for an indicator with both sex and deprivation data.

Line chart (main):
Screenshot 2025-03-31 at 12 18 24

Line chart (this PR):
Screenshot 2025-03-31 at 12 18 45

Inequalities bar chart (main):
Screenshot 2025-03-31 at 12 19 29

Inequalities bar chart (this PR):
Screenshot 2025-03-31 at 12 20 02

Inequalities line chart (main):
Screenshot 2025-03-31 at 12 20 30

Inequalities line chart (this PR):
Screenshot 2025-03-31 at 12 20 47

### Deprivation
Verified that the deprivation data is displayed when the appropriate query parameter is passed in the URL (**note there are some display issues that will be resolved in a subsequent PR**).
Screenshot 2025-03-31 at 12 23 04

Screenshot 2025-03-31 at 12 23 25

Screenshot 2025-03-31 at 12 23 33

Screenshot 2025-03-31 at 12 23 43


api/definition/swagger.yaml
```diff
@@ -481,6 +478,40 @@ components:
...
+          description: Indicates if the datapoint is an aggregated value for the sex dimension.
+    AgeData:
+      type: object
+      description: Details of the age band that this data point belongs to.
+      required:
+        - value
+        - isAggregate
+      properties:
+        value:
+          type: string
```
HowardMayBjss, line 508: Isn't this an enumerated type?

api/definition/swagger.yaml
```diff
@@ -481,6 +478,40 @@ components:
...
+          description: Indicates if the datapoint is an aggregated value for the deprivation dimension.
+    SexData:
+      type: object
+      description: Details of the sex category that this data point belongs to.
+      required:
+        - value
+        - isAggregate
+      properties:
+        value:
+          type: string
```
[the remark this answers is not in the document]
gareth-allan-bjss: Essentially, I've kept the types for the `value` fields of the new sex and age object the same as the existing fields.

For sex, I'm wary of defining a strict list of acceptable sexes. I also think it allows us to drive this based on the data that is uploaded.

api/definition/swagger.yaml
gareth-allan-bjss: I don't think so, as I think different indicators can have different age bands, and I don't think we actually want to limit the API specification to say what these can be.

HowardMayBjss (approved):

PR #369, opened 2025-03-31
Feature/dhscft 292 scroll table with sticky columns

stuc-bjss:
# Description

Jira ticket: [DHSCFT-292](https://bjss-enterprise.atlassian.net/browse/DHSCFT-292)

Enable horizontal scrolling on the line chart table and make the first and last columns sticky
Remove braces around the % measurement unit


gareth-allan-bjss:
Looks good, but one small issue I think.

It looks like the vertical line between the headers for the two areas (when two areas are selected) has disappeared as part of this change:
Screenshot 2025-04-01 at 12 07 44
Screenshot 2025-04-01 at 12 07 50

I know you've got further work waiting for this, so I'm happy if you want to address it as part of that.

gareth-allan-bjss (approved):

PR #372, opened 2025-03-31
Feature/dhscft 542 api inconsistant behaviour

steves-bjss:
# Description

Jira ticket: [DHSCFT-542](https://bjss-enterprise.atlassian.net/browse/DHSCFT-542)

Fix inconsistency of return statuses.

## Changes

- Returns a 200 response if no data is found for a particular indicator but the indicator exists.
- Returns a 404 if information is requested about an unknown indicator.

## Validation

Unit tested and E2E tests executed.  Manually checked.


api/DHSC.FingertipsNext.Modules.HealthData.Service/IndicatorService.cs
```diff
@@ -38,36 +38,42 @@ public class IndicatorService(IHealthDataRepository healthDataRepository, IMappe
...
+
+        var areaHealthData = ((await GetIndicatorAreaDataAsync(
+            indicatorId,
+            areaCodes,
+            areaType,
+            years,
+            inequalities,
+            method,
+            polarity
+        )) ?? []).ToList();
```
dlzhry2, line 62: Does `GetIndicatorAreaDataAsync` return null and require this check? I thought it will return an IEnumerable which could have HealthData or could be empty.

api/DHSC.FingertipsNext.Modules.HealthData.Service/ServiceResponse.cs
```diff
@@ -0,0 +1,16 @@
+namespace DHSC.FingertipsNext.Modules.HealthData.Service;
```
dlzhry2, line 1: Would we want to make something like this common for all services e.g. AreaService could use this too?

api/DHSC.FingertipsNext.Modules.HealthData.UnitTests/Services/IndicatorServiceTests.cs
```diff
@@ -266,15 +266,15 @@ public async Task GetIndicatorData_ShouldReturnExpectedResult_BenchmarkingInequa
...
+        dataResults.ShouldNotBeEmpty();
+        dataResults.Count().ShouldBe(2);
 
-        var areaResults = areaDataResults[1];
+        var areaResults = dataResults.ElementAt(1);
         areaResults.AreaCode.ShouldBeEquivalentTo(expectedAreaCode);
         areaResults.AreaName.ShouldBeEquivalentTo(expectedAreaName);
         areaResults.HealthData.Count().ShouldBe(6);
-
+
```
dlzhry2, line 277:
```suggestion

```

dlzhry2 (approved):
Looks great, just a few questions and minor comments.

api/DHSC.FingertipsNext.Modules.HealthData.Service/IndicatorService.cs
steves-bjss: It is valid to return null and thats fine with IEnumerable but this was put in to specifically return an empty array rather than null.

PR #382, opened 2025-04-02
Feature/dhscft 528 thematic map to show quntiles

WolfieKnee:
# Description

Jira ticket: [DHSCFT-528](https://bjss-enterprise.atlassian.net/browse/DHSCFT-528)/[DHSCFT-228](https://bjss-enterprise.atlassian.net/browse/DHSCFT-228)

Update ThematicMap to show Quintles benchmarks

## Changes
- ThematicMaps & tests: update to use searchState context and interact correctly with loading spinner
- thematicMapHelpers: define highcharts colour mapping for quintiles
- viewPlot: update to use searchState context
- view: update to use searchState context and incude areaType in API call and updated max API response size. 

## Validation
Apperance as expected against local api:
image

e2e screenshots unchanged


frontend/fingertips-frontend/components/organisms/ThematicMap/index.tsx
```diff
@@ -46,22 +49,32 @@ export function ThematicMap({
   // the lint directive doesn't really apply here, and having either no dependency array, or mapOptions as a dependency
   // causes it to loop infinitely. (https://react.dev/reference/react/useEffect#examples-dependencies)
```
stuc-bjss, line 50: redundant comment now that what appear to be all the dependencies have been added?

frontend/fingertips-frontend/components/organisms/ThematicMap/ThematicMap.test.tsx
```diff
@@ -2,41 +2,59 @@ import { render, screen } from '@testing-library/react';
...
+    render(
+      <ThematicMap
+        healthIndicatorData={mockHealthData['92420']}
+        mapGeographyData={mockMapGeographyData}
+        benchmarkComparisonMethod={'Unknown'}
+        polarity={'Unknown'}
+      />
+    );
+
+    const highcharts = await screen.findByTestId(
```
williamwillis11, line 40: is this the best name for this const? 

its one chart right, and i means that the assertion below reads like we are expecting multiple unspecified highcharts to appear when really its just the thematicMap-component

would:

 thematicMapHighchartsComponent
 
 make more sense?

frontend/fingertips-frontend/components/organisms/ThematicMap/ThematicMap.test.tsx
```diff
@@ -2,41 +2,59 @@ import { render, screen } from '@testing-library/react';
...
+        mapGeographyData={mockMapGeographyData}
+        benchmarkComparisonMethod={'Unknown'}
+        polarity={'Unknown'}
+        measurementUnit={''}
+      />
+    );
 
-  const highcharts = await screen.findByTestId('benchmarkLegend-component');
-  expect(highcharts).toBeInTheDocument();
+    const highcharts = await screen.findByTestId('benchmarkLegend-component');
```
williamwillis11, line 57: same comment as above

frontend/fingertips-frontend/components/organisms/ThematicMap/ThematicMap.test.tsx
stuc-bjss: const name seems ok to me, more importantly does this test actually offer any useful info at all, I know we do it in lots of other tests too, but does this one provide anything not already covered by the following tests?

WolfieKnee: I agree with @stuc-bjss - this test is not adding value here, and is more useful (and already in) the ViewPlot tests.

williamwillis11: fair play!

PR #386, opened 2025-04-02
Feature/dhscft 572 white bars on chart when no benchmark comparison

stuc-bjss:
# Description

Jira ticket: [DHSCFT-572](https://bjss-enterprise.atlassian.net/browse/DHSCFT-572)

When benchmarking data is missing, then the bar chart colors should be white with black border

## Changes

- Bar chart colours
- Change hover on sparkline to useHtml
- Changed dot from symbol to span with height, width, rounded edges and border


dlzhry2 (approved):
Looks good to me. Could be good if you could look @priya-thorpe-bjss too if this is touching on a similar area to what you're working?

gareth-allan-bjss (commented):
Looks good, but I think you'll need to update the screenshots, as there are some that cover the compare areas table.

PR #387, opened 2025-04-02
Feature/dhscft 237 two or more indicators england table

priya-thorpe-bjss:
# Description

Jira ticket: [DHSCFT-237](https://bjss-enterprise.atlassian.net/browse/DHSCFT-237)

## Changes
- Added two or more Indicators England view and view plots
- Created a table to show selected indicators and related data including trends
- If there is no value for the selected indicator, an X is displayed in the cell

## Validation
Tests pass, manual validation

Screenshot 2025-04-04 at 16 14 42

I have updated the snap shots.


frontend/fingertips-frontend/components/organisms/EnglandAreaTypeTable/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+        }
+      >
+        {indicatorData.map((item) => (
+          <Table.Row key={item.indicatorId}>
+            <CheckValueInTableCell value={item?.indicatorName} />
+            <CheckValueInTableCell
+              value={item?.period}
+              style={{ textAlign: 'center' }}
+            />
+            <CheckValueInTableCell
```
MonkeyDate, line 92: if you pull from main there should be a FormatNumberInTableCell component you can use

<FormatNumberInTabelCell numberStyle='whole'

frontend/fingertips-frontend/components/organisms/EnglandAreaTypeTable/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+              value={item?.latestEnglandHealthData?.count}
+              style={{ textAlign: 'right' }}
+            />
+            <CheckValueInTableCell
+              value={item?.unitLabel}
+              style={{ textAlign: 'right' }}
+            />
+            <CheckValueInTableCell
```
MonkeyDate, line 100: <FormatNumberInTabelCell

MonkeyDate (changes requested):

frontend/fingertips-frontend/playwright/testHelpers.ts
```diff
@@ -218,14 +225,13 @@ export function getScenarioConfig(
...
-    //     // ChartPage.basicTableComponent,
-    //     // Enable in DHSCFT-225
-    //     // ChartPage.populationPyramidComponent,
-    //   ].includes(component.componentLocator)
-    // );
+    visibleComponents = allComponents.filter((component) =>
+      [
+        ChartPage.englandAreaTypeTableComponent,
+        // Enable in DHSCFT-225
+        // ChartPage.populationPyramidComponent,
```
obarojohnson, line 232: Do you really need this comments?

frontend/fingertips-frontend/lib/chartHelpers/extractingIndicatorHealthData.ts
```diff
@@ -0,0 +1,52 @@
...
+    orderedEnglandData[index] = englandData;
+
+    orderedMetadata[index] = unsortedMetaData?.find(
+      (indicatorMetaData) =>
+        Number(indicatorMetaData?.indicatorID) === indicator.indicatorId
+    );
+  });
+
+  console.log(orderedEnglandData);
+  console.log(orderedMetadata);
```
obarojohnson, line 46: I think you might have forgetton your debugging log or is this intentional?

frontend/fingertips-frontend/components/viewPlots/ViewPlotProps.ts
```diff
@@ -17,3 +17,9 @@ export type TwoOrMoreIndicatorsViewPlotProps = {
   indicatorMetadata: IndicatorDocument[];
   benchmarkStatistics: QuartileData[];
 };
+
+export type TwoOrMoreIndicatorsEnglandViewPlotProps = {
```
obarojohnson, line 21: I am not sure if this is the right home for this TwoOrMoreIndicatorsEnglandViewPlotProps type , I think it should be on the same file with its component without exporting it.

frontend/fingertips-frontend/components/viewPlots/TwoOrMoreIndicatorsEnglandViewPlots/index.tsx
```diff
@@ -1,7 +1,69 @@
 'use client';
 
-import { H2 } from 'govuk-react';
+import { TwoOrMoreIndicatorsEnglandViewPlotProps } from '@/components/viewPlots/ViewPlotProps';
```
obarojohnson, line 3: I think you can move this TwoOrMoreIndicatorsEnglandViewPlotProps to this page so you don't have to export it.

frontend/fingertips-frontend/components/organisms/EnglandAreaTypeTable/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+export interface EnglandAreaTypeIndicatorData {
+  indicatorId: number | undefined;
+  indicatorName: string | undefined;
+  period: string | undefined;
+  latestEnglandHealthData: HealthDataPoint | undefined;
+  unitLabel: string | undefined;
+}
+
+export interface EnglandAreaTypeTableProps {
```
obarojohnson, line 33: I don't think you need to export this EnglandAreaTypeTableProps

obarojohnson (changes requested):

frontend/fingertips-frontend/components/organisms/EnglandAreaTypeTable/index.tsx
priya-thorpe-bjss: Thanks, have updated

frontend/fingertips-frontend/lib/chartHelpers/extractingIndicatorHealthData.ts
priya-thorpe-bjss: Thanks, removed file as its not needed

frontend/fingertips-frontend/playwright/testHelpers.ts
priya-thorpe-bjss: Its to keep track of what components are going in and what ticket they will be enabled in.  The person who is adding that component will replace the comment with the component in the test file.

frontend/fingertips-frontend/components/viewPlots/TwoOrMoreIndicatorsEnglandViewPlots/index.tsx
priya-thorpe-bjss: Ive left it in the viewPlotProps file then all the view plots props are in once place.

frontend/fingertips-frontend/components/organisms/EnglandAreaTypeTable/index.tsx
priya-thorpe-bjss: removed the export

frontend/fingertips-frontend/components/viewPlots/TwoOrMoreIndicatorsEnglandViewPlots/index.tsx
obarojohnson: Interesting. I think it's wrong then. It's only an idea if the props will be shared among other components. If not, I think it can be moved to its component file. I'm sure the initial idea was to share **ViewPlotProps** as the base props for other views, not the concrete component props.

frontend/fingertips-frontend/components/organisms/EnglandAreaTypeTable/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+              numberStyle={'whole'}
+              style={{ textAlign: 'right' }}
+            />
+            <CheckValueInTableCell
+              value={item?.unitLabel}
+              style={{ textAlign: 'right' }}
+            />
+            <FormatNumberInTableCell
+              value={item?.latestEnglandHealthData?.value}
+              numberStyle={'whole'}
```
MonkeyDate, line 106: numberStyle=whole is really for whole numbers, like `count`.

frontend/fingertips-frontend/playwright/testHelpers.ts
gareth-allan-bjss: DHSCFT-225 is marked as done though, so we need to make sure there's a task to track doing this.

frontend/fingertips-frontend/components/organisms/EnglandAreaTypeTable/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+  period: string | undefined;
+  latestEnglandHealthData: HealthDataPoint | undefined;
+  unitLabel: string | undefined;
+}
+
+interface EnglandAreaTypeTableProps {
+  indicatorData: EnglandAreaTypeIndicatorData[];
+}
+
+export function EnglandAreaTypeTable({
```
gareth-allan-bjss, line 40: I don't think this component needs to reference the fact that it's for England specifically. It's just a table that can show multiple indicators for one area, and the only thing England-specific about it as far as I can see is that it puts the text "England" at the top.

Can we name it (and its prop types, etc.) something more generic (like perhaps `OneAreaMultipleIndicatorsTable`, or whatever makes sense) and have it take the area name as a prop, instead of hardcoding it to `England`?

frontend/fingertips-frontend/.test/spec/snaps/chromium/fully_integrated_e2e_tests/core_journeys.spec.ts/BOTH-SUBJECT-AND-AREA-then-select-ONE-INDICATOR-and-ONE-AREA-then-check-the-charts-page-inequalities-component.png
gareth-allan-bjss, line 1: Just for reference (and I've definitely been guilty of this up to this point), we've talked about trying to avoid committing changes to screenshots that aren't major changes or related to components that have been updated in the PR. We know the framework currently results in small changes to the screenshots each time they're updated, so if we can avoid committing these it helps us see what the real changes are without having to wade through a load of screenshots.

frontend/fingertips-frontend/components/organisms/EnglandAreaTypeTable/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+export enum EnglandAreaTypeHeaderEnum {
+  England = 'England',
+}
+
+export interface EnglandAreaTypeIndicatorData {
+  indicatorId: number | undefined;
+  indicatorName: string | undefined;
+  period: string | undefined;
+  latestEnglandHealthData: HealthDataPoint | undefined;
+  unitLabel: string | undefined;
```
gareth-allan-bjss, line 33: When would we expect any of these values (except possibly the `unitLabel`) to be undefined? I don't think it would be valid to pass a table row that's missing any of the other properties to this component, would it?

frontend/fingertips-frontend/components/organisms/EnglandAreaTypeTable/EnglandAreaTypeTable.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+    indicatorName: ' ',
+    period: '2008',
+    latestEnglandHealthData: mockEnglandHealthData.healthData[0],
+    unitLabel: '',
+  },
+  {
+    indicatorId: 2,
+    indicatorName: 'no data indicator',
+    period: '',
+    latestEnglandHealthData: mockNoHealthData.healthData[0],
```
gareth-allan-bjss, line 49: I think this could just be:
```suggestion
    latestEnglandHealthData: undefined,
```

Then you wouldn't need `mockNoHealthData`.

frontend/fingertips-frontend/components/organisms/EnglandAreaTypeTable/EnglandAreaTypeTable.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+      value: 890.305692,
+      lowerCi: 441.69151,
+      upperCi: 578.32766,
+      ageBand: allAgesAge,
+      sex: personsSex,
+      trend: HealthDataPointTrendEnum.NotYetCalculated,
+      deprivation: noDeprivation,
+    },
+  ],
+};
```
gareth-allan-bjss, line 29: You're not actually using any properties of this object except for `healthData`, so it seems redundant declaring it. Also, I'm aware that we have a lot of declarations of health data in our tests and it makes changing the structure of the API response a real pain. I added `healthDataPoint` to `frontend/fingertips-frontend/lib/mocks.ts` as an example `HealthDataPoint` to use if you don't need it to have specific properties (which I think is the case here). Could you use that instead please, so we don't introduce yet another instance of a `HealthDataPoint`?

frontend/fingertips-frontend/components/organisms/EnglandAreaTypeTable/EnglandAreaTypeTable.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+  {
+    indicatorId: 2,
+    indicatorName: 'no data indicator',
+    period: '',
+    latestEnglandHealthData: mockNoHealthData.healthData[0],
+    unitLabel: '',
+  },
+];
+
+describe('EnglandAreaTypeTable', () => {
```
gareth-allan-bjss, line 54: Is this component worth a snapshot test?

frontend/fingertips-frontend/components/views/TwoOrMoreIndicatorsEnglandView/index.tsx
```diff
@@ -4,31 +4,90 @@ import { SearchParams, SearchStateManager } from '@/lib/searchStateManager';
...
+    return healthIndicatorData;
+  };
+
+  const combinedIndicatorData = await Promise.all(
+    indicatorsSelected.map((indicator) => {
+      return getHealthDataForIndicator(indicator);
+    })
+  );
   console.log('TODO: fetch health data with inequalites');
   console.log(`TODO: fetch population data for areas: [${areaCodesToRequest}]`);
```
gareth-allan-bjss, line 79: These `console.log` statements presumably want to be removed.

frontend/fingertips-frontend/components/views/TwoOrMoreIndicatorsEnglandView/index.tsx
```diff
@@ -4,31 +4,90 @@ import { SearchParams, SearchStateManager } from '@/lib/searchStateManager';
...
+        .map((indicatorData) => indicatorData?.areaHealthData ?? [])
+        .flat();
+    } catch (error) {
+      throw new Error(
+        `error getting health indicator data for areas: ${error}`
+      );
+    }
+
+    return healthIndicatorData;
+  };
```
gareth-allan-bjss, line 71: `getHealthDataForIndicator` is exactly the same as the version in `frontend/fingertips-frontend/components/views/TwoOrMoreIndicatorsAreasView/index.tsx`. Can you extract it to a utility function please, and call it from both places?

frontend/fingertips-frontend/components/views/TwoOrMoreIndicatorsEnglandView/index.tsx
```diff
@@ -4,31 +4,90 @@ import { SearchParams, SearchStateManager } from '@/lib/searchStateManager';
...
 
   if (
-    areasSelected?.length !== 1 &&
-    areasSelected?.[0] !== areaCodeForEngland
+    !indicatorsSelected ||
+    indicatorsSelected?.length < 2 ||
+    (areasSelected?.length !== 1 && areasSelected?.[0] !== areaCodeForEngland)
   ) {
     throw new Error('Invalid parameters provided to view');
   }
```
gareth-allan-bjss, line 34: `frontend/fingertips-frontend/components/views/TwoOrMoreIndicatorsAreasView/index.tsx` has checks for the validity of `selectedIndicatorsData` that we don't have here. I think we should replicate them (and have a test to validate the expected error is shown if they aren't present).

frontend/fingertips-frontend/playwright/testHelpers.ts
priya-thorpe-bjss: There are a couple of comments for the pop pyramid in the scenarios for the test file, which Will has added. I think this is how he keeps a track of it.  @williamwillis11  should be able to shed some more light on this.

frontend/fingertips-frontend/components/views/TwoOrMoreIndicatorsEnglandView/TwoOrMoreIndicatorsEnglandView.test.tsx
```diff
@@ -0,0 +1,89 @@
...
+];
+
+const mockIndicatorsApi = mockDeep<IndicatorsApi>();
+ApiClientFactory.getIndicatorsApiClient = () => mockIndicatorsApi;
+
+describe('TwoOrMoreIndicatorsEnglandView', () => {
+  beforeEach(() => {
+    mockIndicatorsApi.getHealthDataForAnIndicator
+      .mockResolvedValueOnce(mockIndicator)
+      .mockResolvedValueOnce(mockIndicator);
```
gareth-allan-bjss, line 57: I think it would be better if the mock returned different indicator data for each call, so we can be sure we're actually passing through data for all selected indicators to the view plot.

frontend/fingertips-frontend/components/views/TwoOrMoreIndicatorsEnglandView/TwoOrMoreIndicatorsEnglandView.test.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+const fullSelectedIndicatorsData: IndicatorDocument[] = [
+  mockIndicatorDocument('id 1'),
+  mockIndicatorDocument('id 2'),
+];
+
+const mockIndicatorsApi = mockDeep<IndicatorsApi>();
+ApiClientFactory.getIndicatorsApiClient = () => mockIndicatorsApi;
+
+describe('TwoOrMoreIndicatorsEnglandView', () => {
```
gareth-allan-bjss, line 53: This view will throw an error if the number of indicators selected is not correct, or if the area code isn't England. We should have unit tests to validate it does so.

frontend/fingertips-frontend/components/viewPlots/TwoOrMoreIndicatorsEnglandViewPlots/index.tsx
```diff
@@ -1,7 +1,80 @@
...
-  return <H2>TBC</H2>;
+type TwoOrMoreIndicatorsEnglandViewPlotProps = {
+  indicatorData: IndicatorWithHealthDataForArea[];
+  searchState: SearchStateParams;
+  indicatorMetadata: IndicatorDocument[] | undefined;
+};
+
+export const getLatestHealthDataPointForEngland = (
+  indicatorData: IndicatorWithHealthDataForArea,
+  latestPeriod: string | undefined
```
gareth-allan-bjss, line 26: Is it valid to pass this function `undefined`? What would that mean?

frontend/fingertips-frontend/components/viewPlots/TwoOrMoreIndicatorsEnglandViewPlots/index.tsx
```diff
@@ -1,7 +1,80 @@
...
+  indicatorData,
+  indicatorMetadata,
+  searchState,
+}: Readonly<TwoOrMoreIndicatorsEnglandViewPlotProps>) {
+  SearchStateManager.initialise(searchState);
+
+  const englandIndicatorData: EnglandAreaTypeIndicatorData[] =
+    indicatorData.map((indicator) => {
+      const hasHealthDataForEngland =
+        indicator.areaHealthData?.[0]?.healthData !== undefined;
```
gareth-allan-bjss, line 46: Why do we need to do this check? Can we not just pass through the `latestDataPeriod` and `unitLabel` if they're present, and only attempt to get the `latestEnglandHealthData` if `latestPeriod` is present?

frontend/fingertips-frontend/components/viewPlots/TwoOrMoreIndicatorsEnglandViewPlots/index.tsx
```diff
@@ -1,7 +1,80 @@
...
+          indicatorMeta?.indicatorID === indicator.indicatorId?.toString()
+      );
+
+      const latestPeriod = hasHealthDataForEngland
+        ? metaDataForIndicator?.latestDataPeriod
+        : undefined;
+
+      const latestEnglandHealthData = hasHealthDataForEngland
+        ? getLatestHealthDataPointForEngland(indicator, latestPeriod)
+        : undefined;
```
gareth-allan-bjss, line 59: I assume we've got functions elsewhere in the code that find the latest data point for an area. Why do this based on the date from the metadata? It's not necessarily wrong, I just wasn't expecting to see it.

frontend/fingertips-frontend/components/viewPlots/TwoOrMoreIndicatorsEnglandViewPlots/TwoOrMoreIndicatorsEnglandViewPlots.test.tsx
```diff
@@ -0,0 +1,124 @@
...
+      value: 890.305692,
+      lowerCi: 441.69151,
+      upperCi: 578.32766,
+      ageBand: allAgesAge,
+      sex: personsSex,
+      trend: HealthDataPointTrendEnum.NotYetCalculated,
+      deprivation: noDeprivation,
+    },
+  ],
+};
```
gareth-allan-bjss, line 37: Again, if we don't really care _what_ the data is, can we try and use values from `lib/mocks.ts` please? You could perhaps move the whole `mockEnglandHealthData` object there (but have it reference the existing mock health data point instead of defining its own).

frontend/fingertips-frontend/components/viewPlots/TwoOrMoreIndicatorsEnglandViewPlots/index.tsx
```diff
@@ -1,7 +1,80 @@
...
+
+export function TwoOrMoreIndicatorsEnglandViewPlots({
+  indicatorData,
+  indicatorMetadata,
+  searchState,
+}: Readonly<TwoOrMoreIndicatorsEnglandViewPlotProps>) {
+  SearchStateManager.initialise(searchState);
+
+  const englandIndicatorData: EnglandAreaTypeIndicatorData[] =
+    indicatorData.map((indicator) => {
```
gareth-allan-bjss, line 44: Can you extract this to a function and add a unit test for it please?

obarojohnson (approved):
I am approving because you have rectify my concerns , please wait for others comment to be rectify as well thanks

frontend/fingertips-frontend/components/organisms/EnglandAreaTypeTable/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+interface EnglandAreaTypeTableProps {
+  indicatorData: EnglandAreaTypeIndicatorData[];
+}
+
+export function EnglandAreaTypeTable({
+  indicatorData,
+}: Readonly<EnglandAreaTypeTableProps>) {
+  return (
+    <div data-testid={'englandAreaTypeTable-component'}>
```
williamwillis11, line 44: given @gareth-allan-bjss comment above if you update the name of this component please can you also ensure you update this id and the references in the playwright test files? :)

frontend/fingertips-frontend/components/viewPlots/TwoOrMoreIndicatorsEnglandViewPlots/index.tsx
priya-thorpe-bjss: Since I’m looping through the data here and the metadata has this property readily available it made sense to do it here.

gareth-allan-bjss: Fair enough.

frontend/fingertips-frontend/components/viewPlots/TwoOrMoreIndicatorsEnglandViewPlots/index.tsx
```diff
@@ -1,7 +1,80 @@
...
+} from '@/components/organisms/EnglandAreaTypeTable';
+import {
+  HealthDataPoint,
+  IndicatorWithHealthDataForArea,
+} from '@/generated-sources/ft-api-client';
+import { IndicatorDocument } from '@/lib/search/searchTypes';
+import { H3 } from 'govuk-react';
 
-export function TwoOrMoreIndicatorsEnglandViewPlots() {
-  return <H2>TBC</H2>;
```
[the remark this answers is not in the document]
priya-thorpe-bjss: The title has already been added on line 76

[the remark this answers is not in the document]
gareth-allan-bjss: Ah, my mistake!

frontend/fingertips-frontend/components/viewPlots/TwoOrMoreIndicatorsEnglandViewPlots/index.tsx
priya-thorpe-bjss: I have changed to latestPeriod?, if its undefined it wouldn't get the data for year requested. If its worth doing I can add in a condition to do that.

frontend/fingertips-frontend/components/organisms/EnglandAreaTypeTable/index.tsx
```diff
@@ -47,7 +44,7 @@ export function EnglandAreaTypeTable({
           <React.Fragment>
             <Table.Row>
               <Table.CellHeader colSpan={6}>
-                {EnglandAreaTypeHeaderEnum.England}
+                {englandArea.name}
```
gareth-allan-bjss, line 47: We shouldn't be using a property from the mock data to drive the name of the table header. As I mentioned in my earlier comment, I think we should probably just take the value for this header as a prop.

frontend/fingertips-frontend/components/views/TwoOrMoreIndicatorsEnglandView/TwoOrMoreIndicatorsEnglandView.test.tsx
```diff
@@ -0,0 +1,125 @@
...
+      [SearchParams.AreasSelected]: ['12345'],
+      [SearchParams.IndicatorsSelected]: ['1', '2'],
+    };
+
+    await expect(async () => {
+      await TwoOrMoreIndicatorsEnglandView({
+        searchState: searchState,
+        selectedIndicatorsData: fullSelectedIndicatorsData,
+      });
+    }).rejects.toThrow('error getting health indicator data for areas');
```
gareth-allan-bjss, line 71: Shouldn't this throw an `Invalid parameters provided to view` error?

frontend/fingertips-frontend/components/views/TwoOrMoreIndicatorsEnglandView/index.tsx
```diff
@@ -32,6 +32,13 @@ export default async function TwoOrMoreIndicatorsEnglandView({
...
 
+  if (
+    !selectedIndicatorsData ||
+    selectedIndicatorsData.length !== indicatorsSelected.length
+  ) {
+    throw new Error('invalid indicator metadata passed to view');
+  }
```
gareth-allan-bjss, line 40: I think you need some tests to validate this error is throw when the incorrect metadata is passed.

frontend/fingertips-frontend/components/viewPlots/TwoOrMoreIndicatorsEnglandViewPlots/index.tsx
gareth-allan-bjss: It's probably fine as it is.

frontend/fingertips-frontend/components/organisms/EnglandAreaTypeTable/index.tsx
priya-thorpe-bjss: The name England Area Type was because of view 5 contents in the confluence page. I can name it OneAreaMultipleIndicatorsTable.

frontend/fingertips-frontend/components/viewPlots/TwoOrMoreIndicatorsEnglandViewPlots/index.tsx
```diff
@@ -36,8 +37,9 @@ export const getLatestPeriodHealthDataPoint = (
 export const getEnglandIndicatorTableData = (
   indicatorData: IndicatorWithHealthDataForArea[],
   indicatorMetadata: IndicatorDocument[]
-): EnglandAreaTypeIndicatorData[] => {
+): OneAreaMultipleIndicatorsData[] => {
   return indicatorData.map((indicator) => {
+    console.log('indicator data', indicatorData);
```
gareth-allan-bjss, line 42: A rogue console log here.

frontend/fingertips-frontend/components/viewPlots/TwoOrMoreIndicatorsEnglandViewPlots/index.tsx
```diff
@@ -1,7 +1,97 @@
...
+  OneAreaMultipleIndicatorsTable,
+} from '@/components/organisms/OneAreaMultipleIndicatorsTable';
+
+type TwoOrMoreIndicatorsEnglandViewPlotProps = {
+  indicatorData: IndicatorWithHealthDataForArea[];
+  searchState: SearchStateParams;
+  indicatorMetadata: IndicatorDocument[];
+};
+
+export const getLatestPeriodHealthDataPoint = (
```
obarojohnson, line 25: There is a function for this already , maybe you can reuse it. its part of the Pyramid function frontend/fingertips-frontend/lib/chartHelpers/chartHelpers.ts

frontend/fingertips-frontend/lib/ViewsHelpers.ts
```diff
@@ -9,3 +18,38 @@ export function chunkArray(
...
+  let healthIndicatorData: IndicatorWithHealthDataForArea | undefined;
+
+  try {
+    const healthIndicatorDataChunks = await Promise.all(
+      chunkArray(areaCodesToRequest, maxNumAreasThatCanBeRequestedAPI).map(
+        (requestAreas) =>
+          indicatorApi.getHealthDataForAnIndicator(
+            {
+              indicatorId: Number(indicatorId),
+              areaCodes: [...requestAreas],
```
gareth-allan-bjss, line 35: I think, given #425, we're going to have to pass an `areaType` parameter here.

gareth-allan-bjss (approved):
Approving on the basis that we can address some of the unresolved comments in a subsequent PR.

PR #391, opened 2025-04-03
DHSCFT-423: Fetch quartiles data

zoenolan:
# Description

Jira ticket: [DHSCFT-423](https://bjss-enterprise.atlassian.net/browse/DHSCFT-423)

Make use of the new API end point to fetch quartile data and populate the best and worst fields in the table

## Changes

- Use the new quartile endpoint
- Pass the quartile data from the view to the table via the view plot

## Validation
- Added extra unit tests to cover the quartile data
- E2E run and updated to match the new data being used for the table
- Manually checked the chart page and spine chart table 

Screenshots for the playwright e2e tests two indicator test have been updated.


frontend/fingertips-frontend/components/organisms/SpineChart/index.tsx
```diff
@@ -0,0 +1,6 @@
+export interface SpineChartProps {
+  best: number;
+  secondBest: number;
```
dlzhry2, line 3: Appreciate we can't really call it upper or lower quartile due to polarity. But `secondBest` and `secondWorst` could be a bit misleading as it could mean secondBest value. Would `best`, `bestQuartile`, `worstQuartile` and `worst` be clearer?

Feel free to dismiss given we wanna get this in ASAP and I don't think my suggestion is particularly great!

frontend/fingertips-frontend/components/organisms/SpineChart/index.tsx
```diff
@@ -0,0 +1,6 @@
...
+  secondWorst: number;
+  worst: number;
```
dlzhry2, line 5: I'm new to this area, but don't we need the median too (q2Value) in order to draw the line in the middle of the spine chart?

frontend/fingertips-frontend/components/viewPlots/TwoOrMoreIndicatorsAreasViewPlots/index.tsx
```diff
@@ -5,58 +5,92 @@ import { Heatmap, HeatmapIndicatorData } from '@/components/organisms/Heatmap';
...
+    definition: rowIndicatorDefinition,
+  };
+
+  return rowIndicator;
+}
+
+export function mapToSpineChartTableStatistics(
+  quartileData: QuartileData
+): SpineChartProps {
+  const q0Value = quartileData.q0Value ?? 0;
```
dlzhry2, line 45: One for later perhaps, but do the future tickets make it clear what to do if one or more quartile values are missing. I think there will need to be some basic handling for such a situation e.g. showing an X or insufficient data flag.

dlzhry2 (commented):
Had a little test looks good to me. Only comment that needs looking at (and I could be wrong) is the one to do with the median.

Happy to approve once discussed.

dlzhry2 (approved):

frontend/fingertips-frontend/components/organisms/SpineChart/index.tsx
zoenolan: I wasn't sold on those names. Your suggestion are better. I'll make that change

zoenolan: and the code updated now

frontend/fingertips-frontend/components/organisms/SpineChart/index.tsx
zoenolan: The spine charts don't use the median for the middle value but what is shown as England value in the table. We never need the median but it made sense for the API to return the data anyway

PR #395, opened 2025-04-03
Feature/dhscft 150 pyramid table

obarojohnson:
# Description

Jira ticket: [DHSCFT-150](https://bjss-enterprise.atlassian.net/browse/DHSCFT-150)

Create a table component for the pyramid chart data.

## Snapshot 
[

https://github.com/user-attachments/assets/c459d325-8c3d-4ab4-acfe-246c8b5bac65

](url)

Single Data Display 
image

## Changes

- New component created to align with the Figma design
- Unit test and component testing added
- Integrated the table component into the TabContainer for the chart and the table
- Etc.


frontend/fingertips-frontend/components/organisms/PopulationPyramid/createChartOptions.ts
```diff
@@ -152,11 +161,20 @@ const createAdditionalChartSeries = (
   const series: Array<SeriesOptionsType> = [];
 
   if (dataForGroup) {
+    const femaleSeries = computeDataPercentages(
+      dataForGroup.femaleSeries,
+      dataForGroup.total
+    );
+    const maleSeries = computeDataPercentages(
```
alibombaye, line 168: I suggest we rename these to `femaleGroupSeries` and `maleGroupSeries` so it doesn't get confused with the series above.

frontend/fingertips-frontend/components/organisms/PopulationPyramid/createChartOptions.ts
```diff
@@ -178,10 +196,18 @@ const createAdditionalChartSeries = (
   }
 
   if (dataForBenchmark) {
+    const femaleSeries = computeDataPercentages(
+      dataForBenchmark.femaleSeries,
+      dataForBenchmark.total
+    );
+    const maleSeries = computeDataPercentages(
```
alibombaye, line 203: Same here. `femaleBenchmarkSeries` and `maleBenchmarkSeries`

frontend/fingertips-frontend/components/organisms/PopulationPyramidChartTable/PopulationDataTable.tsx
```diff
@@ -0,0 +1,147 @@
...
+const StyledAreaNameHeader = styled(StyledAlignLeftHeader)(
+  typography.font({ size: 19 }),
+  {
+    borderTop: `solid #F3F2F1 2px`,
+    textAlign: 'center',
+    fontWeight: 700,
+  }
+);
+
+const getSortAgeBandIndexes = (ageBands: string[] | undefined): number[] => {
```
alibombaye, line 41: wouldn't a function like this already exist to do the same on the Pyramid view too?

frontend/fingertips-frontend/components/organisms/PopulationPyramidChartTable/PopulationDataTable.tsx
```diff
@@ -0,0 +1,147 @@
...
+    const males = healthDataForArea.maleSeries.reduce((prev, current) => {
+      return (prev ?? 0) + (current ?? 0);
+    }, 0);
+
+    const females = healthDataForArea.femaleSeries.reduce((prev, current) => {
+      return (prev ?? 0) + (current ?? 0);
+    }, 0);
+
+    return ['All ages', males, females];
+  })();
```
alibombaye, line 91: Not sure why this needs to be a function. Can't these just be plain const values?

frontend/fingertips-frontend/components/organisms/PopulationPyramidWithTable/index.tsx
```diff
@@ -189,7 +190,15 @@ export const PopulationPyramidWithTable = ({
...
                   title: 'Table',
-                  content: <div>The table here</div>,
+                  content: (
+                    <div>
+                      <PopulationPyramidChartTable
+                        healthDataForArea={selectedArea}
+                        benchmarkData={convertedData.benchmark}
+                        groupData={convertedData.group}
+                      />
+                    </div>
```
alibombaye, line 200: Do we need the surrounding `<div>` tags?

frontend/fingertips-frontend/lib/chartHelpers/preparePopulationData.test.ts
```diff
@@ -212,8 +213,12 @@ describe('convertHealthDataForAreaForPyramidData', () => {
...
     const actual: PopulationDataForArea | undefined =
       convertHealthDataForAreaForPyramidData(mockHealthDataForArea, 2023);
-    expect(actual?.femaleSeries).toEqual(mockFemaleSeries);
-    expect(actual?.maleSeries).toEqual(mockMaleSeries);
+    expect(
+      computeDataPercentages(actual?.femaleSeries ?? [], actual?.total ?? 0)
+    ).toEqual(mockFemaleSeries);
+    expect(
+      computeDataPercentages(actual?.maleSeries ?? [], actual?.total ?? 0)
+    ).toEqual(mockMaleSeries);
```
alibombaye, line 221: Not sure how these tests are working

frontend/fingertips-frontend/components/organisms/PopulationPyramidChartTable/PopulationDataTable.test.tsx
```diff
@@ -0,0 +1,53 @@
...
+  test('take a snapshot', () => {
+    const { asFragment } = render(
+      <PopulationDataTable
+        headers={['Age Band', 'Females', 'Males']}
+        title="Population Data"
+        healthDataForArea={mockHealthDataForArea}
+      />
+    );
+    expect(asFragment()).toMatchSnapshot();
+  });
```
alibombaye, line 52: I think some granular tests for this component would be useful to check the number of rows are as expected and the footer is also rendered. What we have found is the snapshot is not ideal to identify these sorts of issues.

frontend/fingertips-frontend/components/organisms/PopulationPyramidChartTable/PopulationDataTable.tsx
obarojohnson: I don't think so, since since this is the only time we needed to sort using the Indexes against the bandwidth

frontend/fingertips-frontend/components/organisms/PopulationPyramidChartTable/PopulationDataTable.tsx
obarojohnson: The reason I added a function is that I needed to aggregate it. However, I suppose I could do it one by one and then use a constant to capture the result. But the main reason I added a function was to scope the variables I created so they would be out of scope once the function returns

frontend/fingertips-frontend/lib/chartHelpers/preparePopulationData.ts
```diff
@@ -37,13 +38,14 @@ const sortHealthDataByAgeBand = (data: HealthDataPoint[]) => {
   });
 };
 
-function generatePopulationSeries(
-  healthData: HealthDataPoint[],
-  totalPopulation: number
+export function computeDataPercentages(
```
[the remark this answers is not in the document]
obarojohnson: image

frontend/fingertips-frontend/lib/chartHelpers/preparePopulationData.test.ts
obarojohnson: Can go through them with you, Its a regular test to check if the compute data percentage matches what I am expecting.

frontend/fingertips-frontend/components/organisms/PopulationPyramidChartTable/PopulationPyramidChartTable.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+      />
+    );
+
+    expect(screen.getByText('Test Area')).toBeInTheDocument();
+    expect(
+      screen.queryByText('Benchmark: Benchmark Area')
+    ).not.toBeInTheDocument();
+  });
+
+  test('take a snapshot', () => {
```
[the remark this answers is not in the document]
obarojohnson: I agree, but I think it would add more tests since what I wanted to test has already been done to ensure the component is rendered correctly. However, I suppose I can check if the  tables are rendered correctly, adding the snapshot just put more awares if one decided to change the fontsize per initial design and it would appear in review I suppose.

frontend/fingertips-frontend/components/organisms/PopulationPyramidChartTable/PopulationDataTable.tsx
```diff
@@ -0,0 +1,147 @@
...
+              {items.map((value, valueIndex: number) => (
+                <StyledAreaNameHeader
+                  key={`footer-${healthDataForArea.areaName}-${valueIndex}`}
+                >
+                  {valueFormatter(value)}
+                </StyledAreaNameHeader>
+              ))}
+            </Table.Row>
+          );
+        })()}
```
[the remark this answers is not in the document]
obarojohnson: Interesting , I think its because I want to be able to call  `const items = filterValues
            ? filterValues(footerRowItems)
            : footerRowItems;`  first before rendering and at the same time I don't want to pre-render it before the Jsx is return to the renderer.  Maybe there is a better way to do this  :)

frontend/fingertips-frontend/lib/chartHelpers/preparePopulationData.test.ts
alibombaye: ah ok I think it would be better to have some explicit tests for the `computeDataPercentages` function you are not really testing this function. Since in the test you use the function for the expected result. If I'm understanding it correctly.

frontend/fingertips-frontend/components/organisms/PopulationPyramidChartTable/PopulationDataTable.tsx
alibombaye: we show the age ranges on the pop pyramid chart don't we. Is this not the same thing.

frontend/fingertips-frontend/lib/chartHelpers/preparePopulationData.test.ts
obarojohnson: Done now , I knew that is what you meant :) clean it up now

frontend/fingertips-frontend/components/organisms/PopulationPyramidChartTable/PopulationDataTable.tsx
obarojohnson: No its not , this is a sorting algorithm base on the index

obarojohnson: You re refering to this right `const sortHealthDataByAgeBand = (data: HealthDataPoint[]) => {
  return data.sort((a, b) => {
    const getLowerBandValue = (range: string) => {
      if (range.includes('+')) return parseInt(range.split('-')[0]);
      return parseInt(range.split('-')[0]);
    };
    return getLowerBandValue(a.ageBand.value) >
      getLowerBandValue(b.ageBand.value)
      ? -1
      : 1;
  });
};`

alibombaye: yeah that looks very similar. Just wondering if it could be re-used. Specially that sorting part.

frontend/fingertips-frontend/components/organisms/PopulationPyramidChartTable/PopulationDataTable.tsx
```diff
@@ -0,0 +1,155 @@
...
+              return (
+                <StyledAreaNameHeader key={header}>
+                  {header}
+                </StyledAreaNameHeader>
+              );
+            })}
+          </Table.Row>
+        }
+      >
+        {indexes.map((index, _) => {
```
alibombaye, line 118: Do we need the `_`?

alibombaye (approved):

PR #404, opened 2025-04-04
DHSCFT-571: Use new getAreas call to get all selected areas in one request

alibombaye:
# Description

Jira ticket: [DHSCFT-571](https://bjss-enterprise.atlassian.net/browse/DHSCFT-571)

Updates to use new endpoint to get areas selected data in one api request.

Screenshot 2025-04-04 at 16 34 59

## Validation

Tested against the API and MSW


frontend/fingertips-frontend/lib/areaFilterHelpers/getSelectedAreasData.test.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+    const selectedAreaData = await getSelectedAreasDataByAreaType(
+      ['A001', 'A002'],
+      'nhs-regions'
+    );
+
+    expect(selectedAreaData).toEqual([area1]);
+  });
+
+  it('should default to filtering by england areaType when areaTypeSelected in not provided', async () => {
```
williamwillis11, line 109: is not provided

frontend/fingertips-frontend/lib/areaFilterHelpers/getSelectedAreasData.ts
```diff
@@ -0,0 +1,37 @@
...
+  const selectedAreasData =
+    areasSelected && areasSelected?.length > 0
+      ? (
+          await areasApi.getAreas(
+            {
+              areaCodes: areasSelected ?? [],
+            },
+            API_CACHE_CONFIG
+          )
+        ).filter((area) => area.areaType.key === determineAreaTypeSelected)
```
stuc-bjss, line 25: any chance we can do this filtering in the api to avoid passing data we don't need into nextjs

alibombaye: Yeah good point. Myself and @dlzhry2 had this discussion this morning on exactly this and believed it was the correct thing for the api to just return all the areas and let the FE then decide what it needed.

dlzhry2: Yup, the existing endpoint: `/areas/{area_code}` it turns out is not actually idempotent. It just returns the first result it gets, which could be either.

Only applies to some areas which fall into both districts and counties. Ultimately not a problem with the existing endpoint, as the UI knows which area type has been requested and when calling health measures we pass the area code and filter on area type if relevant.

Fawad has dropped a clarification for how this works in the Swagger. But could potentially extend this new endpoint (and perhaps the existing one) to also allow filtering for area type, so you can get one rather than two results back for certain area codes.

frontend/fingertips-frontend/lib/areaFilterHelpers/getSelectedAreasData.ts
```diff
@@ -0,0 +1,37 @@
...
+      : [];
+
+  if (
```
dlzhry2, line 28: Not sure if we have custom behaviour for the generated client, so maybe we would get to this point. But if a area code/set of codes returns nothing, then you get a 404 error. See: `api/DHSC.FingertipsNext.Modules.Area/HttpClientScripts/test-areas-multiple-areas.http`

Would we want to catch the API call and logs out an error message plus the contents of the API error?

frontend/fingertips-frontend/lib/areaFilterHelpers/getSelectedAreasData.test.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+    expect(mockAreasApi.getAreas).toHaveBeenCalledWith(
+      { areaCodes: ['A001', 'A002', 'A003'] },
+      API_CACHE_CONFIG
+    );
+    expect(selectedAreasData).toEqual([area1, area2, area3]);
+  });
+
+  it('should return an array of selectedAreasData and filter out areas that do not match the areaType provided', async () => {
+    const area1 = generateMockArea('A001', 'nhs-regions');
```
dlzhry2, line 70: Minor as it's test data, but I think the only area codes that can belong to both are districts + counties and unitary authorities. Apologies if I'm wrong. But was thinking if that's the only scenario, it could be useful for the scenario to align so it's a bit more descriptive and helpful to whoever picks this project up next.

frontend/fingertips-frontend/lib/areaFilterHelpers/getSelectedAreasData.test.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+    expect(mockAreasApi.getAreas).toHaveBeenCalledWith(
+      { areaCodes: ['A001', 'A002'] },
+      API_CACHE_CONFIG
+    );
+    expect(selectedAreasData).toEqual([area1, area2]);
+  });
+
+  it('should throw an error if no areaData is found for any of the areaSelected', async () => {
+    mockAreasApi.getAreas.mockResolvedValue([]);
```
dlzhry2, line 90: I don't think the API will resolve to this with a 200.

dlzhry2 (commented):
Looks really good to me. Happy to approve after quick discussion - just a minor comment raised regarding 404 scenario.

frontend/fingertips-frontend/lib/areaFilterHelpers/getSelectedAreasData.test.tsx
alibombaye: The function in general has no context about the the area-codes that cause this. Since this can change in the future or more could be added and the test description is no longer valid. But I do think it's something that belongs in more of the e2e tests since it a particular behaviour related to the data we get back. I'll have a chat with @williamwillis11 about potentially adding something there.

dlzhry2 (approved):

PR #412, opened 2025-04-07
don't add population indicators to indicators.json

andyevans-bjss:
# Description

Jira ticket: [DHSCFT-403](https://bjss-enterprise.atlassian.net/browse/DHSCFT-403)

To ensure population indicators don't appear in search results don't put them in indicators.json in the first place. 
Plus a couple of code tidy tasks.

## Validation

Review indicators.json to ensure that the 2 pop indicators are not there


DataCreator/DataCreator/IndicatorEntity.cs
```diff
@@ -58,6 +58,8 @@ public string IndicatorDefinition
         public bool UsedInPoc { get; set; }
 
         public bool HasInequalities { get; set; }
+
+        public bool HideInSearch => IndicatorID == 337 || IndicatorID == 92708;
```
dlzhry2, line 62: Could be worth a comment to explain what and why we're hiding things.

dlzhry2 (approved):

DataCreator/DataCreator/IndicatorEntity.cs
andyevans-bjss: done

PR #416, opened 2025-04-07
Feature/dhscft 605 batch request 100

obarojohnson:
# Description

Jira ticket: [DHSCFT-605](https://bjss-enterprise.atlassian.net/browse/DHSCFT-605)

When navigating to a chart page with a large number of areas selected , areas code are more than 100 will mostly cause a crash on the pyramid fixed.

image


frontend/fingertips-frontend/app/chart/PopulationPyramidWithTableDataProvider/index.tsx
```diff
@@ -15,11 +15,63 @@ import {
...
+  populationIndicatorID: number,
+  areaCodesToRequest: string[]
+): Promise<IndicatorWithHealthDataForArea> => {
+  const indicatorApi = ApiClientFactory.getIndicatorsApiClient();
+  let population: IndicatorWithHealthDataForArea | undefined = undefined;
+
+  for (const requestAreas of chunkArray(areaCodesToRequest)) {
+    if (!requestAreas) continue;
+    try {
+      const data = await indicatorApi.getHealthDataForAnIndicator(
```
steves-bjss, line 43: Can we execute these in parallel please - the latency will become progresivley longer with this implementation as we wait for each call to return.  Have a look at the 'TwoOrMoreIndicatorAreasView' lines 50-60 - this implementation uses Promise.all to make all the calls in parallel.

steves-bjss (changes requested):

frontend/fingertips-frontend/app/chart/PopulationPyramidWithTableDataProvider/index.tsx
gareth-allan-bjss: Can we extract a reusable function for this from the code in `TwoOrMoreIndicatorAreasView` perhaps?

gareth-allan-bjss: @priya-thorpe-bjss has done something similar in #387. I think it would be good to see if we can use the same function here, as then we only have a single place where we've implemented this batching logic.

frontend/fingertips-frontend/app/chart/PopulationPyramidWithTableDataProvider/PopulationPyramidWithTableDataProvider.test.tsx
```diff
@@ -67,6 +67,30 @@ describe('PopulationPyramidWithTableDataProvider', () => {
     expect(mockGetHealthDataForAnIndicator).toHaveBeenCalledTimes(1);
   });
 
+  it('renders PopulationPyramidWithTable with more than 100 areas data to be fetch', async () => {
```
williamwillis11, line 70: to be fetched

frontend/fingertips-frontend/app/chart/PopulationPyramidWithTableDataProvider/index.tsx
obarojohnson: I’ve seen it and it’s done now. It depends on which one goes in first — if this one goes in first, then the function can be reused. If hers goes in, I’ll have to modify it to allow usage without the internal implementations. Or could just add a comment to prevent me changing it if incase hers go in first thanks.

gareth-allan-bjss: #387 has now been merged @obarojohnson.

steves-bjss (approved):

PR #420, opened 2025-04-08
DHSCFT-224: Data type inequalities dropdown

adawari-thomas-bjss:
# Description

Jira ticket: [DHSCFT-224](https://bjss-enterprise.atlassian.net/browse/DHSCFT-224)

## Changes

- Created type selection drop down component for inequalities
- Added type selection component to Inequalities for single area component
- Added type selection component to Inequalities trend component

## Validation

Tested against API

deprivation type selected for single time period
image
image

deprivation type selected for trend
image

sex type selected for trend
image


frontend/fingertips-frontend/components/molecules/Inequalities/InequalitiesTrend/index.tsx
```diff
@@ -47,39 +50,40 @@ export function InequalitiesTrend({
...
+  const inequalityCategories = getInequalityCategories(healthIndicatorData);
+
+  if (!inequalityCategories.length) return null;
 
-  const inequalityCategory = getInequalityCategory(type, healthIndicatorData);
+  const type =
+    inequalityCategories[0] === sexCategory ||
+    inequalityTypeSelected === sexCategory
+      ? InequalitiesTypes.Sex
+      : InequalitiesTypes.Deprivation;
```
gareth-allan-bjss, line 71: Might be worth extracting this to a helper function, given it's the same code as in the `InequalitiesForSingleTimePeriod` component.

frontend/fingertips-frontend/components/molecules/Inequalities/InequalitiesTrend/inequalitiesTrend.test.tsx
```diff
@@ -1,23 +1,87 @@
...
+const mockSearchStateContext: SearchStateContext = {
+  getSearchState: mockGetSearchState,
+  setSearchState: jest.fn(),
+};
+jest.mock('@/context/SearchStateContext', () => {
+  return {
+    useSearchState: () => mockSearchStateContext,
+  };
+});
+
```
gareth-allan-bjss, line 43: I don't think we need to add all this, do we? We're not using `useSearchState` in this component (any more) are we?

frontend/fingertips-frontend/components/molecules/Inequalities/InequalitiesTrend/inequalitiesTrend.test.tsx
```diff
@@ -30,6 +94,14 @@ describe('InequalitiesTrend suite', () => {
     expect(
       screen.getByText(/Inequalities data over time/i)
     ).toBeInTheDocument();
+    expect(
+      screen.getByTestId('inequalitiesTypes-dropDown-component')
+    ).toBeInTheDocument();
+    expect(inequalitiesTypesDropDown).toBeInTheDocument();
```
gareth-allan-bjss, line 100: Isn't this check the same as the one on the line before?

frontend/fingertips-frontend/components/molecules/Inequalities/InequalitiesTrend/inequalitiesTrend.test.tsx
```diff
@@ -30,6 +94,14 @@ describe('InequalitiesTrend suite', () => {
...
+    expect(inequalitiesOptions).toHaveLength(2);
```
gareth-allan-bjss, line 101: Is this meant to be?:
```suggestion
    expect(inequalitiesDropDownOptions).toHaveLength(2);
```

frontend/fingertips-frontend/components/molecules/Inequalities/InequalitiesTypesDropDown/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+export function InequalitiesTypesDropDown({
+  inequalitiesOptions,
+  component,
+}: Readonly<InequalitiesTypeDropDownProps>) {
+  const pathname = usePathname();
+  const { replace } = useRouter();
+  const { getSearchState } = useSearchState();
+  const searchState = getSearchState();
+  const inequalityTypeSelectedSearchParam =
+    mapComponentToStateParams[component];
```
gareth-allan-bjss, line 30: Instead of passing in the `InequalitiesComponentType` value and using it to get a value from `mapComponentToStateParams`, why not just pass in the value to use for `inequalityTypeSelectedSearchParam`?

frontend/fingertips-frontend/components/molecules/Inequalities/InequalitiesTypesDropDown/inequalitiesTypesDropDown.test.tsx
```diff
@@ -0,0 +1,82 @@
...
+  getSearchState: mockGetSearchState,
+  setSearchState: jest.fn(),
+};
+jest.mock('@/context/SearchStateContext', () => {
+  return {
+    useSearchState: () => mockSearchStateContext,
+  };
+});
+
+describe('InequalitiesTypesDropDown suite', () => {
```
gareth-allan-bjss, line 40: I think we need a test that validates the correct option is selected when a (non-default) value is set in the search state

frontend/fingertips-frontend/components/molecules/Inequalities/InequalitiesTrend/inequalitiesTrend.test.tsx
alibombaye: Just sometimes on this. If a child component is using `useSearchState` then we still need to mock it. Or else it will complain.

gareth-allan-bjss: Ah, OK, makes sense. I'm happy for you to resolve this @adawari-thomas-bjss, if this is the case.

frontend/fingertips-frontend/components/molecules/Inequalities/InequalitiesTypesDropDown/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+    replace(searchStateManager.generatePath(pathname), { scroll: false });
+  };
+
+  const { [inequalityTypeSelectedSearchParam]: selectedType } =
+    searchStateManager.getSearchState();
+
+  return (
+    <div data-testid="inequalitiesTypes-dropDown-component">
+      <H5 style={{ marginBottom: '5px' }}>Select an inequality type</H5>
+      <StyledSelect
```
alibombaye, line 48: If we can use `StyleFilterSelect` for these dropdowns. It's what I've got for the dropdown on the areaFilters and what I've done for the select period and select area for inequalities.

```
const StyledFilterSelect = styled(Select)({
  span: {
    fontWeight: 'bold',
  },
  select: {
    width: '100%',
  },
  marginBottom: '2em',
});
```
At the moment this is just duplicated. So can just duplicate this here and we can tech debt to consolidate in a single place.

Then can be changed to something like this. We don't need the H5 tag and have similar styling across the dropdowns.
```
      <StyledFilterSelect
        label="Select an area"
        data-testid="chart-area-select"
        input={{
          value: selectedArea,
          onChange: (e) => chartAreaTypeSelected(e.target.value),
        }}
      >
        {availableAreas?.map((area) => (
          <option key={area.code} value={area.code}>
            {area.name}
          </option>
        ))}
      </StyledFilterSelect>
```

frontend/fingertips-frontend/components/molecules/TimePeriodDropDown/index.tsx
```diff
@@ -34,10 +40,16 @@ export function TimePeriodDropDown({
   const { [SearchParams.InequalityYearSelected]: selectedYear } =
     searchStateManager.getSearchState();
 
+  useEffect(() => {
+    if (selectedYearOverride) setSelectedYear(selectedYearOverride);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
alibombaye, line 46: How come we are doing this here?

adawari-thomas-bjss: I've added that for the edge case that @gareth-allan-bjss pointed out, where one inequality type has data for a particular year, then switching to another inequality type without data for the chosen year causes the chart to be empty. This ensures that the selected year can be set to a desired value when the component renders. In the parent component, it is set to the latest year with data for the chosen inequality when the edge case scenario occurs, otherwise it's undefined.

gareth-allan-bjss (approved):

frontend/fingertips-frontend/components/molecules/TimePeriodDropDown/index.tsx
adawari-thomas-bjss: This has been removed.

frontend/fingertips-frontend/components/organisms/Inequalities/inequalitiesHelpers.ts
```diff
@@ -397,12 +397,19 @@ export const filterHealthData = (
...
+  let disaggregatedDeprivationData = filterHealthData(
     healthIndicatorData.healthData,
     (data) => !data.deprivation.isAggregate
   );
+  if (selectedYear) {
+    disaggregatedDeprivationData = filterHealthData(
+      disaggregatedDeprivationData,
+      (data) => data.year === selectedYear
+    );
+  }
```
gareth-allan-bjss, line 412: Not a massive issue, but it seems a bit inefficient to filter the data twice. Couldn't we combine this into a single filter function?

frontend/fingertips-frontend/components/organisms/Inequalities/inequalitiesHelpers.test.ts
```diff
@@ -1124,6 +1124,18 @@ describe('isSexTypePresent', () => {
   it('should return true if sex type is present', () => {
     expect(isSexTypePresent(MOCK_INEQUALITIES_DATA.healthData)).toBe(true);
   });
+
+  it('should return false if sex type is not present for year provided', () => {
+    expect(isSexTypePresent(MOCK_INEQUALITIES_DATA.healthData, 2020)).toBe(
+      false
+    );
+  });
```
gareth-allan-bjss, line 1132: I think this is passing erroneously because you don't have _any_ data for 2020. We should have a "Persons" data point for 2020 to be sure that this works as expected.

frontend/fingertips-frontend/playwright/testHelpers.ts
```diff
@@ -86,6 +91,7 @@ export function getScenarioConfig(
         isTabTable: false,
         hasTimePeriodDropDown: true,
         hasDetailsExpander: false,
+        hasTypeDropDown: true,
```
gareth-allan-bjss, line 94: Shouldn't we have an equivalent for the inequalities trend component?

frontend/fingertips-frontend/components/organisms/Inequalities/inequalitiesHelpers.ts
gareth-allan-bjss: I think you're still filtering twice, but this doesn't have to block the PR.

gareth-allan-bjss (approved):

gareth-allan-bjss (approved):

PR #422, opened 2025-04-08
fix an issue where some areas had incorrect child areas

andyevans-bjss:
# Description

Jira ticket: [DHSCFT-617](https://bjss-enterprise.atlassian.net/browse/DHSCFT-617)

Fix an issue where some ICBs were not showing correctly

## Changes

Clean area code in DataCreator after the child areas are calculated, otherwise we get incorrect data

## Validation

Create the data, build the database and check the issue is fixed locally


DataCreator/DataCreator/PholioDatabase/PholioDataFetcher.cs
```diff
@@ -119,13 +118,25 @@ public async Task<IEnumerable<AreaEntity>> FetchAreasAsync()
...
 
+        private static void CleanAreaCodes(List<AreaEntity> areas)
+        {
+            foreach (var area in areas)
+            {
+                area.AreaCode = area.AreaCode.CleanAreaCode();
+                foreach (var child in area.ChildAreas)
+                {
+                    child.AreaCode = child.AreaCode.CleanAreaCode();
+                }
```
gareth-allan-bjss, line 136: Can the child areas have their own children? Do we ever need to worry about cleaning their area codes?

api/DHSC.FingertipsNext.Modules.HealthData/HttpClientScripts/test-indicators-benchmark-quartile.http
```diff
@@ -8,9 +8,9 @@ GET http://localhost:5144/indicators/quartiles?indicator_ids=94063&area_code=N85
             "polarity": "NoJudgement",
             "year": 2023,
             "q0Value": 15.789474,
-            "q1Value": 65.25573,
-            "q2Value": 71.98212,
-            "q3Value": 77.030815,
+            "q1Value": 64.92538,
```
gareth-allan-bjss, line 11: Would we expect these values to have changed?

andyevans-bjss: yes, because some areas have different children and that affects the quartile calculation for areas that have the same type

gareth-allan-bjss: OK, makes sense.

DataCreator/DataCreator/PholioDatabase/PholioDataFetcher.cs
andyevans-bjss: no, in this structure areas can have children only. The children are just a list of area codes., it isn't a recursive hierarchy

gareth-allan-bjss: Ah, I see, this is a list of the area codes for children, not a list of more area objects. I was thinking it was a tree structure we might have to work our way down.

PR #424, opened 2025-04-08
DHSCFT-155: Add shapes to spine chart

stuc-bjss:
# Description

Jira ticket: [DHSCFT-155](https://bjss-enterprise.atlassian.net/browse/DHSCFT-155)

## Changes

- Add shapes to spine chart
- 
## Validation
image


frontend/fingertips-frontend/.test/spec/snaps/webkit/fully_integrated_e2e_tests/core_journeys.spec.ts/ONLY-AREA-then-select-TWO-PLUS-INDICATORS-and-TWO-PLUS-AREAS-then-check-the-charts-page-spineChartTable-component.png
dlzhry2, line 1: The second row looks a bit odd. Given that the worst and best are 15.8 and 13.8 respectively, while the values for the 2 areas are 12.2 and 12.7.

This is not introduced by this PR, just wanted to flag this. I will look into it on my branch.

dlzhry2 (approved):
All good from my end. Seems to be an issue with some of the data (think it'll be the way we are sorting/validating on the frontend) but the plotting looks good to me.

PR #430, opened 2025-04-09
DHSCFT-517_API_return_indicator_for_areas_with_no_data

WolfieKnee:
# Description

Jira ticket: [DHSCFT-517](https://bjss-enterprise.atlassian.net/browse/DHSCFT-517)

API returns `healthdata[]` for any requested areas with no data.

## Changes

- Add check for AreasWithNoData in service, and call getAreasAsync in repository to 'fill in' any gaps in the db response. 

## Validation
http://[localhost:5144/indicators/41101/data?area_type=counties-and-unitary-authorities&area_codes=E10000017](http://localhost:5144/indicators/41101/data?area_type=counties-and-unitary-authorities&area_codes=E10000017)
response as expected. 

No changes to screenshots


dlzhry2 (commented):
Nice to have - definitely not a must - but when testing, if you find an area with no health data for a particular indicator it would be lovely to have a .http API integration test to demonstrate that an area with no data will still be returned.

With no healthData key?

dlzhry2 (commented):
We'll need to regenerate the FE API client since we've now added a new param.

WolfieKnee:
> We'll need to regenerate the FE API client since we've now added a new param.

@dlzhry2 - this is being done in 518

api/DHSC.FingertipsNext.Modules.HealthData.UnitTests/Repository/HealthDataRepositoryTests.cs
```diff
@@ -769,6 +769,42 @@ public async Task Repository_ShouldIncludeResultsWithAllInequalityData_IfAllAreS
...
+        // assert
+        result.ShouldNotBeEmpty();
+        result.Count().ShouldBe(3);
+        result.ShouldBeEquivalentTo(new List<HealthMeasureModel>()
+        {
+            ResetKeys(healthMeasure0),
+            ResetKeys(healthMeasure1),
+            ResetKeys(healthMeasure2),
+        });
+    }
```
gareth-allan-bjss, line 806: I don't think this test is checking what the title says it is. You're adding health data to the DB for three areas, then requesting it for four areas and then validating you got data back for three areas.

I don't think you've actually changed anything in the `GetIndicatorDataAsync` method though, so I wouldn't necessarily expect it to return you data for areas that don't have data.

Should this be a unit test to validate your new `GetAreasAsync` method returns the areas requested instead?

api/definition/swagger.yaml
gareth-allan-bjss, line 1: Are you going to regenerate the API client in a follow-up PR?

WolfieKnee: yes, in 518

gareth-allan-bjss: Sorry, just saw you'd already said that!

gareth-allan-bjss (approved):
The change looks good, barring my comment about `HealthDataRepositoryTests.cs`.

PR #432, opened 2025-04-09
DHSCFT-615: Refactor spine chart table

dlzhry2:
# Description

Jira ticket: [DHSCFT-615](https://bjss-enterprise.atlassian.net/browse/DHSCFT-615)

Some refactoring. Hopefully will make the remaining changes a bit easier.
Note: I will defer refactoring the table headings and Spine chart table styling into the next ticket, as there is already too much going on in this one.

## Changes
Main part of the data changes is in `frontend/fingertips-frontend/components/viewPlots/TwoOrMoreIndicatorsAreasViewPlots/index.tsx`

- Streamlines the assembly of data for the Spine Chart into something (hopefully) more coherent
- Ensures that we always retrieve the **latest** data points for the area health data. (Quartiles is already latest by default, so we need to do the same to match up on the area data side).
- 1 functional change - do not show the group column if group is England since it is the benchmark.

## Validation

Tested locally and regenerated screenshots. Because the data is up to date, the plots look slightly more like expected. However, there are some remaining wrinkles to iron out to do with the plotting of data which is being looked at separately.


frontend/fingertips-frontend/components/organisms/SpineChartTable/index.tsx
```diff
@@ -1,159 +1,58 @@
...
- * I.e. the latest data point for both areas is the same year.
- */
-const checkLatestPeriodMatches = (
-  rowData: SpineChartTableRowProps
-): boolean => {
-  return (
-    rowData.indicatorHealthDataAreaOne.healthData[0].year ===
-    rowData.indicatorHealthDataAreaTwo?.healthData[0].year
+const sortByIndicator = (indicatorData: SpineChartIndicatorData[]) =>
+  indicatorData.toSorted(
```
dlzhry2, line 25: Any business requirement for this. Would have thought alphabetical if anything, or last updated. IndicatorID although known by some data analysts is not really relevant to any kind of sorting - in terms of being useful?

frontend/fingertips-frontend/components/organisms/SpineChartTable/SpineChartTableRow.tsx
```diff
@@ -57,92 +39,117 @@ export function SpineChartMissingValue({
...
       <StyledBenchmarkChart data-testid={`benchmark-range`}>
         <SpineChart
-          benchmarkValue={benchmarkValue ?? 0}
-          quartileData={benchmarkStatistics}
-          areaOneValue={areaOneValue}
-          areaTwoValue={areaTwoValue}
-          areaOneOutcome={areaOneOutcome}
-          areaTwoOutcome={areaTwoOutcome}
-          groupValue={groupValue}
+          benchmarkValue={quartileData.englandValue ?? 0}
```
dlzhry2, line 140: I have avoided refactoring the props for SpineChart. Hopefully this is useful @stuc-bjss @zoenolan since you are making changes in there.

In future (if time - ha!) could simple pass it the indicator data object, or split it down but at least provide it the `areasHealthData ` array so the child can just map over the values instead of needing a specific item for each.

frontend/fingertips-frontend/.test/spec/snaps/chromium/fully_integrated_e2e_tests/core_journeys.spec.ts/ONLY-AREA-then-select-TWO-PLUS-INDICATORS-and-TWO-PLUS-AREAS-then-check-the-charts-page-spineChartTable-component.png
dlzhry2, line 1: Group column and group plot no longer occurs when the group is England - good.
Data is now relevant for the latest period unlike before - not too bad.

Plotting fixup required due to the use of absolute value. We need to consider A. polarity of the indicator and B. consider that the point may be to the left of the centre value and hence need to be negative.

frontend/fingertips-frontend/components/organisms/SpineChartTable/index.tsx
zoenolan: I just checked [DHSCFT-153](https://bjss-enterprise.atlassian.net/browse/DHSCFT-153) that says 

> Then the data must be default sorted on Indicator name

dlzhry2: Will tackle in the final sub task.

PR #450, opened 2025-04-11
DHSCFT-634: omit rows from spine chart table if any data is not found

stuc-bjss:
# Description

Jira ticket: [DHSCFT-634](https://bjss-enterprise.atlassian.net/browse/DHSCFT-634)

Stopped spine charts from crashing if there is data missing

## Changes

- Returned null in place of throwing an error
- Filtered out null results

## Consideration

Began updating code to show X's when data was missing, but this turned out to be a much bigger issue than simply omitting the incomplete row data altogether.

## Validation

http://20.26.160.30/chart?si=age&is=30302&is=20401&as=E12000004&as=E12000006&ats=regions&gts=england&gs=E92000001&iys=2015

Indicator: "Hepatitis B vaccination coverage aged 2 years" is missing in the spine chart table illustrated below as there is missing data.
image

### DIFF
--- a/frontend/fingertips-frontend/components/organisms/SpineChartTable/SpineChartTableRow.test.tsx
+++ b/frontend/fingertips-frontend/components/organisms/SpineChartTable/SpineChartTableRow.test.tsx
@@ -104,7 +104,7 @@ describe('Spine chart table row', () => {
   });
 
   it('should have X for missing data', () => {
-    const indicatorWithMissingData = {
+    const indicatorWithMissingData: SpineChartIndicatorData = {
       ...mockIndicatorData,
       groupData: {
         ...mockIndicatorData.groupData,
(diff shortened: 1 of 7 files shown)


dlzhry2 (approved):
Looks good. Wonder if some console logs would be useful as it might be confusing as a dev in terms of why you've searched for something and there's no row there for the indicator.

Longer term, we'd also wanna do something about it for the user as it's a bit of an odd experience but this should do it for PoC right?

zoenolan (approved):

zoenolan:
> Longer term, we'd also wanna do something about it for the user as it's a bit of an odd experience but this should do it for PoC right?

Yes. Think it is fine for the PoC. Like you say the code could do with a refactor later on, not just for the missing data

zoenolan (approved):

PR #452, opened 2025-04-11
DHSCFT-611: Allow > 100 areas to be selected in the filter pane

gareth-allan-bjss:
# Description

Jira ticket: [DHSCFT-611](https://bjss-enterprise.atlassian.net/browse/DHSCFT-611)

This PR updates the `getSelectedAreasDataByAreaType` function used to retrieve area details for the area filter so that it does not request more than 100 areas at once. This is because the API returns an error if > 100 areas are requested, in order to prevent requests which would consume too many resources. The frontend now makes multiple requests in parallel.

## Changes

- Made `getSelectedAreasDataByAreaType` split requests to the `/areas` endpoint into batches of requests for 100 area codes, rather than attempting to request > 100 area codes
- Extracted the `chunkArray` helper function to its own file, to allow it to be reused
- Removed the brackets around the value unit in the line chart table's group column header (it didn't match the other column headers)

## Validation

Update to line chart table:
Screenshot 2025-04-11 at 17 26 31

Following the reproduction steps from the ticket now results in a page being displayed:
Screenshot 2025-04-11 at 17 30 36

Doing the same on the search results page no longer results in an error (it does on main):
Screenshot 2025-04-11 at 17 32 30

Doing the same on the home page no longer results in an error (it does on main):
Screenshot 2025-04-11 at 17 33 44


frontend/fingertips-frontend/lib/chunkArray.ts
```diff
@@ -0,0 +1,12 @@
+export const maxNumAreasThatCanBeRequestedAPI = 100;
```
obarojohnson, line 1: Do you need to export this variable?  - Its only use within the file

obarojohnson (changes requested):

obarojohnson (commented):
?

frontend/fingertips-frontend/lib/chunkArray.ts
gareth-allan-bjss: Good point, updated.

frontend/fingertips-frontend/lib/areaFilterHelpers/getSelectedAreasData.ts
```diff
@@ -4,25 +4,36 @@ import {
...
-      ? (
-          await areasApi.getAreas(
-            {
-              areaCodes: areasSelected ?? [],
-            },
-            API_CACHE_CONFIG
-          )
-        ).filter((area) => area.areaType.key === determineAreaTypeSelected)
+      ? (await getAreas(areasSelected)).filter(
+          (area) => area.areaType.key === determineAreaTypeSelected
```
dlzhry2, line 35: Optional comment: not related to your change but that existing variable name `determineAreaTypeSelected` is more suited as a method/function name than a var name. An option could be -> `selectedAreaType`.

gareth-allan-bjss: Good shout, I've updated.

obarojohnson (approved):

PR #453, opened 2025-04-14
DHSCFT-640: try to improve render speed - refactor heatmap hovers

stuc-bjss:
# Description

Jira ticket: [DHSCFT-640](https://bjss-enterprise.atlassian.net/browse/DHSCFT-640)

## Changes

- Separate table and hover actions into different components
- Make hover action more React standard pattern
- Reduce unnecessary renders


frontend/fingertips-frontend/components/organisms/Heatmap/heatmapHover.tsx
```diff
@@ -71,10 +71,15 @@ export const HeatmapHover: FC<HeatmapHoverProps> = ({
...
+  left,
+  top,
 }) => {
+  const styles = {
+    left: `${left}px`,
+    top: `${top}px`,
+  };
   return (
-    <StyledDivHover id={hoverId}>
+    <StyledDivHover style={styles}>
```
Joe-Hendley, line 82: I'm not 100% sure about passing through the style directly when we're using styled components

frontend/fingertips-frontend/components/organisms/Heatmap/useHeatmapTableData.ts
```diff
@@ -0,0 +1,26 @@
...
+  groupAreaCode?: string
+) => {
+  return useMemo((): MemoDataPrep => {
+    const { areas, indicators, dataPoints } =
+      extractSortedAreasIndicatorsAndDataPoints(indicatorData, groupAreaCode);
+    return {
+      headers: generateHeaders(areas, groupAreaCode),
+      rows: generateRows(areas, indicators, dataPoints),
+    };
+  }, [groupAreaCode, indicatorData]);
```
dlzhry2, line 25: I like the idea of using `useMemo` but in the case of this usage I'm not sure what it protects against. What else in the heatmap can cause a re-render that we are now avoiding using this.

The only thing that comes to mind are when the area codes or indicators are changed, in which case these functions will need to be rerun to get the correct data for the heatmap.

stuc-bjss:
```
  const { headers, rows } = useHeatmapTableData(indicatorData, groupAreaCode);
  const { hover, left, top, handleMouseOverCell } = useHeatmapHover();
```
useHeatmapHover, uses useState which causes a re-render in the Heatmap component in index.tsx everytime you hover over a different cell, which in turn calls this hook again hence the useMemo to reduce the iterating and sorting of data which hasn't changed

dlzhry2: Ah, I had failed to follow that bit. Well in that case that'll be quite handy! Thanks for the explanation.

PR #454, opened 2025-04-14
DHSCFT-626: Layout updates

zoenolan:
# Description

Jira ticket: [DHSCFT-626](https://bjss-enterprise.atlassian.net/browse/DHSCFT-626)

Makes some changes to the table layout based on feedback in the BA review.

## Changes

- Remove the indicator name from the hovers check mark button 
- We should have a space between the unit and value in the hover, when the unit isn’t %  check mark button 
- Value on the benchmark needs padding to the left check mark button
- Left align Value Unit column the same as the Period column  check mark button 
- Float the hovers above the chart div  check mark button 

## Validation
Ran unit tests. Manual checks of the table. 

Updated the chart component base screenshots in the playwright e2e tests.


stuc-bjss (approved):

frontend/fingertips-frontend/components/organisms/SpineChart/SpineChartOptions.ts
```diff
@@ -70,15 +75,13 @@ function formatSymbolHover(
               ${title}
               </span>
               <span style="display: block;">${period}</span>
-              <span style="display: block;">${indicatorName}</span>
-              <span style="display: block; float: right;">${formatNumber(value)}${units}</span></br/>
-              <span style="display: block; float: right;">${outcome}</span></br/>
-              <span style="display: block; float: right;">${benchmarkComparisonMethodToString(benchmarkComparisonMethod)}</span>
+              <span style="display: block;">${formatNumber(value)}${formatUnits(units)}</span></br/>
+              <span style="display: block;">${outcome}</span></br/>
```
obarojohnson, line 79: use 'div' here to prevent using <br> tag or just set the style property of span to display: block . I might have forget but break is not really good for creating new lines. There is a document [here ](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/br) that will explain why?

obarojohnson (approved):

dlzhry2 (approved):

frontend/fingertips-frontend/components/organisms/SpineChart/SpineChartOptions.ts
zoenolan: Thanks. <BR>'s removed

PR #455, opened 2025-04-14
Feature/dhscft 637 swap female and male order

obarojohnson:
# Description

Jira ticket: [DHSCFT-637](https://bjss-enterprise.atlassian.net/browse/DHSCFT-637)

Fix a bug

### DIFF
--- a/frontend/fingertips-frontend/components/organisms/PopulationPyramidChartTable/PopulationDataTable.tsx
+++ b/frontend/fingertips-frontend/components/organisms/PopulationPyramidChartTable/PopulationDataTable.tsx
@@ -90,7 +90,7 @@ export const PopulationDataTable = ({
   const points = healthDataForArea.ageCategories.map((value, index) => [
     value,
     healthDataForArea.maleSeries[index],
-    healthDataForArea.femaleSeries[index]
+    healthDataForArea.femaleSeries[index],
   ]);
 
   const footerRowItems = computeFooterItems(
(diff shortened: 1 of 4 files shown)


Nayira-BJSS:
Need to regenerate the snapshots (unit test) otherwise this looks good to me

dlzhry2 (commented):
Looks like some snapshots need regenerating or some tests fixing. Also, if there's any screenshots in place for the Pop Pyramid, this may need regenerating, since the data will be different now.

frontend/fingertips-frontend/components/organisms/PopulationPyramidChartTable/PopulationDataTable.tsx
```diff
@@ -89,8 +89,8 @@ export const PopulationDataTable = ({
 
   const points = healthDataForArea.ageCategories.map((value, index) => [
     value,
-    healthDataForArea.femaleSeries[index],
     healthDataForArea.maleSeries[index],
+    healthDataForArea.femaleSeries[index],
```
dlzhry2, line 93: Looks good. Would suggest in future it might be more readable if we had constructed a clear interface e.g. containing a key for `ageCategory` `maleValue` and `femaleValue`, as that might have made it clearer as to which data belonged with which sex rather than relying on the ordering that we put something into an array.

This looks like it'll do the job, and probably beyond the scope of what we need to do for now to get deep into refactoring this.

dlzhry2 (approved):

frontend/fingertips-frontend/components/organisms/PopulationPyramidChartTable/PopulationDataTable.tsx
dlzhry2: Looks good to me :)

PR #459, opened 2025-04-14
Feature/dhscft 518 thematic map to handle no data rebased

stuc-bjss:
F# Description

Jira ticket: [DHSCFT-518](https://bjss-enterprise.atlassian.net/browse/DHSCFT-518)

ThematicMap will show areas with 'No data available'

## Changes

- correct naming of tooltip components
- update tooltip to show 'X No data available' for `healthData[]`
- update thematicMapHelpers to handle `healthData[]`
- update OneIndicatorTwoOrMoreAreasViewPlots to pass indicatorDataAllAreas to ThematicMap 
- update OneIndicatorTwoOrMoreAreasView to use helper function to make additional API call for indicatorDataAvailableAreas to be used in the ThematicMap
- re-generate API msw for changes in 517

## Validation
[behaviour as expected against local mocks: http://localhost:3000/chart?si=94103&is=94103&ats=districts-and-unitary-authorities&gts=regions&gs=E12000002&gas=ALL](http://localhost:3000/chart?si=hospital&is=41101&ats=counties-and-unitary-authorities&gts=regions&gs=E12000002&gas=ALL)
image

No changes made to screenshots


frontend/fingertips-frontend/components/atoms/BenchmarkTooltipArea/index.tsx
```diff
@@ -28,10 +28,25 @@ export function BenchmarkTooltipArea({
   measurementUnit,
   tooltipType,
 }: Readonly<BenchmarkTooltipArea>) {
+  const indicatorDataForAreaForMostRecentYear =
+    getIndicatorDataForAreasForMostRecentYearOnly([indicatorData]);
```
alibombaye, line 32: Since this is just for a single areas indicator data. Can we use this function instead `sortHealthDataPointsByDescendingYear` 

Hopefully that may need the need to get the first element of the array underneath.

frontend/fingertips-frontend/lib/ViewsHelpers.ts
```diff
@@ -3,8 +3,35 @@ import {
...
+import { areaCodeForEngland } from '@/lib/chartHelpers/constants';
+
+export const maxNumAreasThatCanBeRequestedAPI = 100;
+
+/**
+ * Returns arrayToChunk broken into multiple arrays of a size allowed to be used in API calls
+ * @param arrayToChunk
+ * @param chunkSize optional chunk size, which will be coerced within a range the API supports
+ */
+export function chunkArray(
```
alibombaye, line 20: We already have a chunkArray function in `frontend/fingertips-frontend/lib/chunkArray.ts` would it be worth consolidating.

MonkeyDate: I think that might already have been done in main as I can only find one chunkArray function in my current branch

MonkeyDate:
> Not sure if its part of this PR - Is the API returning a different response when there is no health data for an area? We have a check in `ViewsWrapper` to show a message when there is no health data at all for all indicators and all areas. Does the helper function `hasHealthDataCheck` need updating too?

AFAIK everything was good. David was requesting his data with an extra flag (includesEmpty or something) that brings back empty datasets for real areas that have no data matching the request. Everywhere else that flag was omitted or defaulted to false so the data would be the same for all other components.

alibombaye:
> > Not sure if its part of this PR - Is the API returning a different response when there is no health data for an area? We have a check in `ViewsWrapper` to show a message when there is no health data at all for all indicators and all areas. Does the helper function `hasHealthDataCheck` need updating too?
> 
> AFAIK everything was good. David was requesting his data with an extra flag (includesEmpty or something) that brings back empty datasets for real areas that have no data matching the request. Everywhere else that flag was omitted or defaulted to false so the data would be the same for all other components.

OK just pulled this locally.  Everything still seems good in terms of that error handling functionality still working. But I'm not getting the hover with No data available.

frontend/fingertips-frontend/components/molecules/BenchmarkTooltip/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+        />
+      ) : null}
+      <BenchmarkTooltipArea
+        indicatorData={indicatorData}
+        benchmarkComparisonMethod={benchmarkComparisonMethod}
+        measurementUnit={measurementUnit}
+        tooltipType={'area'}
+      />
+    </div>
+  );
```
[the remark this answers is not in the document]
stuc-bjss: There are [here](https://github.com/dhsc-govuk/FingertipsNext/blob/main/frontend/fingertips-frontend/components/molecules/BenchmarkTooltip/BenchmarkTooltip.test.tsx) it's just that they didn't need changing and therefore didn't show up in the PR

stuc-bjss:
> > > Not sure if its part of this PR - Is the API returning a different response when there is no health data for an area? We have a check in `ViewsWrapper` to show a message when there is no health data at all for all indicators and all areas. Does the helper function `hasHealthDataCheck` need updating too?
> > 
> > 
> > AFAIK everything was good. David was requesting his data with an extra flag (includesEmpty or something) that brings back empty datasets for real areas that have no data matching the request. Everywhere else that flag was omitted or defaulted to false so the data would be the same for all other components.
> 
> OK just pulled this locally. Everything still seems good in terms of that error handling functionality still working. But I'm not getting the hover with No data available.

It is requesting data with the extra flag to include empty, but then filters out those areas with no health data for the rest of the charts.

alibombaye (approved):

PR #461, opened 2025-04-14
DHSCFT-646 - remove single decimal place zero

MonkeyDate:
# Description

Jira ticket: [DHSCFT-646](https://bjss-enterprise.atlassian.net/browse/DHSCFT-646)

Mainly changing the number format so that we don't require a single decimal place for `.0` values.

## Changes

- Change `formatNumber` function to use *up-to* 1 decimal place
- Inequalities line chart count now a whole number (do decimal place)
- Confidence intervals on compare areas Bar England are now formatted correctly.

## Validation

TBA


frontend/fingertips-frontend/.test/spec/snaps/webkit/fully_integrated_e2e_tests/core_journeys.spec.ts/ONLY-SUBJECT-then-select-ONE-INDICATOR-and-THREE-PLUS-AREAS-then-check-the-charts-page-populationPyramid-component.png
MonkeyDate, line 1: after discussion it is believed this update was missed from a previous, unrelated PR

PR #463, opened 2025-04-14
Feature/dhscft 578 latest date

andyevans-bjss:
# Description

Jira ticket: [DHSCFT-578](https://bjss-enterprise.atlassian.net/browse/DHSCFT-578)

Add a query string parameter to the indicator health data API endpoint that allows the client to request only the latest data rather than for all years.

## Validation

Run all unit tests and create a new unit test for this scenario

### DIFF
--- a/api/DHSC.FingertipsNext.Modules.HealthData.Repository/HealthDataRepository.cs
+++ b/api/DHSC.FingertipsNext.Modules.HealthData.Repository/HealthDataRepository.cs
@@ -17,18 +17,23 @@ public class HealthDataRepository(HealthDataDbContext healthDataDbContext) : IHe
 
     public async Task<IndicatorDimensionModel> GetIndicatorDimensionAsync(int indicatorId)
     {
-        var results = await _dbContext.IndicatorDimension
-            .Where(indicator => indicator.IndicatorId == indicatorId)
-            .Select(indicator => new IndicatorDimensionModel
-            {
-                IndicatorId = indicator.IndicatorId,
-                IndicatorKey = indicator.IndicatorKey,
-                Name = indicator.Name,
-                Polarity = indicator.Polarity,
-                BenchmarkComparisonMethod = indicator.BenchmarkComparisonMethod
-            })
-            .FirstOrDefaultAsync();
-        return results;
+        var model=await _dbContext.HealthMeasure
+           .Where(healthMeasure => healthMeasure.IndicatorDimension.IndicatorId == indicatorId)
+           .OrderByDescending(healthMeasure => healthMeasure.Year)
+           .Include(healthMeasure => healthMeasure.IndicatorDimension)
+           .Select(healthMeasure => new IndicatorDimensionModel
+           {
+               IndicatorId = healthMeasure.IndicatorDimension.IndicatorId,
+               IndicatorKey = healthMeasure.IndicatorKey,
+               Name = healthMeasure.IndicatorDimension.Name,
+               Polarity = healthMeasure.IndicatorDimension.Polarity,
+               BenchmarkComparisonMethod = healthMeasure.IndicatorDimension.BenchmarkComparisonMethod,
+               LatestYear = healthMeasure.Year
+           })
+           .Take(1)
+           .FirstOrDefaultAsync();
+       
+        return model;
     }
     
     public async Task<IEnumerable<HealthMeasureModel>> GetIndicatorDataAsync(int indicatorId, string[] areaCodes, int[] years, string[] inequalities)
--- a/api/DHSC.FingertipsNext.Modules.HealthData.Repository/HealthDataRepository.cs
+++ b/api/DHSC.FingertipsNext.Modules.HealthData.Repository/HealthDataRepository.cs
@@ -4,7 +4,6 @@ using Microsoft.Data.SqlClient;
 using Microsoft.EntityFrameworkCore;
 using System.Data;
 
-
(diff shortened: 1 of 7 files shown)


api/definition/swagger.yaml
```diff
@@ -766,6 +766,12 @@ components:
       description: Determines if areas with no data are returned as empty arrays, the default is false.
       schema:
         type: boolean
+    latest_only:
+      in: query
+      name: latest_only
+      description: Set to true to get data for the latest date period only, default is false. This overrides the years parameter if set to true.
+      schema:
+        type: boolean
```
alibombaye, line 774: Can you re-generate the FE api client. Just to make sure we haven't broken anything with the types. Should be ok by the looks of things.

api/DHSC.FingertipsNext.Modules.HealthData.Repository/HealthDataRepository.cs
```diff
@@ -18,20 +17,23 @@ public class HealthDataRepository(HealthDataDbContext healthDataDbContext) : IHe
...
-                IndicatorKey = x.IndicatorKey,
-                Name = x.Name,
-                StartDate = x.StartDate,
-                EndDate = x.EndDate,
-                Polarity = x.Polarity,
-                BenchmarkComparisonMethod = x.BenchmarkComparisonMethod,
-            })
-            .ToListAsync();
-        return results.FirstOrDefault();
+        var model=await _dbContext.HealthMeasure
```
dlzhry2, line 20:
```suggestion
        var mode l= await _dbContext.HealthMeasure
```

Minor but formatting. Don't think we have any linting imposed on the backend like we do on the frontend.

api/DHSC.FingertipsNext.Modules.HealthData.Service/IndicatorService.cs
```diff
@@ -117,6 +120,7 @@ IndicatorPolarity polarity
                 areaType
                 );
 
+
```
dlzhry2, line 123: Again, minor but probably doesn't need the extra new line.

api/DHSC.FingertipsNext.Modules.HealthData.Repository/HealthDataRepository.cs
```diff
@@ -18,20 +17,23 @@ public class HealthDataRepository(HealthDataDbContext healthDataDbContext) : IHe
...
+               IndicatorKey = healthMeasure.IndicatorKey,
+               Name = healthMeasure.IndicatorDimension.Name,
+               Polarity = healthMeasure.IndicatorDimension.Polarity,
+               BenchmarkComparisonMethod = healthMeasure.IndicatorDimension.BenchmarkComparisonMethod,
+               LatestYear = healthMeasure.Year
+           })
+           .Take(1)
+           .FirstOrDefaultAsync();
+       
+        return model;
```
dlzhry2, line 36: Not related to your change, but I wonder if we should modify line 48 to be consistent with 47 in this file. I.e. use EF.Constant. I know the list of years will never be as long as areaCodes, but it could be good for consistency and increases safety in case the endpoint ever changes to allow more years.

api/DHSC.FingertipsNext.Modules.HealthData.UnitTests/Services/IndicatorServiceTests.cs
```diff
@@ -85,6 +84,41 @@ public async Task GetIndicatorData_ShouldReturnExpectedResult()
         result.Content.AreaHealthData.ElementAt(0).ShouldBeEquivalentTo(expected);
     }
 
+    [Fact]
+    public async Task GetIndicatorData_ShouldReturnExpectedResult_LatestDataOnly()
+    {
+        const string expectedAreaCode2 = "Code2";
+        const string expectedAreaName2 = "Area 2";
+
+        var healthMeasure1 = new HealthMeasureModelHelper(year: 2023)
```
dlzhry2, line 93: Think there's limited/no value in defining healthMeasure1 and healthMeasure3. At the Service level, we are defining what the repository will hand back to us, so there is no filtering going on here. Could probably be dropped.

What we should be doing for this PR is testing at the repository level: `GetIndicatorDimensionAsync` needs at least one test where we setup one or two records against an indicator and then assert that the correct value for `latestYear` is returned.

dlzhry2 (commented):
A couple of minor comments plus one request for an additional bit of testing at the repository level.

Other than that looks good.

api/DHSC.FingertipsNext.Modules.HealthData.Repository/HealthDataRepository.cs
andyevans-bjss: done

api/DHSC.FingertipsNext.Modules.HealthData.Repository/HealthDataRepository.cs
andyevans-bjss: done

api/DHSC.FingertipsNext.Modules.HealthData.UnitTests/Services/IndicatorServiceTests.cs
andyevans-bjss: changed

dlzhry2 (commented):
Thanks for the changes - still think it'd be good if we add tests at the repository level.

dlzhry2 (approved):

dlzhry2 (approved):

PR #465, opened 2025-04-15
DHSCFT-643 Add support for indicators with financial year periods

dlzhry2:
# Description

Jira ticket: [DHSCFT-643](https://bjss-enterprise.atlassian.net/browse/DHSCFT-643)

Currently a spike - above ticket will need fleshing out once we have agreed on the approach.

As raised by stakeholders in Sprint review, there is more complexity to the period for many indicators. Besides the fact that some may be collected with a higher frequency than annual e.g. quarterly, some are also collected in line with financial year rather than the calendar year.

This spike only looks into rendering financial year - it is agreed with stakeholders that accounting for different period frequencies is something for beyond PoC..

## Changes

- Updated the DB and DataCreator so that there is an additional concept of `periodLabel` for health measure data
- Simple change to the API to retrieve and provide this data
- Frontend changes (Line Chart only for the spike) to display the periodLabel in the chart and table
- Also fixed up the linting issues associated with the addition of the new field (quite a bit of work)

## Next steps

Discussion with team and stakeholders required to size and decide whether to proceed. Additionally, this spike will inform the technical approach.

## Validation

Manual only at this stage

## Screenshots

Existing fingertips:
image

New Fingertips (GDS text):
image

image


frontend/fingertips-frontend/components/organisms/LineChart/lineChartHelpers.ts
```diff
@@ -44,7 +45,13 @@ export const lineChartDefaultOptions: Highcharts.Options = {
   xAxis: {
     tickLength: 0,
     allowDecimals: false,
-    labels: { style: { fontSize: AXIS_LABEL_FONT_SIZE } },
+    labels: {
+      style: {
+        fontSize: AXIS_LABEL_FONT_SIZE,
+        width: 40,
```
dlzhry2, line 51: Probably only need to do this if it's financial year. Also there will be a better way to do this beyond spike.

I had hoped [allowOverlap](https://api.highcharts.com/highcharts/xAxis.labels.allowOverlap) -> false would have done it, but Highcharts is not playing ball.

frontend/fingertips-frontend/components/organisms/LineChart/lineChartHelpers.ts
```diff
@@ -17,8 +17,9 @@ export enum LineChartVariant {
   Inequalities = 'inequalities',
 }
 
+// This is a hacky way to do it. Can make the label transformation common
```
dlzhry2, line 20: Not only does this need to be made common, this is also a bug as it will format points which need to be calendar year rather than financial year. So it'll need to take an additional parameter.

dlzhry2:
Looks like we might be deferring work in this area until after PoC. Will close once confirmed.

dlzhry2:
Confirmed with BAs that we are closing this. Will keep the branch, in case required for a write up.

PR #470, opened 2025-04-15
DHSCFT-652: Only show heatmap when all areas in group selected

dlzhry2:
# Description

Jira ticket: [DHSCFT-652](https://bjss-enterprise.atlassian.net/browse/DHSCFT-652)

We had not coded for this scenario and this was highlighted in the demo.

## Changes

- Added extra logic to check for the `SearchParams.GroupAreaSelected` value and render accordingly
- Added additional unit test for this scenario

## Validation

Tested locally and also ran a regression checking all the other scenarios documented in the ticket.

### DIFF
--- a/frontend/fingertips-frontend/components/viewPlots/TwoOrMoreIndicatorsAreasViewPlots/index.tsx
+++ b/frontend/fingertips-frontend/components/viewPlots/TwoOrMoreIndicatorsAreasViewPlots/index.tsx
@@ -18,25 +18,23 @@ import {
 import { determineAreaCodes } from '@/lib/chartHelpers/chartHelpers';
 import { ALL_AREAS_SELECTED } from '@/lib/areaFilterHelpers/constants';
 
-interface ComponentRenderFlags {
-  showHeatmap: boolean;
-  showSpineChart: boolean;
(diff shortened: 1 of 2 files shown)


frontend/fingertips-frontend/components/viewPlots/TwoOrMoreIndicatorsAreasViewPlots/index.tsx
```diff
@@ -11,8 +11,33 @@ import { IndicatorDocument } from '@/lib/search/searchTypes';
...
+  const areAllAreasSelected = groupAreaSelected === ALL_AREAS_SELECTED;
+
+  return {
+    showHeatmap: areaCodes.length > 1 || areAllAreasSelected,
+    showSpineChart:
+      areaCodes.length < 3 &&
+      spineChartIndicatorData.length > 0 &&
+      !areAllAreasSelected,
+  };
+}
```
stuc-bjss, line 40: would it be clearer to split this into two functions each one returning a boolean...
shouldShowSpineChart(...)
shouldShowHeatmap(...)
or similar?

stuc-bjss (approved):
left a suggestion, but it's subjective

PR #473, opened 2025-04-16
Bugfix/dhscft 644 error bar state

priya-thorpe-bjss:
# Description

Jira ticket: [DHSCFT-644](https://bjss-enterprise.atlassian.net/browse/DHSCFT-644)

Bug fix for line chart and inequalities trend chart legends.

## Changes

- Render confidence interval bars conditionally instead of them always being present.

## Validation
 Manual validation

### DIFF
--- a/frontend/fingertips-frontend/components/organisms/Inequalities/inequalitiesHelpers.test.ts
+++ b/frontend/fingertips-frontend/components/organisms/Inequalities/inequalitiesHelpers.test.ts
@@ -498,24 +498,8 @@ describe('generateLineChartSeriesData', () => {
     dashStyle: 'Solid',
   };
 
-  const confidenceIntervalSeries = {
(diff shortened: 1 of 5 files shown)


alibombaye (approved):

dlzhry2 (commented):
Looks good to me. Would suggest we add at least one test in `lineChartHelpers.test.ts` to test `generateSeriesData` when true is passed for showing the confidence intervals.

dlzhry2 (approved):

dlzhry2:
> Looks good to me. Would suggest we add at least one test in `lineChartHelpers.test.ts` to test `generateSeriesData` when true is passed for showing the confidence intervals.

Checked with @priya-thorpe-bjss - although there is not a test for this shown in the diff, there is sufficient existing tests in the file that test this particular condition.

PR #474, opened 2025-04-16
bugfix/DHSCFT-657: quintile colours on map and embedded bar

stuc-bjss:
# Description

Jira ticket: [DHSCFT-657](https://bjss-enterprise.atlassian.net/browse/DHSCFT-657)

## Changes

- Updated swagger to remove props no longer used
- Updated quintiles comparison stored proc
- Added colour override for embedded bar chart

## Validation

image
image


dlzhry2 (commented):
Looks good - just a test that needs updating called `check No Judgement - current site doesn't`. Looks like there's two failing assertions due to the expected data not having been updated.

PR #482, opened 2025-04-17
DHSCFT-TECH: Improve API test output

dlzhry2:
# Description

Jira ticket: [DHSCFT-661](https://bjss-enterprise.atlassian.net/browse/DHSCFT-661)

See linked pipeline in ticket. We found that the output of the API test is truncated and it is also very difficult to identify what actually went wrong. This PR aims to improve this.

## Changes

- Updates the command that is run in the pipeline. The command will now 1. only show the failed output and overall summary (much less verbose) 2. show the full failing API response
- Adds a couple of improvements to one test file to ensure that errors are surfaced correctly
- Adds details to the Readme to make the usage and intent of httpyac clearer

## Validation

I created some failing tests so we can validate that A. the pipeline fails and B. the output is improved.

You can see the results of that pipeline run here: https://github.com/dhsc-govuk/FingertipsNext/actions/runs/14513047508?pr=482

I then reverted the failing pipeline and ensures that it passes on this PR.

Finally, I also verified that the command runs as expected locally.

### DIFF
--- a/.github/workflows/e2e-and-api-tests.yml
+++ b/.github/workflows/e2e-and-api-tests.yml
@@ -67,7 +67,7 @@ jobs:
       - name: Run .http API tests
         working-directory: ./api
         run: |
-          npx httpyac send ./**/*.http --all --json --output-failed short --output short | tee httpyacresults.json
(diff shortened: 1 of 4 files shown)


Joe-Hendley (approved):

stuc-bjss (approved):
SO MUCH BETTER!

PR #485, opened 2025-04-17
DHSCFT-642: fix bug with using incorrect areaType being used

alibombaye:
# Description

Jira ticket: [DHSCFT-642](https://bjss-enterprise.atlassian.net/browse/DHSCFT-642)

Fixes a bug when selecting the areaType would result in the chart not being shown for when the areaSelected is defaulted to england.
Screenshot 2025-04-17 at 13 24 32

Updates logic when an indicator and an area has insufficient health data points.
Screenshot 2025-04-17 at 13 25 18
Screenshot 2025-04-17 at 13 27 16
^ just one health data point 

## Validation

Tested manually


frontend/fingertips-frontend/components/organisms/ViewsWrapper/hasSufficientHealthDataCheck.ts
```diff
@@ -1,34 +1,43 @@
...
+          (areaSelected) => areaSelected === areaHealthData.areaCode
         );
 
         if (hasAreaSelectedHealthData) {
-          return true;
+          const hasSufficientAreaHealthData =
+            indicatorsRequested === 1 && areasSelected.length <= 2
+              ? areaHealthData.healthData.length > 1
+              : true;
+          if (hasSufficientAreaHealthData) return true;
```
steves-bjss, line 20: Do we need to check the status of hasSufficentAreaHealthData? - we could just return that value or skip the line as it will be returned on line 22.

frontend/fingertips-frontend/components/organisms/ViewsWrapper/hasSufficientHealthDataCheck.ts
```diff
@@ -1,34 +1,43 @@
 import { IndicatorWithHealthDataForArea } from '@/generated-sources/ft-api-client';
 
-const hasAreaHealthData = (
+const hasSufficientAreaHealthData = (
```
steves-bjss, line 3: The same variable name is used on line 3 and line 16 - could we choose an alternative to make things clearer.

frontend/fingertips-frontend/components/organisms/ViewsWrapper/hasSufficientHealthDataCheck.ts
alibombaye: We want to only overwrite the state of `hasSufficientAreaHealthData` when we have sufficient health data for the area. If we remove this check we overwrite the state and this test then fails;
`should return true if the indicator has sufficient health data for at least one of the areaSelected`
The test here has 2 areas with the last one having insufficient health data. We then incorrectly set the entire state to false when it should be true since the first area had sufficient health data.

frontend/fingertips-frontend/components/organisms/ViewsWrapper/hasSufficientHealthDataCheck.ts
alibombaye: Yeah updating naming has helped. Good spot 👍

frontend/fingertips-frontend/components/views/OneIndicatorOneAreaView/index.tsx
```diff
@@ -43,6 +44,11 @@ export default async function OneIndicatorOneAreaView({
   await connection();
   const indicatorApi = ApiClientFactory.getIndicatorsApiClient();
 
+  const areaTypeToUse =
+    areaCodes.length === 1 && areaCodes[0] === areaCodeForEngland
+      ? englandAreaType.key
+      : areaTypeSelected;
```
gareth-allan-bjss, line 50: I think, given the block above that will throw an error if `areaCodes.length !==1`, we can assume we only have a single area code in this view.
```suggestion
  const areaTypeToUse =
    areaCodes[0] === areaCodeForEngland
      ? englandAreaType.key
      : areaTypeSelected;
```

frontend/fingertips-frontend/components/organisms/ViewsWrapper/hasSufficientHealthDataCheck.ts
```diff
@@ -0,0 +1,47 @@
...
+    return indicatorDataForArea.areaHealthData.reduce<boolean>(
+      (hasSufficientAreaHealthData, areaHealthData) => {
+        const healthDataForSelectedArea = areasSelected?.find(
+          (areaSelected) => areaSelected === areaHealthData.areaCode
+        );
+
+        if (healthDataForSelectedArea) {
+          const hasSufficientHealthData =
+            indicatorsRequested === 1 && areasSelected.length <= 2
+              ? areaHealthData.healthData.length > 1
```
gareth-allan-bjss, line 18: This is because we need to display a line chart for 1 indicator/1 area and 1 indicator/2 areas, and we need > 1 data point for a line chart, right? Might be worth a comment/extracting to a nicely named function to make it clear what's going on, as it took me a while to work through it.

frontend/fingertips-frontend/components/organisms/ViewsWrapper/hasSufficientHealthDataCheck.ts
```diff
@@ -13,10 +13,15 @@ const checkAreaHasSufficientHealthData = (
         );
 
         if (healthDataForSelectedArea) {
-          const hasSufficientHealthData =
-            indicatorsRequested === 1 && areasSelected.length <= 2
-              ? areaHealthData.healthData.length > 1
-              : true;
+          // We only return a single and the most recent health data point when more that one indicator or 2 areas have have been selected.
+          // Therefore we can only check there is sufficient data when there is one indicator and 2 or less areas selected.
+          const hasOnlyLatestHealthData =
```
gareth-allan-bjss, line 18: Perhaps `requiresOnlyLatestHealthData` is slightly clearer?
```suggestion
          const requiresOnlyLatestHealthData =
```

frontend/fingertips-frontend/components/organisms/ViewsWrapper/hasSufficientHealthDataCheck.test.ts
```diff
@@ -0,0 +1,251 @@
...
+      const result = hasSufficientHealthDataCheck(
+        indicatorsToCheck,
+        areasSelected
+      );
+
+      expect(result).toBe(false);
+    });
+  });
+
+  describe('When there are multiple indicators', () => {
```
gareth-allan-bjss, line 188: I think you're now missing the `should return false if all the indicator has no health data for all of the areaSelected` case that used to exist in this describe block, aren't you?

gareth-allan-bjss (approved):

PR #488, opened 2025-04-17
Bugfix/dhscft 663 suppress intentional error logging

Joe-Hendley:
part of https://bjss-enterprise.atlassian.net/browse/DHSCFT-663

`jest.SpyOn(...).mockImplementation` only mocks in individual tests rather than for the whole suite, so this approach shouldn't be problematic if actual errors occur

### DIFF
--- a/frontend/fingertips-frontend/lib/search/indicatorSearchService.test.ts
+++ b/frontend/fingertips-frontend/lib/search/indicatorSearchService.test.ts
@@ -275,11 +275,13 @@ describe('IndicatorSearchService', () => {
     mockGetDocument.mockImplementation(async () => {
       throw new Error('some error');
     });
+    const spyLog = jest.spyOn(console, 'error').mockImplementation(() => {});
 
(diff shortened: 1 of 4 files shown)


frontend/fingertips-frontend/lib/search/indicatorSearchService.test.ts
```diff
@@ -275,11 +275,13 @@ describe('IndicatorSearchService', () => {
     mockGetDocument.mockImplementation(async () => {
       throw new Error('some error');
     });
+    const spyLog = jest.spyOn(console, 'error').mockImplementation(() => {});
 
     const searchService = SearchServiceFactory.getIndicatorSearchService();
     const result = await searchService.getIndicator('123');
 
     expect(result).toBeUndefined();
+    expect(spyLog).toHaveBeenCalled();
```
dlzhry2, line 284: Could maybe assert what it is called with?

dlzhry2 (approved):

alibombaye (approved):

PR #494, opened 2025-04-22
DHSCFT-662: Improve get latest data functionality

dlzhry2:
# Description

Jira ticket: [DHSCFT-662](https://bjss-enterprise.atlassian.net/browse/DHSCFT-662)

Ensure we return data for the latest year when requested, scoped to the area codes provided by the client and excluding the latest date for England.

## Changes

- Upped the chunking limit to 300 for both the indicator data and areas API endpoints. This means that the largest collection i.e. all districts for England should not be chunked up which avoids the issue (edge case) where a set of areas might have a different latest year to another. (Due to the way data is collected this is unlikely but worth guarding against).
- Extended backend repository method so the latest year will be considering the area codes provided and will exclude England for the reasons articulated on the ticket.
- Added required additional tests for the backend and ensured UI tests continue to pass.

## Validation

Local manual testing. E2e tests continue to pass.

Validated the scenario described in the ticket in addition to a variety of other indicators. Re the scenario in the ticket, there is currently no change - it is not just an issue with the year for England in this situation - there is no data at all for any of the regions for any year.

Do we still want to show the empty areas and an outline to the map if not a single area has any data or would we want a different behaviour?

Regarding further scenarios, one example would be to compare chart?si=383&is=383&ats=districts-and-unitary-authorities&gts=regions&gs=E12000002&gas=ALL on local vs. deployed. The data will now be surfaced rather than empty.

QUESTION: the benchmark still shows the latest data for England, would we want to switch this to be the latest data that is being used for the areas? If so, part of this ticket or a follow on?

### DIFF
--- a/api/DHSC.FingertipsNext.Modules.HealthData.Repository/HealthDataRepository.cs
+++ b/api/DHSC.FingertipsNext.Modules.HealthData.Repository/HealthDataRepository.cs
@@ -9,16 +9,19 @@ namespace DHSC.FingertipsNext.Modules.HealthData.Repository;
 [SuppressMessage("ReSharper", "SimplifyConditionalTernaryExpression")]
 public class HealthDataRepository(HealthDataDbContext healthDataDbContext) : IHealthDataRepository
 {
+    private const string ENGLAND_AREA_CODE = "E92000001";
     private const string SEX = "sex";
     private const string AGE = "age";
     private const string DEPRIVATION = "deprivation";
 
     private readonly HealthDataDbContext _dbContext = healthDataDbContext ?? throw new ArgumentNullException(nameof(healthDataDbContext));
 
-    public async Task<IndicatorDimensionModel> GetIndicatorDimensionAsync(int indicatorId)
+    public async Task<IndicatorDimensionModel> GetIndicatorDimensionAsync(int indicatorId, string[] areaCodes)
     {
         var model = await _dbContext.HealthMeasure
            .Where(healthMeasure => healthMeasure.IndicatorDimension.IndicatorId == indicatorId)
+           .Where(healthMeasure => areaCodes.Length == 0 || EF.Constant(areaCodes).Contains(healthMeasure.AreaDimension.Code))
+           .Where(healthMeasure => areaCodes.Length == 1 || healthMeasure.AreaDimension.Code != ENGLAND_AREA_CODE)
            .OrderByDescending(healthMeasure => healthMeasure.Year)
            .Include(healthMeasure => healthMeasure.IndicatorDimension)
            .Select(healthMeasure => new IndicatorDimensionModel
--- a/api/DHSC.FingertipsNext.Modules.HealthData.Repository/HealthDataRepository.cs
+++ b/api/DHSC.FingertipsNext.Modules.HealthData.Repository/HealthDataRepository.cs
@@ -35,8 +35,28 @@ public class HealthDataRepository(HealthDataDbContext healthDataDbContext) : IHe
            })
            .Take(1)
            .FirstOrDefaultAsync();
-       
-        return model;
+        
+        if (model != null) {
+            return model;
+        }
+
+        // Default to the latest year for England if none of the requested areas have data
+        return await _dbContext.HealthMeasure
+           .Where(healthMeasure => healthMeasure.IndicatorDimension.IndicatorId == indicatorId)
+           .Where(healthMeasure => healthMeasure.AreaDimension.Code == ENGLAND_AREA_CODE)
+           .OrderByDescending(healthMeasure => healthMeasure.Year)
+           .Include(healthMeasure => healthMeasure.IndicatorDimension)
+           .Select(healthMeasure => new IndicatorDimensionModel
+           {
+               IndicatorId = healthMeasure.IndicatorDimension.IndicatorId,
+               IndicatorKey = healthMeasure.IndicatorKey,
+               Name = healthMeasure.IndicatorDimension.Name,
+               Polarity = healthMeasure.IndicatorDimension.Polarity,
+               BenchmarkComparisonMethod = healthMeasure.IndicatorDimension.BenchmarkComparisonMethod,
+               LatestYear = healthMeasure.Year
+           })
+           .Take(1)
+           .FirstOrDefaultAsync();;
     }
     
     public async Task<IEnumerable<HealthMeasureModel>> GetIndicatorDataAsync(int indicatorId, string[] areaCodes, int[] years, string[] inequalities)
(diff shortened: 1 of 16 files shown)


api/DHSC.FingertipsNext.Modules.HealthData.Repository/HealthDataRepository.cs
```diff
@@ -9,16 +9,19 @@ namespace DHSC.FingertipsNext.Modules.HealthData.Repository;
...
 
     private readonly HealthDataDbContext _dbContext = healthDataDbContext ?? throw new ArgumentNullException(nameof(healthDataDbContext));
 
-    public async Task<IndicatorDimensionModel> GetIndicatorDimensionAsync(int indicatorId)
+    public async Task<IndicatorDimensionModel> GetIndicatorDimensionAsync(int indicatorId, string[] areaCodes)
     {
         var model = await _dbContext.HealthMeasure
            .Where(healthMeasure => healthMeasure.IndicatorDimension.IndicatorId == indicatorId)
+           .Where(healthMeasure => areaCodes.Length == 0 || EF.Constant(areaCodes).Contains(healthMeasure.AreaDimension.Code))
+           .Where(healthMeasure => areaCodes.Length == 1 || healthMeasure.AreaDimension.Code != ENGLAND_AREA_CODE)
```
dlzhry2, line 24: Condition for scenario where the only area code requested is England.

HowardMayBjss, line 24: I'm struggling to understand the intent of this line. If only one area is requested then exclude data for England ... why is this?

dlzhry2: If we request only one area code and that area code is England, then without `areaCodes.Length == 1` it's going to find no indicator data and no latest year because of the `!= ENGLAND_AREA_CODE`. There is a unit test for this too.

dlzhry2: In that scenario we will grab the latest year for England. But in all other scenarios we will want to exclude England from the calculation of the latest year. You can test but this works for no area codes provided, multiple incl. and excl. England and one area code provided where it is England and where it is not England.

However, in Slack I have documented one scenario that is not optimal and I believe we would need to deal with: all area codes in the request (besides England) do not actually contain data.

More widely, this method also exposes the fact that I think it is doing too much. If I were coming into the codebase fresh `GetIndicatorDataAsync` it would surprise me that it is not a simple method that queries an indicator table by a primary key.

api/DHSC.FingertipsNext.Modules.HealthData.Repository/HealthDataRepository.cs
```diff
@@ -9,16 +9,41 @@ namespace DHSC.FingertipsNext.Modules.HealthData.Repository;
...
+               LatestYear = healthMeasure.Year
+           })
+           .Take(1)
+           .FirstOrDefaultAsync();
+        
+        if (model != null) {
+            return model;
+        }
+
+        // Default to the latest year for England if none of the requested areas have data
```
dlzhry2, line 43: Yuck! I would like to refactor this as per Slack thread: https://bjss.slack.com/archives/C08335HN952/p1745327438314749

api/DHSC.FingertipsNext.Modules.HealthData.Repository/HealthDataRepository.cs
HowardMayBjss: Yeah - the intent of this function is not reflected by the name of the function - If I understand correctly it is both getting the IndicatorDimension and ALSO caluculating the latest year for which there is data ... according to some complex rules

dlzhry2: Spot on

frontend/fingertips-frontend/app/chart/PopulationPyramidWithTableDataProvider/PopulationPyramidWithTableDataProvider.test.tsx
```diff
@@ -76,7 +76,7 @@ describe('PopulationPyramidWithTableDataProvider', () => {
         results.push(area);
       }
       return results;
-    })(130);
+    })(330);
```
williamwillis11, line 79: can this be based off of maxNumAreasThatCanBeRequestedAPI?

frontend/fingertips-frontend/lib/ViewsHelpers.test.ts
```diff
@@ -51,7 +51,7 @@ describe('getHealthDataForIndicator', () => {
   });
 
   it('should make the appropriate number of API calls when a long list of areas is requested', async () => {
-    const testAreas = new Array(101).fill('a', 0, 101);
+    const testAreas = new Array(301).fill('a', 0, 301);
```
williamwillis11, line 54: can this be based off of maxNumAreasThatCanBeRequestedAPI?

frontend/fingertips-frontend/lib/areaFilterHelpers/getSelectedAreasData.test.tsx
```diff
@@ -121,15 +121,15 @@ describe('getSelectedAreasDataByAreaType', () => {
 
     mockAreasApi.getAreas.mockResolvedValue([area1, area2]);
 
-    const testAreas = new Array(101).fill('a');
+    const testAreas = new Array(301).fill('a');
```
williamwillis11, line 124: can this be based off of maxNumAreasThatCanBeRequestedAPI?

frontend/fingertips-frontend/app/chart/PopulationPyramidWithTableDataProvider/PopulationPyramidWithTableDataProvider.test.tsx
dlzhry2: Actioned - please check and the others below.

api/DHSC.FingertipsNext.Modules.HealthData.Repository/HealthDataRepository.cs
```diff
@@ -9,16 +9,51 @@ namespace DHSC.FingertipsNext.Modules.HealthData.Repository;
...
 
     private readonly HealthDataDbContext _dbContext = healthDataDbContext ?? throw new ArgumentNullException(nameof(healthDataDbContext));
 
-    public async Task<IndicatorDimensionModel> GetIndicatorDimensionAsync(int indicatorId)
+    /// <summary>
+    /// Will retrieve the indicator dimension data for the requested indicator ID while also
+    /// obtaining the latest year that data is available for the indicator based on the area codes provided.
+    /// 
+    /// Note: both in terms of intent and performance this is sub-optimal and should be refactored at the earliest opportunity
+    /// post-POC.
```
MonkeyDate, line 24: is there a ticket we could reference here to cover that future work?

dlzhry2: I was planning to do it straight after/after I've finished some Trend App documentation. I'll create a ticket though and link it up - it's a good shout.

dlzhry2: Done, I've tagged it up with a ticket and also fixed the issue I had with the predicates.

HowardMayBjss (approved):

PR #497, opened 2025-04-22
Update DataCreator paths to work on Mac

gareth-allan-bjss:
# Description

Jira ticket: None.

This PR makes updates to the DataCreator application so that it can be run on Mac.

## Changes

- Replaced string literal paths with use of [`Path.Join`](https://learn.microsoft.com/en-us/dotnet/api/system.io.path.join?view=net-9.0), so they are not always separated with `\`

## Validation

Deleted the contents of the `DataCreator/DataCreator/data/out` directory, ran the application using the latest Pholio DB export and without changing the other input data files, then verified using git that the files produced in `DataCreator/DataCreator/data/out` were unchanged.

@HowardMayBjss has also verified that he is able to perform the above verification on Windows.

### DIFF
--- a/DataCreator/DataCreator/DataFileReader.cs
+++ b/DataCreator/DataCreator/DataFileReader.cs
@@ -6,12 +6,12 @@ namespace DataCreator
     public static class DataFileReader
     {
         private const string Persons = "Persons";
-        private static readonly string InFilePath = Path.Combine("..","..","..","data","in");
-        private static readonly string TempDirPath = Path.Combine(InFilePath, "temp");
+        private static readonly string InFilePath = Path.Join("..","..","..","data","in");
+        private static readonly string TempDirPath = Path.Join(InFilePath, "temp");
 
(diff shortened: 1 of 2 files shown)


HowardMayBjss (approved):

steves-bjss (approved):

gareth-allan-bjss:
> Just to check as not completely clear from the PR description, the output folder in the data creator was checked.
> 
> I assume we also checked the file that was published to some of the asset folders in other directories e.g. trend analysis and search setup?

I used git to verify that no files have been changed. I've just tried again, and made sure to delete the following files before running, to verify they're re-created as expected:
- `search-setup/assets/areas.json`
- `search-setup/assets/indicators.json`
- `trend-analysis/TrendAnalysisApp/SearchData/assets/indicators.json`

Re-running the application after doing this generates the above files and Git reports them as unchanged.

dlzhry2 (approved):

PR #499, opened 2025-04-23
Feature/dhscft 679 align benchmark and group year linechart

WolfieKnee:
# Description

Jira ticket: [DHSCFT-679](https://bjss-enterprise.atlassian.net/browse/DHSCFT-679)/[DHSCFT-601](https://bjss-enterprise.atlassian.net/browse/DHSCFT-601)

Determine first and last year of data for selected areas and filter table data/limit chart axis to only show these years.

## Changes
- update utils functions to find first & last years
- filter table data for years within those limits
- set axis max/min to those years

## Validation
updated behaviour as expected against local api:
http://localhost:3000/chart?is=241&as=E38000231&ats=nhs-sub-integrated-care-boards&gts=england&gs=E92000001
image
image

compared with current deployment:
http://20.26.160.30/chart?is=241&as=E38000231&ats=nhs-sub-integrated-care-boards&gts=england&gs=E92000001
image
image

### DIFF
--- a/frontend/fingertips-frontend/components/organisms/LineChartTable/index.tsx
+++ b/frontend/fingertips-frontend/components/organisms/LineChartTable/index.tsx
@@ -25,7 +25,11 @@ import {
 } from '@/lib/tableHelpers';
 import { BenchmarkLabel } from '@/components/organisms/BenchmarkLabel';
 import { TrendTag } from '@/components/molecules/TrendTag';
-import { getConfidenceLimitNumber } from '@/lib/chartHelpers/chartHelpers';
+import {
+  getConfidenceLimitNumber,
+  getFirstYearForAreas,
+  getMostRecentYearForAreas,
+} from '@/lib/chartHelpers/chartHelpers';
 import { formatNumber, formatWholeNumber } from '@/lib/numberFormatter';
 import { areaCodeForEngland } from '@/lib/chartHelpers/constants';
 
@@ -230,7 +234,16 @@ export function LineChartTable({
     ...(groupIndicatorData?.healthData ?? []),
     ...healthIndicatorData.flatMap((area) => area.healthData),
   ].map(({ year }) => year);
-  const allYears = [...new Set(allHealthPointYears)].sort((a, b) => a - b);
+
+  const firstYear = getFirstYearForAreas(healthIndicatorData);
+  const lastYear = getMostRecentYearForAreas(healthIndicatorData);
+  if (!firstYear || !lastYear) {
+    throw new Error('no data for any years');
+  }
+
+  const allYears = [...new Set(allHealthPointYears)]
+    .filter((year) => year >= firstYear && year <= lastYear)
+    .sort((a, b) => a - b);
 
   const rowData = allYears
     .map((year) => {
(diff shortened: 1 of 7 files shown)


frontend/fingertips-frontend/components/organisms/LineChartTable/index.tsx
```diff
@@ -230,7 +234,16 @@ export function LineChartTable({
     ...(groupIndicatorData?.healthData ?? []),
     ...healthIndicatorData.flatMap((area) => area.healthData),
   ].map(({ year }) => year);
-  const allYears = [...new Set(allHealthPointYears)].sort((a, b) => a - b);
+
+  const firstYear = getFirstYearForAreas(healthIndicatorData);
+  const lastYear = getLatestYearForAreas(healthIndicatorData);
+  if (!firstYear || !lastYear) {
+    throw new Error('no data for any years');
```
stuc-bjss, line 241: if we show an error, where does it get caught, or does it just kill the entire page?

dlzhry2, line 241: I assume this would be pre-validated before it comes to rendering the chart. At least when hitting an area I know that does not have health data (indicator 383) I get a message stating that there is no data: http://20.26.160.30/chart?si=383&is=383&as=E12000004&ats=regions&gts=england&gs=E92000001

Still, happy for us to have this message - just wanted to check we get the same behaviour locally as we do on the deployed site - could you give that above check a spin?

dlzhry2: Additionally, might be nice if we mention the name of the indicator in question that caused the error. As above, hopefully it's not something that would be called, but if it did it would be useful to have some more information to aid investigation.

WolfieKnee: Good point, this will cause the page to hit the error page, but it'd be better to just not show the component (e.g. `return null`?)

WolfieKnee: http://localhost:3000/chart?si=383&is=383&as=E12000004&ats=regions&gts=england&gs=E92000001 does hit the no-data page before throwing this error: 
image

frontend/fingertips-frontend/lib/chartHelpers/chartHelpers.ts
```diff
@@ -193,7 +193,18 @@ export function getLatestYear(
...
 }
 
-function getMostRecentYearForAreas(
+export function getFirstYear(
+  points: HealthDataPoint[] | undefined
+): number | undefined {
+  if (!points || points.length < 1) return undefined;
+
+  const year = points.reduce((previous, point) => {
+    return Math.min(previous, point.year);
```
dlzhry2, line 202: Think you could just have the arrow function rather than {} and a return statement.

stuc-bjss (approved):

frontend/fingertips-frontend/components/organisms/LineChartTable/index.tsx
dlzhry2: Wait, so does it still throw that error? Or does it not even reach the code because some sort of parent does not render the component if there is no data? Either way that's good :)

dlzhry2 (approved):

frontend/fingertips-frontend/components/organisms/LineChartTable/index.tsx
dlzhry2: Seems not to reach that code when testing locally. Marking as resolved.

PR #500, opened 2025-04-23
Bugfix/dhscft 648 error bar visibility

stuc-bjss:
# Description

Jira ticket: [DHSCFT-648](https://bjss-enterprise.atlassian.net/browse/DHSCFT-648)

## Changes

- Link React state and high charts events click
- Involves ugly mutation of HighCharts config options object
- Create a link between errorBars and their respective data series
- Brings state management of show/high CI bars into the Line chart component

### DIFF
--- a/frontend/fingertips-frontend/components/molecules/Inequalities/InequalitiesTrend/index.tsx
+++ b/frontend/fingertips-frontend/components/molecules/Inequalities/InequalitiesTrend/index.tsx
@@ -13,7 +13,6 @@ import {
   SearchStateManager,
   SearchStateParams,
 } from '@/lib/searchStateManager';
-import { useState } from 'react';
 import {
   generateInequalitiesLineChartOptions,
   getDynamicKeys,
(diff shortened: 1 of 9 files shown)


frontend/fingertips-frontend/components/molecules/Inequalities/InequalitiesTrend/index.tsx
```diff
@@ -186,7 +180,7 @@ export function InequalitiesTrend({
       lineChartData,
       dynamicKeys,
       type,
-      showInequalitiesLineChartConfidenceIntervals,
+      true,
```
MonkeyDate, line 183: why is this now always true?

stuc-bjss: because the decision to show or hide those error bars happens within the line chart component, and so they have to be generated in the first place.

frontend/fingertips-frontend/components/organisms/LineChart/lineChartHelpers.test.ts
```diff
@@ -368,3 +375,85 @@ describe('generateStandardLineChartOptions', () => {
...
+      undefined,
+      true
+    );
+
+    setVisibility = jest.fn() as Dispatch<
+      SetStateAction<Record<string, boolean>>
+    >;
+  });
+
+  it('should add add show and hide functions which call setVisibility (state)', () => {
```
dlzhry2, line 398:
```suggestion
  it('should add show and hide functions which call setVisibility (state)', () => {
```

PR #510, opened 2025-04-23
DHSCFT-682 compare areas table sort order

MonkeyDate:
# Description

Jira ticket: [DHSCFT-682](https://bjss-enterprise.atlassian.net/browse/DHSCFT-682)

Compare areas table should sort in consistent order when rows have the same value.

## Changes

- Sort table by value, then by area name
- Update a bunch of unit tests to use act() to reduce unit test noise.

### DIFF
--- a/frontend/fingertips-frontend/components/organisms/BarChartEmbeddedTable/barChartEmbeddedTable.test.tsx
+++ b/frontend/fingertips-frontend/components/organisms/BarChartEmbeddedTable/barChartEmbeddedTable.test.tsx
@@ -1,4 +1,4 @@
-import { render, screen } from '@testing-library/react';
+import { act, render, screen } from '@testing-library/react';
 import { BarChartEmbeddedTable } from '@/components/organisms/BarChartEmbeddedTable/index';
 import {
   HealthDataForArea,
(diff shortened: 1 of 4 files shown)


frontend/fingertips-frontend/components/organisms/BarChartEmbeddedTable/barChartEmbeddedTable.test.tsx
```diff
@@ -222,56 +242,98 @@ describe('BarChartEmbeddedTable', () => {
...
+    const valueCells = areaRows.map((areaRow) => {
+      const cellsInRow = within(areaRow).getAllByRole('cell');
+      return cellsInRow[valueColumnIndex].textContent;
+    });
+
+    expect(valueCells).toEqual(expectedValues);
+  });
+
+  it('should order the data displayed by by are name is values match', async () => {
+    const mockData = cloneDeep(mockHealthIndicatorData);
```
stuc-bjss, line 272: I'd prefer not to add another dependency library just to do this.
You could make a deepClone function e.g.
`const cloneDeep = (input) => JSON.parse(JSON.stringify(input));`
you might need try catch but in test env is less important.

or 
```
const mockData = mockHealthIndicatorData.map(row => {
return { ...row, healthData: row.healthData.map((point) => ({...point, value: 186.734 })};
})
```

frontend/fingertips-frontend/package.json
```diff
@@ -85,6 +86,7 @@
     "jest-environment-jsdom": "^29.7.0",
     "jest-mock-extended": "^4.0.0-beta1",
     "jest-styled-components": "^7.2.0",
+    "lodash": "^4.5.0",
```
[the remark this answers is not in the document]
MonkeyDate: removed

frontend/fingertips-frontend/components/organisms/BarChartEmbeddedTable/barChartEmbeddedTable.test.tsx
MonkeyDate: using json version

PR #513, opened 2025-04-24
DHSCFT-659 - Add test for DHSCFT-689

Nayira-BJSS:
# Description

Jira ticket: [DHSCFT-659](https://bjss-enterprise.atlassian.net/browse/DHSCFT-659)

Add new e2e test for One Indicator/England Only area scenario

## Changes

- Added new scenario to core journey pack
- Made small refactor to getScenarioConfig to reduce cognitive complexity
- Updated screenshots to add new ones relevant to new test

## Validation

New screenshots added for the new e2e tests. Existing test still working.


frontend/fingertips-frontend/playwright/testHelpers.ts
```diff
@@ -188,118 +188,77 @@ export function getScenarioConfig(
...
-    visibleComponents = allComponents.filter((component) =>
-      [
-        ChartPage.spineChartTableComponent,
-        ChartPage.heatMapComponent,
-        ChartPage.populationPyramidComponent,
-      ].includes(component.componentLocator)
-    );
-  } else {
+  const visibleComponentMap: Record<string, string[]> = {
+    [`${IndicatorMode.ONE_INDICATOR}-${AreaMode.ONE_AREA}`]: [
```
williamwillis11, line 194: could we use something more type safe like the enums directly instead of building a string from them?

williamwillis11: tbh is it crazy to be wondering if we need to add a unit test file for this function?

williamwillis11: ignore me :)

PR #514, opened 2025-04-24
DHSCFT-694: Show spine chart row when an area is missing

stuc-bjss:
# Description

Jira ticket: [DHSCFT-694](https://bjss-enterprise.atlassian.net/browse/DHSCFT-694)

## Changes

- Include spine chart row in table even if an area is missing data
- Pull out the show two areas so that the row is told rather than deduce from the data, as if the data is missing it would render alternate layout for a single area

## Validation

Before row missing...

image

Showing the row but not forcing the row into multiple area rendering view...
image

After, note the second row is showing but has X for Derbyshire and only shows one marker on the spine
image...

When the only area is missing
image

### DIFF
--- a/frontend/fingertips-frontend/components/organisms/SpineChartTable/spineChartTableHelpers.tsx
+++ b/frontend/fingertips-frontend/components/organisms/SpineChartTable/spineChartTableHelpers.tsx
@@ -90,7 +90,7 @@ export const buildSpineChartIndicatorData = (
       if (!areasHealthData[0]) {
         // there was missing data for an area
         console.warn(`No area data`);
-        return null;
+        // return null;
       }
 
       const matchedQuartileData = quartileData.find(
@@ -116,8 +116,8 @@ export const buildSpineChartIndicatorData = (
         benchmarkComparisonMethod: indicatorData.benchmarkMethod,
         // The latest period for the first area's data (health data is sorted be year ASC)
         latestDataPeriod:
-          areasHealthData[0].healthData[
-            areasHealthData[0].healthData.length - 1
+          areasHealthData[0]?.healthData[
+            areasHealthData[0]?.healthData.length - 1
           ].year,
         areasHealthData,
         groupData,
--- a/frontend/fingertips-frontend/components/organisms/SpineChartTable/spineChartTableHelpers.tsx
+++ b/frontend/fingertips-frontend/components/organisms/SpineChartTable/spineChartTableHelpers.tsx
@@ -6,11 +6,7 @@ import {
 } from '@/generated-sources/ft-api-client';
 import { IndicatorDocument } from '@/lib/search/searchTypes';
 
-export const spineChartImproperUsageError =
-  'Improper usage: Spine chart should only be shown when 1-2 areas are selected';
 export const spineChartIndicatorTitleColumnMinWidth = 240;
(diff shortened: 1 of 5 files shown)


frontend/fingertips-frontend/components/organisms/SpineChartTable/spineChartTableHelpers.tsx
```diff
@@ -96,6 +100,7 @@ export const buildSpineChartIndicatorData = (
 
       if (!matchedQuartileData) {
         // No quartile data found for the requested indicator ID: ${indicatorData.indicatorId}
+        console.warn('No quartile data found');
```
dlzhry2, line 103: Did we want the above comment as the text within the console warning, or does it just need removing?

frontend/fingertips-frontend/components/organisms/SpineChartTable/spineChartTableHelpers.tsx
```diff
@@ -77,16 +76,21 @@ export const buildSpineChartIndicatorData = (
...
 
-      if (
-        areasHealthData.length !== areasSelected.length ||
-        !areasHealthData[0]
-      ) {
+      if (!areasHealthData[0]) {
         // there was missing data for an area
-        return null;
+        console.warn(`No area data`);
+        // return null;
```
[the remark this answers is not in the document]
dlzhry2: Do we even need this warning if we have the warning above anyway, in fact at line 84?

frontend/fingertips-frontend/components/organisms/SpineChartTable/spineChartTableHelpers.tsx
stuc-bjss: it's a good question - it seems like a genuine issue which might warrant a warning being produced - somewhere, is this the right way to surface it and pragmatic at this stage of the project? Open to suggestions

dlzhry2 (approved):
Looks good to me. Just a couple of minor comments about the error message and some commented out code.

frontend/fingertips-frontend/components/organisms/SpineChartTable/spineChartTableHelpers.tsx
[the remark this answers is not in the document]
stuc-bjss: it helps to debug whats going on, I'd parcel it with the question about the warnings. If we're ok with console.warning then I'd leave it in, obviously if not then they all get removed.

frontend/fingertips-frontend/components/organisms/SpineChartTable/spineChartTableHelpers.tsx
dlzhry2: @HowardMayBjss you worked on the backend changes for quartile data. Not sure who else to tag. This might be sufficient so far - my knowledge of Azure is more limited than AWS but I assume it'd be simple enough to direct our Frontend container logs to a monitoring service to alert us to such errors in future.

And simply raising a warning should be okay for now.

dlzhry2: Oh good point - of course, it's client side!

dlzhry2 (approved):

PR #515, opened 2025-04-24
DHSCFT-690 scroll to the middle of wide chart components

williamwillis11:
# Description

Jira ticket: [DHSCFT-690](https://bjss-enterprise.atlassian.net/browse/DHSCFT-690)

The current screenshot captured for the spineChartTable is not capturing most of the component as it is very wide. Most of the important/complex component is hidden off to the right of the screenshot capture. This PR is going to scroll to the middle of the the wide component so that we see the most of it.

## Changes

- add extra componentProps for wideComponent and set it to false for all except the spineChartTable
- for components that are wideComponent true we now scroll to the middle of that component before taking the screenshot 
- tried to ensure that the wide components (ie just the spine chart table) is checked last so that its scroll to middle doesnt impact the screenshots taken for other components by shuffling around the core journeys + visible component list + scenario configs however due to the tests running parallel this wasn't 100% successful as you can see in the PR - the key point is that these new baseline screenshots are stable and the same (just slightly larger)
- regenerated required screenshots - note some have correctly changed due to the new scroll for wideComponents (e.g. the spineChartTable)
- renamed OneAreaMultipleIndicatorsTableComponent and its supporting files/imports the to basicTableComponent to match other naming convention and https://ukhsa.atlassian.net/wiki/spaces/FTN/pages/171448117/Area+Indicator+journeys - this resulted in some newly named screenshots being created and old ones being deleted

## Validation

unit and e2e tests passing


frontend/fingertips-frontend/.test/spec/snaps/chromium/fully_integrated_e2e_tests/core_journeys.spec.ts/BOTH-SUBJECT-AND-AREA-then-select-ONE-INDICATOR-and-ONE-AREA-then-check-the-charts-page-inequalitiesBarChartTable-component.png
williamwillis11, line 1: @gareth-allan-bjss - I tried to ensure that the wide components (ie just the spine chart table) is checked last so that its scroll to middle doesnt impact the screenshots taken for other components by shuffling around the core journeys + visible component list + scenario configs however due to the tests running parallel this wasn't 100% successful as you can see in the PR - the key point is that these new baseline screenshots are stable

frontend/fingertips-frontend/jest.config.ts
```diff
@@ -11,7 +11,6 @@ const config: Config = {
   preset: 'ts-jest',
   testTimeout: 10000,
   coverageProvider: 'v8',
-  collectCoverage: true,
```
williamwillis11, line 14: forgot to remove this from the PR that added code coverage - @dlzhry2  thanks for pointing this out :)

frontend/fingertips-frontend/.test/spec/snaps/chromium/fully_integrated_e2e_tests/core_journeys.spec.ts/ONLY-AREA-then-select-TWO-INDICATORS-and-THREE-PLUS-AREAS-then-check-the-charts-page-spineChartTable-component.png
williamwillis11, line 1: this is better than it was and massively simpler than taking 2 screenshots which turned into a nightmare

frontend/fingertips-frontend/playwright/page-objects/pages/chartPage.ts
```diff
@@ -176,6 +175,19 @@ export default class ChartPage extends AreaFilter {
             .getByText('Show population data')
```
gareth-allan-bjss, line 175: Looks like the above line `.getByTestId('populationPyramidWithTable-component')` is only working because the pop pyramid is the only component with `hasDetailsExpander` set to `true`. Shouldn't it be using `visibleComponent.componentLocator` as you've done in the new code below?

gareth-allan-bjss (approved):
Looks good. A couple of suggestions, but no must-haves so I'm happy to approve.

frontend/fingertips-frontend/playwright/page-objects/pages/chartPage.ts
williamwillis11: nice spot

gareth-allan-bjss (approved):

PR #524, opened 2025-04-28
DHSCFT-693: Fix select all indicator bug for absent mock data 

adawari-thomas-bjss:
# Description

Jira ticket: [DHSCFT-693](https://bjss-enterprise.atlassian.net/browse/DHSCFT-693)

## Changes

- Removed non-existent indicators from mock AI search data
- Removed England filter so the dbContext defaults to returning the latest data for all data present within an indicator

## Validation

Selecting all results returned by the `age` search term, then viewing data returns the expected chart page

image


api/DHSC.FingertipsNext.Modules.HealthData.UnitTests/Repository/HealthDataRepositoryTests.cs
```diff
@@ -120,26 +120,26 @@ public async Task Repository_ShouldReturnCorrectIndicatorAndLatestYearWhenEnglan
...
+        PopulateDatabase(new HealthMeasureModelHelper(year: LATEST_YEAR)
             .WithIndicatorDimension(indicatorId: INDICATORID)
-            .WithAreaDimension(code: ENGLAND_AREA_CODE)
+            .WithAreaDimension(code: AREA_CODE)
             .Build());
 
         // act
         var result = await _healthDataRepository.GetIndicatorDimensionAsync(1, [
-            ENGLAND_AREA_CODE,
+            AREA_CODE,
```
dlzhry2, line 136: You can remove this line.

Previously, the `ENGLAND_AREA_CODE` could be passed in because the initial query will ignore the England area code. However, now we are no longer using "E92000001" and the fallback query checks all areas rather than just England.

Therefore, we should only provide `TESTAREA_ONE` and `TESTAREA_TWO`. Neither of these areas have data, so the fallback will check ALL data and should find the year for `AREA_CODE` "E92000002" and return 2024.

Hope this makes sense :)

dlzhry2 (commented):
Looks good - one minor change requested at the unit test level.

api/DHSC.FingertipsNext.Modules.HealthData.UnitTests/Repository/HealthDataRepositoryTests.cs
adawari-thomas-bjss: Ahh makes sense

dlzhry2 (approved):

gareth-allan-bjss (approved):

PR #533, opened 2025-05-01
DHSCFT-715 screenshots plan C - only create update screenshots in CI

williamwillis11:
# Description

Jira ticket: [DHSCFT-715](https://bjss-enterprise.atlassian.net/browse/DHSCFT-715)

Instead of creating screenshots locally using an npm script we will now create them in a CI job *only* - this is to achieve stability with the screenshots

## Changes

- Added new manual CI job to generate new screenshots if desired (made it so triggering this job only triggers the e2e tests so we dont waste time/resources
- Removed all screenshots in codebase
- Regenerated them in new CI job that generates the screenshots
- Change local e2e npm scripts to ignore screenshots
- update readme to reflect new dev flow
- update message thrown if screenshots dont match in CI job
- massively reduced the allowed pixel difference threshold to just 10 pixels
- removed spurious 'Send slack message if e2e tests fail' CI step in the e2e CI execution 

## Validation

Tested out positive and negative scenarios
 @alibombaye tested out the flow on a branched off PR (now closed) - https://github.com/dhsc-govuk/FingertipsNext/pull/534


.github/workflows/e2e-and-api-tests.yml
```diff
@@ -105,13 +137,6 @@ jobs:
             ${{ env.frontend-directory }}/test-results/*/*.png
           retention-days: 5
 
-      - name: Send slack message if e2e tests fail - for CD test execution
```
dlzhry2, line 108: Think I might have missed this and the rationale for the change. Would still want to get notifications if anything fails when run against the deployed env right? Keeps us honest.

.github/workflows/fingertips-workflow.yml
```diff
@@ -272,6 +278,7 @@ jobs:
     uses: ./.github/workflows/e2e-and-api-tests.yml
     with:
       playwright-mode: docker
+      update_screenshots: ${{ inputs.update_screenshots || false }}
```
dlzhry2, line 281:
```suggestion
      update_screenshots: ${{ inputs.update_screenshots }}
```

Any issue with this? It defaults to false right?

.github/workflows/e2e-and-api-tests.yml
williamwillis11: this step is failing currently

this step was only intended to run when the tests were running in CD ie against the azure environment because its post-merge of code. 

If the the tests are running in CI and fail, you cant merge your code. There is going to be a lot of this as people develop code so having this working would cause a lot of noise.

this step was added in, in error :) by Euan when he split the e2e template file into CI (top of the file) and CD (bottom of the file). Its still in the CD section at the bottom and still works - i tested it :)

.github/workflows/fingertips-workflow.yml
williamwillis11: i guess i just wanted to make sure that if this is null or undefined for some reason then the fallback position is false 

maybe i am just being paranoid

dlzhry2: Yup, I think the default value covers us in this case.

.github/workflows/fingertips-workflow.yml
```diff
@@ -160,7 +171,8 @@ jobs:
   trend-analysis-app-test:
     needs: [check-changes, determine-version]
     if: >
-      github.event_name == 'workflow_dispatch' ||
+      (github.event_name == 'workflow_dispatch' &&
+      inputs.update_screenshots == false) ||
```
epercy-bjss, line 175: formatting

.github/workflows/fingertips-workflow.yml
```diff
@@ -187,7 +199,8 @@ jobs:
   frontend-build-and-test:
     needs: [check-changes, determine-version]
     if: >
-      github.event_name == 'workflow_dispatch' ||
+      (github.event_name == 'workflow_dispatch' &&
+      inputs.update_screenshots == false) ||
```
epercy-bjss, line 203: formatting

Nayira-BJSS (approved):

PR #535, opened 2025-05-01
DHSCFT-724: Update trend not available string

dlzhry2:
# Description

Jira ticket: [DHSCFT-724](https://bjss-enterprise.atlassian.net/browse/DHSCFT-724)

Very basic - was nice that we were using consts sensibly. Made the change pretty simple.


alibombaye (approved):

frontend/fingertips-frontend/components/molecules/Result/searchResult.test.tsx
```diff
@@ -179,7 +179,9 @@ describe('content', () => {
           showTrends={true}
         />
       );
-      expect(screen.queryByText('No trend data available')).toBeInTheDocument();
+      expect(
+        screen.queryByText('No recent trend data available')
+      ).toBeInTheDocument();
```
gareth-allan-bjss, line 184: A very minor point, but the title of this test still refers to the old `No trend data available` text. Might be worth updating it (I always favour avoiding including specific values in the title so we don't have this problem when they change).

gareth-allan-bjss (approved):

gareth-allan-bjss (approved):

PR #539, opened 2025-05-02
populate state fully when england selected

Joe-Hendley:
https://bjss-enterprise.atlassian.net/browse/DHSCFT-639

When England is selected from the area search results, the state is now populated in the same way as it would be from the area filter.

### DIFF
--- a/frontend/fingertips-frontend/components/molecules/AreaSuggestionPanel/index.tsx
+++ b/frontend/fingertips-frontend/components/molecules/AreaSuggestionPanel/index.tsx
@@ -12,6 +12,7 @@ import {
   allAreaTypes,
   englandAreaType,
 } from '@/lib/areaFilterHelpers/areaType';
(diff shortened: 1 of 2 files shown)


frontend/fingertips-frontend/components/molecules/AreaSuggestionPanel/index.tsx
```diff
@@ -90,7 +91,23 @@ export const AreaAutoCompleteSuggestionPanel = ({
...
+          SearchParams.GroupTypeSelected,
+          englandAreaType.key
+        );
+        stateManager.addParamValueToState(
+          SearchParams.GroupSelected,
+          areaCodeForEngland
+        );
+        stateManager.addParamValueToState(
+          SearchParams.GroupAreaSelected,
+          'ALL'
```
alibombaye, line 106: We have a const for this 
`import { ALL_AREAS_SELECTED } from './areaFilterHelpers/constants';`

alibombaye (approved):

PR #542, opened 2025-05-02
DHSCFT-735 - remove automapper and manually create mappers

MonkeyDate:
# Description

Jira ticket: [DHSCFT-725](https://bjss-enterprise.atlassian.net/browse/DHSCFT-725)

Remove all uses of automapper.

## Changes

- remove automaper nuget package from project
- manually create mappers
- update any references to mappers in codebase
- test.

## Validation

Unit tests and automated tests.


api/DHSC.FingertipsNext.Modules.HealthData.Service/IMapper.cs
```diff
@@ -0,0 +1,18 @@
...
+    BenchmarkComparisonMethod MapBenchmarkComparisonMethod(string source);
+    BenchmarkComparison? Map(BenchmarkComparisonModel? source);
+    Deprivation? Map(DeprivationDimensionModel? source);
+    Age? Map(AgeDimensionModel? source);
+    Sex? Map(SexDimensionModel? source);
+    HealthDataPoint? Map(HealthMeasureModel? source);
+    List<HealthDataPoint> Map(IList<HealthMeasureModel>? source);
+    IndicatorQuartileData? Map(QuartileDataModel? source);
+    List<IndicatorQuartileData> Map(IList<QuartileDataModel>? source);
+};
```
dlzhry2, line 18: New line needed.

api/DHSC.FingertipsNext.Modules.HealthData.Service/IMapper.cs
```diff
@@ -0,0 +1,18 @@
+﻿using DHSC.FingertipsNext.Modules.HealthData.Repository.Models;
+using DHSC.FingertipsNext.Modules.HealthData.Schemas;
+
+namespace DHSC.FingertipsNext.Modules.HealthData.Service;
+
+public interface IMapper
+{
+    IndicatorPolarity MapIndicatorPolarity(string source);
+    BenchmarkComparisonMethod MapBenchmarkComparisonMethod(string source);
+    BenchmarkComparison? Map(BenchmarkComparisonModel? source);
```
dlzhry2, line 10: Definitely need to discuss with .NET lead about all these warnings in the API package. Do we need to make them non-nullable in the mapper - assume it was done for a specific reason.

https://learn.microsoft.com/en-us/ef/core/miscellaneous/nullable-reference-types

api/DHSC.FingertipsNext.Modules.HealthData/Mappings/Mapper.cs
```diff
@@ -0,0 +1,147 @@
...
+            {"BETTER", BenchmarkOutcome.Better},
+            {"BEST", BenchmarkOutcome.Best},
+            {"WORSE", BenchmarkOutcome.Worse},
+            {"WORST", BenchmarkOutcome.Worst},
+            {"SIMILAR", BenchmarkOutcome.Similar}
+        };
+
+        return new BenchmarkComparison
+        {
+            Outcome = outcomeDict[source.Outcome],
```
dlzhry2, line 57: What if the value is not in the dictionary? I can see we have done something to handle unexpected values in the switch statement in the method above...maybe that?

api/DHSC.FingertipsNext.Modules.HealthData/Mappings/Mapper.cs
```diff
@@ -35,12 +35,26 @@ public BenchmarkComparisonMethod MapBenchmarkComparisonMethod(string? source)
         if (source == null)
             return null;
 
+        var outcomeDict = new Dictionary<string, BenchmarkOutcome>
```
dlzhry2, line 38: This dictionary will be created every single time the method is called. Recommend creating a Read only dictionary and making it privately accessible in this class only.

Or switch statement also good.

api/DHSC.FingertipsNext.Modules.HealthData/Mappings/Mapper.cs
WolfieKnee: changed to use switch pattern including `_` case.

api/DHSC.FingertipsNext.Modules.HealthData/Mappings/Mapper.cs
WolfieKnee: see above

api/DHSC.FingertipsNext.Modules.HealthData/Mappings/Mapper.cs
```diff
@@ -0,0 +1,147 @@
...
+    public HealthDataPoint? Map(HealthMeasureModel? source)
+    {
+        return source == null
+            ? null
+            : MapNotNull(source);
+    }
+   
+    public List<HealthDataPoint> Map(IList<HealthMeasureModel>? source)
+    {
+        return source == null ? [] : source.Select(MapNotNull).ToList();
```
dlzhry2, line 99: I'm not sure why this would ever be null. If no health measure objects are found for a given query then it'll just be an empty list right? Not null. Same would go for any other list we are mapping over.

dlzhry2: At least every instance of something passed to the map function I can see is always using .toList() to is always a list of items or an empty list.

dlzhry2 (commented):
Having looked at the background on this: https://learn.microsoft.com/en-us/ef/core/miscellaneous/nullable-reference-types I would go with us respecting the existing models in the packages where we have added our own mappers.

I.e. if a class attribute is not marked as nullable, then treat it as such in the mapper. Of course, there are instances where we might have not requested a particular dimension i.e. navigation e.g. retrieved a health measure without deprivation for example.

But it should be with the developer to use the mapper appropriately based on the query.

api/DHSC.FingertipsNext.Modules.HealthData/Mappings/Mapper.cs
WolfieKnee: Agreed - I've removed the null handling, and nullable from the signature and interface. All unit tests and http tests passing. I'll work on patterning this approach throughout.

api/DHSC.FingertipsNext.Modules.HealthData/HttpClientScripts/test-indicators-benchmark-quartile.http
```diff
@@ -7,14 +7,14 @@ GET http://localhost:5144/indicators/quartiles?indicator_ids=94063&area_code=N85
...
-            "q0Value": 15.789474,
-            "q1Value": 64.92538,
-            "q2Value": 71.830986,
-            "q3Value": 76.90058,
-            "q4Value": 88.64353,
-            "areaValue": 76.190475,
-            "ancestorValue": 76.455025,
-            "englandValue": 69.89724
+            "q0Value": 15.7894736842105,
+            "q1Value": 64.9253731343284,
```
HowardMayBjss, line 11: Just out of interest - the original code had rounded this up when I would have expected it to round down. Looks like we were on the limits of accuracy of a float?

api/DHSC.FingertipsNext.Modules.Area.Service/AreaService.cs
```diff
@@ -98,5 +97,5 @@ public async Task<List<AreaType>> GetAreaTypes(string? hierarchyType = null) =>
     /// </summary>
     /// <param name="services"></param>
     public static void RegisterMappings(IServiceCollection services) => 
-        services.AddAutoMapper(typeof(AutoMapperProfiles));
+        services.AddScoped<IMapper, Mapper>();
```
dlzhry2, line 100: Could this be a singleton? I think the mapper is threadsafe right? Could safe a bit on memory...

api/DHSC.FingertipsNext.Modules.Area.Service/Mapper.cs
```diff
@@ -0,0 +1,72 @@
...
+        return new AreaType
+        {
+            Key = source.AreaTypeKey,
+            Name = source.AreaTypeName,
+            HierarchyName = source.HierarchyType,
+            Level = source.Level,
+        };
+    }
+
+    public List<AreaType> Map(IList<AreaTypeModel>? source)
```
dlzhry2, line 29: Does this need to be nullable?

api/DHSC.FingertipsNext.Modules.Area.Service/Mapper.cs
```diff
@@ -0,0 +1,72 @@
...
+    {
+        return new Schemas.Area
+        {
+            Code = source.AreaCode,
+            Name = source.AreaName,
+            AreaType = Map(source.AreaType)
+        };
+    }
+
+    public List<Schemas.Area> Map(IList<AreaModel>? source)
```
dlzhry2, line 59: Nullable?

api/DHSC.FingertipsNext.Modules.Area.Service/AreaService.cs
dlzhry2: I took a look and am fairly sure it's threadsafe - but the Playwright tests and nice and parallel, which should give it a good workout.

trend-analysis/TrendAnalysisApp.UnitTests/Helpers/HealthMeasureModelHelper.cs
```diff
@@ -0,0 +1,87 @@
...
+
+public class HealthMeasureModelHelper(
+    int key = 1,
+    short year = 2025,
+    double? count = 1.0,
+    double value = 1.0,
+    double denominator = 1.0,
+    double? lowerCi = 1.0,
+    double? upperCi = 1.0
+
```
dlzhry2, line 13: Empty line - can remove.

trend-analysis/TrendAnalysisApp.UnitTests/Helpers/HealthMeasureModelHelper.cs
```diff
@@ -0,0 +1,87 @@
...
+    {
+        return WithIndicatorDimension(new IndicatorDimensionModel
+        {
+            IndicatorKey = (short)key,
+            Name = name,
+            IndicatorId = indicatorId
+        });
+    }
+    
+    // private IndicatorDimensionModel DefaultIndicatorDimension()
```
dlzhry2, line 38: Commented code.

trend-analysis/TrendAnalysisApp.UnitTests/Helpers/HealthMeasureModelHelper.cs
```diff
@@ -0,0 +1,87 @@
...
+        return new HealthMeasureModel
+        {
+            HealthMeasureKey = key,
+            Count = count,
+            Value = value,
+            Denominator = denominator,
+            LowerCI = lowerCi,
+            UpperCI = upperCi,
+            Year = year,
+            // AreaKey = areaDimension.AreaKey,
```
dlzhry2, line 71: Lots of commented code.

trend-analysis/TrendAnalysisApp.UnitTests/Helpers/HealthMeasureModelHelper.cs
```diff
@@ -0,0 +1,87 @@
...
+            // AgeDimension = ageDimension,
+            IndicatorDimension = indicatorDimension,
+            // SexDimension = sexDimension,
+            TrendDimension = trendDimension,
+            // DeprivationDimension = deprivationDimension,
+            // IsAggregate = isAggregate
+        };
+    }
+    
+}
```
dlzhry2, line 87: New line.

trend-analysis/TrendAnalysisApp.UnitTests/TrendDataProcessorTests.cs
```diff
@@ -55,4 +80,70 @@ public void TestUpdateIndicatorSearchDataWritesCorrectDataToFile() {
             Arg.Is<JArray>(ja => JToken.DeepEquals(ja, expectedUpdatedFileContents))
         );
     }
+    
+    // DHSCFT-559: Test 1 - trends are not calculated for population indicators (ones to skip)
+    [Fact]
+    public async Task TestTrendDataProcessor_DoesNotReturnTrendsForIndicatorsToSkip()
+    {
+        // arrange
+        // TODO: populate inMemoryDb with mock data for testing
```
dlzhry2, line 89: Where did these changes come from?

dlzhry2 (commented):
Couple of general comments need addressing.

api/DHSC.FingertipsNext.Modules.HealthData.Service/DHSC.FingertipsNext.Modules.HealthData.Service.csproj
```diff
@@ -7,10 +7,7 @@
     </PropertyGroup>
 
     <ItemGroup>
-      <PackageReference Include="AutoMapper" Version="13.0.1" />
-    </ItemGroup>
-
-    <ItemGroup>
+      <ProjectReference Include="..\DHSC.FingertipsNext.Modules.Area.Service\DHSC.FingertipsNext.Modules.Area.Service.csproj" />
```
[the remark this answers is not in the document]
WolfieKnee: not that I'm aware of - I've removed

trend-analysis/TrendAnalysisApp.UnitTests/TrendDataProcessorTests.cs
WolfieKnee: merge issues - I've rolledback

api/DHSC.FingertipsNext.Modules.HealthData.Service/IMapper.cs
WolfieKnee: new line added :)

api/DHSC.FingertipsNext.Modules.Area.Service/Mapper.cs
WolfieKnee: yes, in this case there are usages which could be null.

api/DHSC.FingertipsNext.Modules.Area.Service/Mapper.cs
WolfieKnee: yes - usagein areaService may b null

api/DHSC.FingertipsNext.Modules.Area.Service/Mapper.cs
dlzhry2: Could you point me to this? I had a brief walk of the code and could not see this, but might be missing some of the detail.

api/DHSC.FingertipsNext.Modules.Area.Service/Mapper.cs
dlzhry2: Could you point me to this? I had a brief walk of the code and could not see this, but might be missing some of the detail.

api/DHSC.FingertipsNext.Modules.HealthData/HttpClientScripts/test-indicators-benchmark-quartile.http
WolfieKnee: ["The float type can represent values ranging from approximately 1.5 × 10⁻⁴⁵ to 3.4 × 10³⁸ with a precision of 7 digits."](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/types#837-floating-point-types) - so yes, it looks like it.

api/DHSC.FingertipsNext.Modules.Area.Service/AreaService.cs
WolfieKnee: Yes, I've tried this and it looks to be OK.

api/DHSC.FingertipsNext.Modules.Area.Service/Mapper.cs
HowardMayBjss: From looking at the code I think that the Mapper code in question here is exclusively dealing with values passed back from the AreaRepository. I think we are consistently using ToListAsync() which will convert an empty set from the DB query into an empty Array. And so it won't ever be NULL. Worth checking to see if I've missed any cases where this isn't so.

api/DHSC.FingertipsNext.Modules.Area.Service/Mapper.cs
```diff
@@ -0,0 +1,72 @@
+﻿using DHSC.FingertipsNext.Modules.Area.Repository.Models;
+using DHSC.FingertipsNext.Modules.Area.Schemas;
+
+namespace DHSC.FingertipsNext.Modules.Area.Service;
+
+public interface IMapper
```
obarojohnson, line 6: **IMapper** sounds general , which is good , but it depends on more concreate DataType at this point. Should we change the naming to more descriptive naming.  **IAreaModelMapping** or **IAreaMapping**

api/DHSC.FingertipsNext.Modules.Area.Service/Mapper.cs
WolfieKnee: I follow this a bit deeper and the implementations all have `ToList()` so you're correct that it doesn't need to be nullable. I'll remove here as well.

api/DHSC.FingertipsNext.Modules.Area.Service/Mapper.cs
```diff
@@ -0,0 +1,72 @@
...
+{
+    AreaType Map(AreaTypeModel source);
+    List<AreaType> Map(IList<AreaTypeModel>? source);
+    AreaWithRelations Map(AreaWithRelationsModel source);
+    Schemas.Area Map(AreaModel source);
+    List<Schemas.Area> Map(IList<AreaModel>? source);
+    RootArea MapToRootArea(AreaModel source);
+}
+
+public class Mapper : IMapper
```
obarojohnson, line 16: **Mapper**  has a class is very general here. Looking at the AreaServce class where its just it looks like its mapping the area repository models to the C# class for the API. Should we be naming this to reflect  what its doing.  I think its doing something like **RepositoryAreaModelsMapper** feel free to suggest a better name.

api/DHSC.FingertipsNext.Modules.HealthData.Service/IMapper.cs
```diff
@@ -0,0 +1,13 @@
+﻿using DHSC.FingertipsNext.Modules.HealthData.Repository.Models;
+using DHSC.FingertipsNext.Modules.HealthData.Schemas;
+
+namespace DHSC.FingertipsNext.Modules.HealthData.Service;
+
+public interface IMapper
```
obarojohnson, line 6: Same with the naming here as mentioned above. A more descriptive name here will help readers understand what the class interface is doing.

api/DHSC.FingertipsNext.Modules.HealthData/Mappings/Mapper.cs
```diff
@@ -0,0 +1,130 @@
+﻿using DHSC.FingertipsNext.Modules.HealthData.Repository.Models;
+using DHSC.FingertipsNext.Modules.HealthData.Schemas;
+using DHSC.FingertipsNext.Modules.HealthData.Service;
+
+namespace DHSC.FingertipsNext.Modules.HealthData.Mappings;
+
+public class Mapper : IMapper
```
obarojohnson, line 7: Same here this Mapper is more related to a health data . **HealthDataIndicatorMapper**  also please feel free to suggestion more descriptive naming here

obarojohnson (commented):
I am also begin  to foreseen that the IMapper class we become a share models across the ```CSharp
solutions.  interface IFingertipsMapper <TModelSource, TModelDestination>{
     IList<TModelDestination> Map(TModelSource model);
     IList<TModelDestination> Map(IList<TModelSource> models);
}  
```

But not now. Just a small naming clarity here

api/DHSC.FingertipsNext.Modules.Area.Service/Mapper.cs
obarojohnson: Or something shorter  **AreaModelsMapper**

api/DHSC.FingertipsNext.Modules.Area.Service/Mapper.cs
WolfieKnee: I think these general names came from using the AutoMapper library. I've renamed based on `AreaMapper`.

api/DHSC.FingertipsNext.Modules.Area.Service/Mapper.cs
WolfieKnee: This opens the question of whether it should be named for what it maps from (e.g. `MapfromAreasModel`), or what is maps to (e.g. `MapToAreasSchema`) which becomes quite verbose and specific. I think `AreaMapper` is a nice balance of specificity. Are you OK with that?

api/DHSC.FingertipsNext.Modules.HealthData.Service/IMapper.cs
WolfieKnee: Renames to `HealthdataMapper` consistent with areas above.

api/DHSC.FingertipsNext.Modules.HealthData/Mappings/Mapper.cs
WolfieKnee: Again, renamed based on `HeathDataMapper`.

HowardMayBjss (approved):

PR #545, opened 2025-05-06
DHSCFT-713: Show single area when there is a full match

adawari-thomas-bjss:
# Description

Jira ticket: [DHSCFT-713](https://bjss-enterprise.atlassian.net/browse/DHSCFT-713)

## Changes

- Disabled fuzzy matching
- Added a conditional playwright test that runs only in the pipelines to validate that the area suggester returns only a single area when there is a full match

## Validation

UI test

image


frontend/fingertips-frontend/playwright/page-objects/pages/homePage.ts
```diff
@@ -110,4 +111,24 @@ export default class HomePage extends AreaFilter {
...
     );
   }
+
+  async checkAreaSuggestionPanelContainsSingleItem(
+    areaSearchTerm: string,
+    areaName: string
+  ) {
+    await this.fillAndAwaitLoadingComplete(
+      this.page.getByTestId(this.areaSearchField).getByRole('textbox'),
+      areaSearchTerm!
```
dlzhry2, line 121:
```suggestion
      areaSearchTerm
```

Because of the usage of this function I think we can reliably say it will always be a string provided.

dlzhry2 (commented):
Couple of minor comments but I'm happy to approve. Just need a bit of a steer from @williamwillis11 and @Nayira-BJSS on how we'd like to do the conditional tests e.g. using .skip or if they'd prefer a different approach. Probs worth flagging to them in the Slack post for the PR.

frontend/fingertips-frontend/playwright/tests/fully_integrated_e2e_tests/ai_search_tests.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+import { AreaDocument } from '@/lib/search/searchTypes';
+
+const areaSearchTerm: AreaDocument = {
+  areaCode: 'E12000002',
+  areaType: 'Regions',
+  areaName: 'North West Region',
+};
+
+test.describe('Area search suggestions', () => {
+  test.skip(
```
adawari-thomas-bjss, line 12: @williamwillis11 @Nayira-BJSS this line causes a lint warning in the code editor. Is there a desired approach to handling conditional tests?

gareth-allan-bjss: Looks like Playwright supports tags: https://playwright.dev/docs/test-annotations#tag-tests

I think this would get us round this issue (and potentially show the tags in the report too, which might be nice).

frontend/fingertips-frontend/playwright/page-objects/pages/homePage.ts
```diff
@@ -110,4 +111,24 @@ export default class HomePage extends AreaFilter {
       expectedValidationMessage
     );
   }
+
+  async checkAreaSuggestionPanelContainsSingleItem(
```
williamwillis11, line 115: could we rename this to something like:

checkAreaSuggestionPanelContainsItems

and have a number parameter named something like:

numberOfSuggestions

 with a default value of 1 
 
 and in the assertion do: 
 
 .toHaveCount(numberOfSuggestions);

frontend/fingertips-frontend/playwright/page-objects/pages/homePage.ts
```diff
@@ -112,23 +111,26 @@ export default class HomePage extends AreaFilter {
...
-      this.page.getByTestId(this.searchSuggestionPanel).getByRole('listitem')
-    ).toHaveCount(1);
-    await expect(
-      this.page
-        .getByTestId(this.searchSuggestionPanel)
-        .getByRole('listitem')
-        .nth(0)
-    ).toContainText(areaName);
+      this.page.getByTestId(this.suggestedAreasPanel).getByRole('listitem')
+    ).toHaveCount(resultCount);
```
gareth-allan-bjss, line 126: A minor point, but do you need `resultCount` as a separate parameter? Couldn't you just check there are `results.length` instances?

gareth-allan-bjss (approved):

PR #546, opened 2025-05-06
Feature/dhscft 739 e2e test data source

williamwillis11:
# Description

Jira ticket: [DHSCFT-739](https://bjss-enterprise.atlassian.net/browse/DHSCFT-739)

Add test coverage for the data source displayed on the chart page on the one indicator view. The approach should allow us to check for more things on the chart page easily.

## Changes

- Add 

## Validation

e2e tests pass locally and in ...

### DIFF
--- a/frontend/fingertips-frontend/package.json
+++ b/frontend/fingertips-frontend/package.json
@@ -23,7 +23,7 @@
     "test-ci": "jest --ci --silent",
     "open-jest-coverage": "open coverage/lcov-report/index.html",
     "test-ui-local-mocks": "playwright test isolated_ui_tests --ignore-snapshots",
-    "test-e2e-local-docker": "cross-env MOCK_SERVER=false npm run start:all && playwright test fully_integrated_e2e_tests --ignore-snapshots || true && npm run stop:all",
+    "test-e2e-local-docker": "cross-env MOCK_SERVER=false npm run start:all && playwright test fully_integrated_e2e_tests --ignore-snapshots --ui",
     "test-e2e-ci": "MOCK_SERVER=false playwright test fully_integrated_e2e_tests",
     "test-e2e-ui-mode": "MOCK_SERVER=false playwright test fully_integrated_e2e_tests --ui --ignore-snapshots",
     "prettier": "prettier --write .",
--- a/frontend/fingertips-frontend/playwright/page-objects/pages/chartPage.ts
+++ b/frontend/fingertips-frontend/playwright/page-objects/pages/chartPage.ts
@@ -117,6 +117,12 @@ export default class ChartPage extends AreaFilter {
           )
         );
       }
+      // if its one of the chart components that has a data source attributed then check it
+      if (visibleComponent.componentProps.hasDataSource) {
+        await expect(
+          this.page.getByTestId(visibleComponent.componentLocator)
+        ).toHaveText(`Data source: ${visibleComponent.componentLocator}`);
+      }
       // if its one of the chart components that has a single time period dropdown then select the last in the list
       if (visibleComponent.componentProps.hasTimePeriodDropDown) {
         const combobox = this.page
--- a/frontend/fingertips-frontend/playwright/page-objects/pages/chartPage.ts
+++ b/frontend/fingertips-frontend/playwright/page-objects/pages/chartPage.ts
@@ -12,6 +12,7 @@ import {
   TestType,
 } from '@playwright/test';
 import AreaFilter from '../components/areaFilter';
+import { SimpleIndicatorDocument } from '@/playwright/tests/fully_integrated_e2e_tests/core_journeys.spec';
 
 export default class ChartPage extends AreaFilter {
   readonly backLink = 'chart-page-back-link';
@@ -78,7 +79,8 @@ export default class ChartPage extends AreaFilter {
     test: TestType<
       PlaywrightTestArgs & PlaywrightTestOptions,
       PlaywrightWorkerArgs & PlaywrightWorkerOptions
-    >
+    >,
+    selectedIndicators: SimpleIndicatorDocument[]
   ) {
     const testInfo = test.info();
     const testName = testInfo.title;
@@ -86,6 +88,7 @@ export default class ChartPage extends AreaFilter {
       indicatorMode,
       areaMode
     );
+
     console.log(
       `for indicator mode: ${indicatorMode} + area mode: ${areaMode} - checking that chart components: ${visibleComponents
         .map(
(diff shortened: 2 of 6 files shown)


frontend/fingertips-frontend/package.json
```diff
@@ -23,7 +23,7 @@
     "test-ci": "jest --ci --silent",
     "open-jest-coverage": "open coverage/lcov-report/index.html",
     "test-ui-local-mocks": "playwright test isolated_ui_tests --ignore-snapshots",
-    "test-e2e-local-docker": "cross-env MOCK_SERVER=false npm run start:all && playwright test fully_integrated_e2e_tests --ignore-snapshots || true && npm run stop:all",
+    "test-e2e-local-docker": "cross-env MOCK_SERVER=false npm run start:all && playwright test fully_integrated_e2e_tests --ignore-snapshots",
```
Nayira-BJSS, line 26: Did you mean to push this?

williamwillis11: nope :)

frontend/fingertips-frontend/playwright/page-objects/components/areaFilter.ts
```diff
@@ -30,6 +35,28 @@ export default class AreaFilter extends BasePage {
     return Promise.all(pillFilterNames.map(async (l) => await l.textContent()));
   }
 
+  async checkSelectedIndicatorPillsText(
+    expectedPillText: SimpleIndicatorDocument[]
+  ) {
+    const pillElements = await this.page
```
dlzhry2, line 41: Would be good if we could use the pattern at line 24 for getting this information. I.e. define an `indicatorPills()` method.

frontend/fingertips-frontend/playwright/page-objects/components/areaFilter.ts
```diff
@@ -30,6 +35,28 @@ export default class AreaFilter extends BasePage {
...
+      .getByTestId(this.pillContainer)
+      .getByTestId(this.filterName)
+      .all();
+
+    const pillTexts = await Promise.all(
+      pillElements.map((pill) => pill.textContent())
+    );
+
+    const cleanedPillTexts = pillTexts.map((text) =>
+      text!.replace(/View background information$/, '').trim()
```
dlzhry2, line 52: We could actually do the checks in here to avoid 2 for loops. (But see comment below for flipside argument).

frontend/fingertips-frontend/playwright/page-objects/components/areaFilter.ts
```diff
@@ -30,6 +35,28 @@ export default class AreaFilter extends BasePage {
...
+    );
+
+    for (const pillText of expectedPillText) {
```
dlzhry2, line 55: Could be worth expecting that the list lengths are the same, otherwise could mean that there are extra items showing in the actual.

frontend/fingertips-frontend/playwright/page-objects/pages/chartPage.ts
```diff
@@ -104,6 +107,23 @@ export default class ChartPage extends AreaFilter {
...
+        .allTextContents();
+
+      allDataSources.forEach((dataSource) => {
+        expect(dataSource).toBe(
+          `Data source: ${selectedIndicators[0].dataSource}`
+        );
+      });
+    }
+    // and check that the data source is not displayed for the other indicator modes
+    if (indicatorMode != IndicatorMode.ONE_INDICATOR) {
```
dlzhry2, line 123:
```suggestion
    if (indicatorMode !== IndicatorMode.ONE_INDICATOR) {
```
Or could use else

frontend/fingertips-frontend/playwright/page-objects/pages/resultsPage.ts
```diff
@@ -97,50 +96,49 @@ export default class ResultsPage extends AreaFilter {
...
+        .getByTestId(this.indicatorCheckboxContainer)
+        .getByRole('checkbox')
+        .all();
+
+      // check that the displayed indicator IDs are valid for the searched for criteria
+      for (const checkbox of displayedIndicatorCheckboxList) {
+        const indicatorDataTestID = await checkbox.getAttribute('value');
+        if (
+          indicatorDataTestID &&
+          !JSON.stringify(allValidIndicators).includes(indicatorDataTestID)
```
dlzhry2, line 126: Should we have a const for this stringified JSON rather than calling it every time in the for loop?

frontend/fingertips-frontend/playwright/testHelpers.ts
```diff
@@ -244,12 +275,49 @@ export function getAllIndicatorIdsForSearchTerm(
...
+  searchTerm: string
+): SimpleIndicatorDocument[] {
+  if (!searchTerm) return [];
+
+  const lowerCasedSearchTerm = searchTerm.toLowerCase();
+  return indicators
+    .filter(
+      (indicator) =>
+        indicatorsUsedInPOC(indicator) &&
+        indicatorsMatchingSearchTerm(indicator, lowerCasedSearchTerm)
```
dlzhry2, line 289: It is worth keeping in mind that the logic of this function is only a full match for what we have running in CI/using AI Search mocks.

With the work planned in this 4 week SoW, we will extend it to have synonym functionality. I.e. search heart, you may also get results for cardiovascular. Likewise, in future, we may extend further to use full AI.

This might hold up while we have a limited number of PoC indicators, and for certain search terms, but it is worth keeping in mind that, depending on the search term, you might get additional valid results when testing against Azure. Hope that makes sense!

For now this will be fine - it's just worth keeping in mind.

williamwillis11: yep got pagination in mind 

also the check at the moment is just that the results returned on page 1 are valid (given current simple searching approach that will need to change in a few weeks like you say)

frontend/fingertips-frontend/playwright/page-objects/components/areaFilter.ts
williamwillis11: the selected indicators only appear in the side filter when on the chart page (by design) so i dont think renaming this file to areaAndIndicatorFilter.ts would make sense, and really we dont do any filtering of indicators here, we just display them on only the chart page

frontend/fingertips-frontend/playwright/testHelpers.ts
dlzhry2: Perfect, sounds good. Marking resolved.

frontend/fingertips-frontend/playwright/page-objects/components/areaFilter.ts
dlzhry2: Only comment still needing looking at - all others have been resolved.

dlzhry2 (approved):
Looks good - just one comment that is not yet addressed to maybe tighten up one of the assertions.

frontend/fingertips-frontend/playwright/page-objects/components/areaFilter.ts
williamwillis11: thanks - addressed :)

PR #549, opened 2025-05-07
DHSCFT-734: Add title to the Heatmap

dlzhry2:
# Description

Jira ticket: [DHSCFT-734](https://bjss-enterprise.atlassian.net/browse/DHSCFT-734): there are a couple of clarifications which I asked the design team about the ticket. All is documented on the JIRA.

It was observed during testing that the Heatmap chart did not have a title.

## Changes

- Added the title with the correct styling
- Updated unit test snapshots
- Updated screenshots

## Validation

Local testing and regeneration of screenshots.


HowardMayBjss:
Why did the screenshots change? I had thought now they were only generated in the CI they would be stable and not change without a clear cause.

HowardMayBjss (approved):

PR #550, opened 2025-05-07
DHSCFT-746: show version numbers

stuc-bjss:
# Description

Jira ticket: [ABC-1234](https://example.com)

## Changes

- Etc.

## Validation


.github/workflows/fingertips-workflow.yml
```diff
@@ -591,13 +591,26 @@ jobs:
...
+    runs-on: ubuntu-latest
+    outputs:
+      git_hash: ${{ steps.setvars.outputs.git_hash }}
+      git_tag: ${{ steps.setvars.outputs.git_tag }}
+    steps:
+      - uses: actions/checkout@v4
+      - id: setvars
+        run: |
+          echo "git_hash=$(git rev-parse HEAD)" >> $GITHUB_OUTPUT
+          echo "git_tag=$(git describe --tags --abbrev=0 --always)" >> $GITHUB_OUTPUT
```
HowardMayBjss, line 604: what happens if it doesn't have a tag?

stuc-bjss: interestingly with --always it returns a commit hash instead, without it it returns 'fatal: No names found, cannot describe anything.'

.github/workflows/fingertips-workflow.yml
```diff
@@ -214,6 +214,15 @@ jobs:
...
         run: npm ci
 
+      - name: Make script executable
+        run: chmod +x ./updateVersion.sh
+
+      - name: Update version tag and hash
+        run: ./updateVersion.sh
+
+      - name: Show contents version file
+        run: cat ./lib/version.ts
```
dlzhry2, line 224: As I understand it, each job in a Github action will have a separate environment. We can modify the file here, but when it comes to pushing the image we use `.github/workflows/build-container.yml` see line 307 in fingertips-workflow.

It uses a fresh checkout of the code, so will only see the original file - not the updated one.

dlzhry2: https://github.com/actions/upload-artifact action can be used to make a file available to a subsequent job to use the download action.

However, we might want to extend the `.github/workflows/build-container.yml` to do the file update rather than passing files. The only issue is that has been written in a nice and generic way, so we'd have to do some sort of if statement for only performing this step if we're working with the frontend.

I prefer the approach of passing env vars to the container - any reason this was dropped out?

dlzhry2: Not 100% but won't really know until it's tested

dlzhry2: Do we still need this here if we're going to do this in the `build-container` workflow? Do we need it before test is run for particular assertions...hopefully not

.github/workflows/build-container.yml
```diff
@@ -32,6 +35,18 @@ jobs:
       - name: Checkout source
         uses: actions/checkout@v4
 
+      - name: Make script executable
+        if: inputs.is_frontend == 'true'
```
dlzhry2, line 39:
```suggestion
        if: inputs.is_frontend
```

Think this would suffice as it is actually a bool. The reason it's like this in some places in the fingertips workflow is because some of the outputs are exported to the GITHUB_OUTPUT as strings.

.github/workflows/build-container.yml
```diff
@@ -32,6 +35,18 @@ jobs:
...
+        run: chmod +x ./updateVersion.sh
+
+      - name: Update version tag and hash
```
dlzhry2, line 42: This and the other steps will not work unless we specify the working directory. We could hardcode it or get it from inputs.context-location.

It's at that point I do wonder if we should stick with the container env var approach.

PR #552, opened 2025-05-08
Feature/dhscft 747 enhancement of pyramid

obarojohnson:
# Description

Jira ticket: [DHSCFT-747](https://bjss-enterprise.atlassian.net/browse/DHSCFT-747)

Pop Pyramid Chart

## Capture
image

- Change the spacing and also the axis labels


frontend/fingertips-frontend/components/organisms/PopulationPyramid/createChartOptions.ts
```diff
@@ -145,11 +168,12 @@ const createPopPyramidSeriesOptions = (
         tickmarkPlacement: 'on',
         tickColor: GovukColours.DarkSlateGray,
         accessibility: {
-          description: '{xAxisTitle} degrees {series.name}',
+          description: '{xAxisMaleTitle} degrees {series.name}',
```
dlzhry2, line 171: Not strictly related to this PR - looks like an existing thing. But what does degrees mean here?

Also I think we'd need to use `${xAxisTitle}...` to make it render dynamically. I notice that we handle the accessibility description differently in other charts e.g. see the line chart options in `frontend/fingertips-frontend/components/viewPlots/OneIndicatorOneAreaViewPlots/index.tsx`

dlzhry2 (approved):
Looks good to me - one minor clarification on the accessibility options.

frontend/fingertips-frontend/components/organisms/PopulationPyramid/createChartOptions.ts
obarojohnson: That is a good point,  But I am not sure why but maybe I can guess because Pyramid is not as  reusable as the the line chart one.

PR #555, opened 2025-05-08
Support benchmarking against area group

HowardMayBjss:
# Description

Jira ticket: [DHSCFT-711](https://example.com](https://bjss-enterprise.atlassian.net/browse/DHSCFT-711))

This updates the API to support benchmarking against an AreaGroup rather than England 

## Changes

- New parameters on the API / swagger
- Implementation in the Indicator Service
- Http API tests
- Unit test in addition to Http tests

## Validation

Additional tests added and existing tests validated. 
Further manual validation will be possible once the UI has been updated in follow on tickets


api/DHSC.FingertipsNext.Modules.HealthData.Service/IndicatorService.cs
```diff
@@ -137,7 +149,7 @@ IndicatorPolarity polarity
         }
 
         //if RAG is the benchmark method use England as the comparison area and add England to the areas we want data for
-        var benchmarkAreaCode = AreaCodeEngland; //for PoC all benchmarking is against England, post PoC this will be passed in as a variable
+        var benchmarkAreaCode = benchmarkRefType == BenchmarkReferenceType.AreaGroup ? areaGroup: AreaCodeEngland; //for PoC all benchmarking is against England, post PoC this will be passed in as a variable
```
alibombaye, line 152: Does the comment need updating to say use the areaGroup if the BenchmarkReferenceType is now AreaGroup. Actually do we need the top comment at all anymore?

api/DHSC.FingertipsNext.Modules.HealthData/Controllers/V1/IndicatorsController.cs
```diff
@@ -61,11 +65,20 @@ public async Task<IActionResult> GetIndicatorDataAsync(
...
+                    $"Missing parameter 'area_group'. When benchmark_ref_type is set to AreaGroup then the area_group parameter must be set"
+            });
+
         var indicatorData = await _indicatorsService.GetIndicatorDataAsync
         (
             indicatorId,
             areaCodes ?? [],
             areaType,
+            areaGroup, 
+            benchmarkRefType ?? BenchmarkReferenceType.Unknown,
```
alibombaye, line 81: `benchmarkRefType` defaults to `Unknown` do we still need to do this?

api/DHSC.FingertipsNext.Modules.HealthData.UnitTests/Controllers/V1/IndicatorsControllerTests.cs
```diff
@@ -144,7 +146,7 @@ public async Task GetIndicatorData_ReturnsBadResponse_WhenMoreThan10CodesSupplie
     {
         var response = await _controller.GetIndicatorDataAsync(3,
             ["areaCode1", "ac2", "ac3", "ac4", "ac5", "ac6", "ac7", "ac8", "ac9", "ac10", "ac11"], "",
-            [1]) as BadRequestObjectResult;
+            years:[1]) as BadRequestObjectResult;
 
         response?.StatusCode.ShouldBe(400);
         (response?.Value as SimpleError)?.Message.Contains("Too many values supplied for parameter area_codes").ShouldBeTrue();
```
alibombaye, line 152: Do we need an additional test for the badRequest for when area_group isn't provided as expected?

api/DHSC.FingertipsNext.Modules.HealthData/Controllers/V1/IndicatorsController.cs
```diff
@@ -42,6 +44,8 @@ public async Task<IActionResult> GetIndicatorDataAsync(
         [FromRoute] int indicatorId,
         [FromQuery(Name = "area_codes")] string[]? areaCodes = null,
         [FromQuery(Name = "area_type")] string areaType = "",
+        [FromQuery(Name = "area_group")] string? areaGroup = "",
```
HowardMayBjss, line 47: Sonar is complaining that this function has more than 7 parameters - this waterfalls down to a couple of other functions which also use these params. I don't see that making changes to reduce the number of params is helpful in this instance and suggest we tell Sonar to back off. But would value reviewer input on this

api/definition/swagger.yaml
```diff
@@ -804,12 +806,28 @@ components:
...
         type: string
+    benchmark_ref_type:
+      in: query
+      name: benchmark_ref_type
+      description: Controls whether benchmarking is performed against England or an Ancestor
+      schema:
+        type: string
+        enum:
+            - England
+            - AreaGroup
```
stuc-bjss, line 830: Can we reference this at a higher level of the swagger doc such that the exported enum is `BenchmarkRefType` rather than the rather ugly `GetHealthDataForAnIndicatorBenchmarkRefTypeEnum` which the auto client generator produces

stuc-bjss (approved):

api/DHSC.FingertipsNext.Modules.HealthData/Controllers/V1/IndicatorsController.cs
stuc-bjss: in a regular function it would smell a bit but as these are optional values from a GET query string it doesn't bother me

alibombaye (approved):

PR #559, opened 2025-05-09
Feature/dhscft 593 bug fix search problem

obarojohnson:
# Description

Jira ticket: [DHSCFT-593](https://bjss-enterprise.atlassian.net/browse/DHSCFT-593)

image

Escape the search string, as it may be the reason why the regular expression search is failing in AI Search.
Looking at the Azure SDK [source code](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/search/azure-search-documents/azure/search/documents/_search_client.py#L145), you can see that the string is not being escaped, even though regular expressions are accepted.

# Manual Testing
This has also be manually tested locally with AI search and it works as shown on the Image above.


frontend/fingertips-frontend/lib/search/indicatorSearchService.ts
```diff
@@ -12,6 +12,10 @@ import {
 } from './searchTypes';
 import { IndicatorMapper } from './indicatorMapper';
 
+const escapeString = (text: string): string => {
+  return text.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
```
gareth-allan-bjss, line 16: It would be good to understand where this list of characters has come from. The [AI search docs on the full Lucene syntax](https://learn.microsoft.com/en-us/azure/search/query-lucene-syntax#escaping-special-characters) look to have a different list of characters that need escaping.

obarojohnson: The one I added are basic regular url parameters expression escape . Good point I will add use this characters  https://learn.microsoft.com/en-us/azure/search/query-lucene-syntax#escaping-special-characters instead. thanks Good spot

dlzhry2 (commented):
If you can, it would be really good if you could point your local at AI search and manually test. `npm run test-e2e-cd`. Until we have more environments we don't have a good way to test code that changes what we pass to AI search.

frontend/fingertips-frontend/lib/escapeString.test.ts
```diff
@@ -0,0 +1,50 @@
+import { escapeString } from './escapeString';
+
+describe('escapeString', () => {
+  it('should escape special characters correctly', () => {
+    const input = '+-&|!(){}[]^"~*?:\\/';
```
gareth-allan-bjss, line 5: Why two backslashes?
```suggestion
    const input = '+-&|!(){}[]^"~*?:\/';
```

frontend/fingertips-frontend/lib/escapeString.test.ts
```diff
@@ -0,0 +1,50 @@
...
+    const expected =
+      '\\+\\-\\&\\|\\!\\(\\)\\{\\}\\[\\]\\^\\\"\\~\\*\\?\\:\\\\\\/';
```
gareth-allan-bjss, line 7: The [documentation](https://learn.microsoft.com/en-us/azure/search/query-lucene-syntax#escaping-special-characters) says:
> escape the character by prefixing it with a single backslash (\). For example, for a wildcard search on https://, where :// is part of the query string, you would specify search=https\:\/\/*

Aren't we ending up with too many backslashes (two per special character) here? It would be nice if we could find a way to verify that with this escaping we can actually match as expected with special characters. Perhaps we can find an indicator with a special character in its title/description, or modify the data in a test index to include it?

gareth-allan-bjss, line 7: Why three slashes before the double quote?
```suggestion
      '\\+\\-\\&\\|\\!\\(\\)\\{\\}\\[\\]\\^\\"\\~\\*\\?\\:\\\\\\/';
```

frontend/fingertips-frontend/lib/escapeString.test.ts
```diff
@@ -0,0 +1,50 @@
...
+      'This is a test\\: \\+\\-\\&\\|\\!\\(\\)\\{\\}\\[\\]\\^\\"\\~\\*\\?\\:\\\\\\/';
+    const actual = escapeString(input);
+    expect(actual).toBe(expected);
+  });
+
+  it('should escape only the special characters in the string', () => {
+    const input = 'Test string with special chars: +, -, *, &';
+    const expected = 'Test string with special chars\\: \\+, \\-, \\*, \\&';
+    expect(escapeString(input)).toBe(expected);
+  });
```
gareth-allan-bjss, line 49: Is this a different scenario than `should escape some special characters correctly` above?

frontend/fingertips-frontend/lib/escapeString.test.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  it('should escape all special characters in the input string', () => {
+    const input = 'This is a test: +-&|!(){}[]^"~*?:\\/';
+    const expected =
+      'This is a test\\: \\+\\-\\&\\|\\!\\(\\)\\{\\}\\[\\]\\^\\"\\~\\*\\?\\:\\\\\\/';
+    const actual = escapeString(input);
+    expect(actual).toBe(expected);
+  });
```
gareth-allan-bjss, line 43: I'm not sure this adds much above the first test (which verifies all expected characters are escaped) and the `should escape some special characters correctly` test (which verifies we escape only the special characters.

I love to see thorough tests, but I think this may be leaning towards more maintenance overhead for little extra value.

frontend/fingertips-frontend/lib/escapeString.ts
```diff
@@ -0,0 +1,8 @@
+/*
+   escape characters in string before use with AI search 
+   please see this documentation for more details
+   reference: https://learn.microsoft.com/en-us/azure/search/query-lucene-syntax#escaping-special-characters
+*/
+export const escapeString = (text: string): string => {
+  return text.replace(/[+\-&|!(){}\[\]^"~*?:\\\/]/g, '\\$&');
```
gareth-allan-bjss, line 7: You've got a couple of Sonaqube issues (see the comment on this PR from Sonarqube) flagging unnecessary escape characters here that we should address.

frontend/fingertips-frontend/lib/escapeString.test.ts
obarojohnson: The first backslash is used to escape the second backslash. In JavaScript strings, the backslash has a special meaning, I believe.

You can test it here: https://regex101.com/.
\\ means "I want to match a single backslash (\)".
However, if I write it as \/, it match anything  /. because the first \ is use to escape the \ to appear.  and that is what we are looking for. 

Normally, we don't need to escape /, but because we're using replace instead of directly using a regular expression which is a string type we need to escape with it  / 
 So to be able to match \/ we have to do this \\\/ .

frontend/fingertips-frontend/lib/escapeString.test.ts
obarojohnson: No , its confusing but its not really. If you test it with https://*  it will produce the expected value. If you console it you will see the values as expected. \backslash have a special meaning in strings which means you have to escape.

obarojohnson: But if you console it out you will see \* which makes it a literal character taking out the special meaning.

frontend/fingertips-frontend/lib/escapeString.test.ts
obarojohnson: I was trying to check what happens if the special character appears at the end, but yes, you're right — it's the same as escaping it in the middle or anywhere else in the string. Thanks!

frontend/fingertips-frontend/lib/escapeString.test.ts
obarojohnson: I have remove it , since its mostly repetition and then  add a new test on https:// instead.

frontend/fingertips-frontend/lib/escapeString.test.ts
gareth-allan-bjss: OK, yeah, if I read it in VSCode it does some highlighting that makes it easier to follow!

frontend/fingertips-frontend/lib/escapeString.test.ts
gareth-allan-bjss: Yeah, makes sense now I look at it again.

frontend/fingertips-frontend/lib/escapeString.test.ts
gareth-allan-bjss: I think this is still there.

gareth-allan-bjss (approved):

PR #562, opened 2025-05-09
Feature/dhscft 731 long scrolling

stuc-bjss:
# Description

Jira ticket: [DHSCFT-731](https://bjss-enterprise.atlassian.net/browse/DHSCFT-731)

The issue was scrolling to the bottom of a very long page when the content of that page is rendering as you scroll, the solution presented here is to add placeholder rows which render quickly instead of the rows with data, thus padding out the height of the document to something approximating the real elements.

## Changes

- Added placeholder rows

## Validation

Particularly noticeable here http://localhost:3000/chart?si=hospital&is=41101&ats=districts-and-unitary-authorities&gts=england&gs=E92000001&gas=ALL when trying to scroll to the very bottom of the page.


frontend/fingertips-frontend/components/organisms/BarChartEmbeddedTable/BarChartEmbeddedPlaceholderRows.tsx
```diff
@@ -0,0 +1,47 @@
+import React, { FC, RefObject, useLayoutEffect } from 'react';
+import { Table } from 'govuk-react';
+import { InViewTrigger } from '@/components/hooks/InViewTrigger';
+
+const getAverageHeight = () => {
+  const elements = document.getElementsByClassName('BarChartEmbeddedRow');
```
obarojohnson, line 6: I’m thinking that we could make this  variable className a constant, so if it’s mistakenly changed in the future, it won’t break existing functionality.

frontend/fingertips-frontend/components/organisms/BarChartEmbeddedTable/BarChartEmbeddedRow.tsx
```diff
@@ -31,7 +31,7 @@ export const BarChartEmbeddedRow: FC<BarChartEmbeddedRowProps> = ({
   measurementUnit = '',
 }) => {
   return (
-    <Table.Row key={`${item.area}`}>
+    <Table.Row key={`${item.area}`} className="BarChartEmbeddedRow">
```
obarojohnson, line 34: This className is use in multi-places could make it a constant so it could be share see first comment.

frontend/fingertips-frontend/components/organisms/BarChartEmbeddedTable/BarChartEmbeddedPlaceholderRows.tsx
```diff
@@ -0,0 +1,50 @@
+import React, { FC, RefObject, useLayoutEffect } from 'react';
+import { Table } from 'govuk-react';
+import { InViewTrigger } from '@/components/hooks/InViewTrigger';
+import { barChartEmbeddedRowClassName } from '@/components/organisms/BarChartEmbeddedTable/barChartEmbeddedTableHelpers';
+
+const getAverageHeight = () => {
```
obarojohnson, line 6: I was also thinking if we could use something like `calculateAverageVerticalSpacing()` or something that describes what the function does. Since you're doing a calculation, it should sound like we're retrieving a value from the store state.

obarojohnson (approved):

PR #564, opened 2025-05-09
DHSCFT-721: load synonym mappings into indicator index

adawari-thomas-bjss:
# Description

Jira ticket: [DHSCFT-721](https://bjss-enterprise.atlassian.net/browse/DHSCFT-721)

## Changes

- Added synonym maps json file
- Added a function that creates the synonym map in the search setup tool
- Attached synonym map to the indicatorName and indicator definition fields during index creation
- Added integration tests and playwright tests

## Validation

Manual testing on the deployed site without synonym mapping compared to localhost with synonym mapping

Search for `chd` on the deployed environment without synonym mapping produces 0 results

image

Search for `chd` on localhost with synonym mapping produces results

image

Search for equivalent synonyms diabetes and diabetic on localhost returns the same results

image
image

Search for diabetes and diabetic returns a single result each on the deployed site

image
image


search-setup/src/utils/indexHelper.ts
```diff
@@ -15,3 +20,33 @@ export async function createAndPopulateIndex<
...
+    name: synonymMapName,
+    synonyms: generateSynonyms(synonymData),
+  };
+  await indexClient.createSynonymMap(synonymMap);
+  console.log(`Created synonym map named: ${synonymMapName}`);
+}
+
+function generateSynonyms(synonymData: SynonymData): string[] {
+  return Object.entries(synonymData).map((wordMapping) => {
+    const keyword = wordMapping[0];
```
dlzhry2, line 46: Retrieving by index makes it a bit tricky to read. As a developer I have to go back to the Type and the original JSON file to work out what is happening. My recommendation would be to restructure the object like this so it's a bit more transparent:

`{
    keyword: "a&e",
    equivalentTerms: ["accident", "emergency"],
    isAbbrOrAcronym: true
}`

search-setup/src/utils/indexHelper.ts
```diff
@@ -15,3 +20,33 @@ export async function createAndPopulateIndex<
...
+    const synonymObject = wordMapping[1];
+    return synonymObject.isAcronym
+      ? `${keyword} => ${synonymObject.terms.join(', ')}`
```
dlzhry2, line 49: We should also include the term in here as per the acceptance criteria.

`copd => copd, chronic, obstructive, pulmonary, disorder` This will mean, if we get a query for `copd` we will still keep it in the list of terms we pass to Azure AI Search. This might help us if the acronym appears in the description but at the same time means we also get to use the expanded terms.

search-setup/src/types.ts
```diff
@@ -41,3 +41,11 @@ export interface AutoCompleteResult {
 export interface SuggestionResult {
   value: { '@search.text': string }[];
 }
+
+export type SynonymData = Record<
```
dlzhry2, line 45: As per other comment - recommend we make a minor adjustment to this structure.

search-setup/integration_tests/search.test.ts
```diff
@@ -135,6 +136,60 @@ describe('AI search index creation and data loading', () => {
...
+          headers: {
+            'api-key': apiKey,
+          },
+        });
+        const result = await response.json();
+        expect(result.name).toBe(INDICATOR_SEARCH_SYNONYM_MAP_NAME);
+        expect(result.synonyms).toMatchSnapshot();
+      });
+
+      it('should search by synonyms', async () => {
```
dlzhry2, line 153: This test relates to acronyms. I would also suggest this is a misleading test case and could have problems in future. E.g. if we introduce future indicator names to do with heart or coronary then we are going to get a mismatch.

See on the ticket, I had a look at the PoC indicators and the only valid acronym is IMD...we should add a definition for this as per the note on my ticket, and we could assert that the FIRST result is the Indices of Multiple Deprivation indicator. This is not perfect but might be less flaky.

search-setup/integration_tests/search.test.ts
```diff
@@ -135,6 +136,60 @@ describe('AI search index creation and data loading', () => {
...
+        const synonymResponse = await searchIndicatorsRequest('baby');
+        const synonymResult = await synonymResponse.json();
+
+        expect(wordResult.value).toEqual(synonymResult.value);
+      });
+
+      it('should map explicit synonyms one way', async () => {
+        const wordResponse = await searchIndicatorsRequest('chd');
+        const wordResult = await wordResponse.json();
+        const synonymResponse = await searchIndicatorsRequest('chronic');
```
dlzhry2, line 187: I am not sure what this is testing. Chronic is not present in the chd acronym.

search-setup/assets/synonyms.json
```diff
@@ -0,0 +1,274 @@
+{
+  "a&e": {
```
dlzhry2, line 2: See other comment: recommend we minorly adjust the structure and the names of the keys.

search-setup/assets/synonyms.json
```diff
@@ -0,0 +1,274 @@
...
+    "terms": ["accident", "emergency"],
+    "isAcronym": false
+  },
+  "depr": {
+    "terms": ["deprivation"],
+    "isAcronym": true
```
dlzhry2, line 8: Technically it's not an acronym, it's an abbreviation. Suggest we change the name of the key - but it's true we should treat this in the same way we treat acronyms.

search-setup/assets/synonyms.json
```diff
@@ -0,0 +1,274 @@
+{
+  "a&e": {
+    "terms": ["accident", "emergency"],
+    "isAcronym": false
```
dlzhry2, line 4: Technically it is an acronym right? But it could really be treated as either in the `solr` syntax, so don't mind.

frontend/fingertips-frontend/playwright/tests/fully_integrated_e2e_tests/ai_search_tests/indicator_search.spec.ts
```diff
@@ -64,5 +64,68 @@ test.describe(
...
+        await homePage.clickSearchButton();
+
+        await resultsPage.checkNumberOfResults(1);
+        await resultsPage.checkFirstResultHasName(
+          'Smokers at time of childbirth delivery'
+        );
+      });
+
+      await test.step('search for synonym of indicator and check results match', async () => {
+        const subjectSearchTerm = 'baby';
```
dlzhry2, line 88: Infant would be better. If you check the description for that indicator you'll find baby appears several times so it's working via a direct match rather than synonym.

frontend/fingertips-frontend/playwright/tests/fully_integrated_e2e_tests/ai_search_tests/indicator_search.spec.ts
```diff
@@ -64,5 +64,68 @@ test.describe(
...
+        await resultsPage.clickIndicatorSearchButton();
+
+        await resultsPage.checkNumberOfResults(1);
+        await resultsPage.checkFirstResultHasName(
+          'Smokers at time of childbirth delivery'
+        );
+      });
+    });
+
+    test('searching for the synonym of an explicitly mapped term should not return results of the mapped term', async ({
```
dlzhry2, line 101: It actually should. If I search for chd if that acronym appears in the indicator name or description I would like it to be returned.

Likewise, if chd does not appear but 'coronary' does appear somewhere I would like those examples to be returned. As per separate comment, I suggest we put in a mapping for IMD and test using that as we have no other acronyms in the indicators I spotted.

dlzhry2 (changes requested):
It's a good start. Some minor comments about the structure of the JSON. I also think we need a minor adjustment to the mappings.

And finally we need some valid test cases for the acronym/abbreviation mappings.

search-setup/assets/synonyms.json
```diff
@@ -0,0 +1,274 @@
+{
```
dlzhry2, line 1: Might be nice to have a docstring explaining the document structure and Azure synonums.

search-setup/src/utils/indexHelper.ts
dlzhry2: https://learn.microsoft.com/en-us/azure/search/search-synonyms?tabs=rest%2Crest-assign For further info on how the synonym mapping works I recommend.

search-setup/integration_tests/search.test.ts
```diff
@@ -135,6 +136,60 @@ describe('AI search index creation and data loading', () => {
...
+      it('should create synonym map', async () => {
+        const url = `${searchEndpoint}/synonymmaps/${INDICATOR_SEARCH_SYNONYM_MAP_NAME}${URL_SUFFIX}`;
+        const response = await fetch(url, {
+          headers: {
+            'api-key': apiKey,
+          },
+        });
+        const result = await response.json();
+        expect(result.name).toBe(INDICATOR_SEARCH_SYNONYM_MAP_NAME);
+        expect(result.synonyms).toMatchSnapshot();
```
HowardMayBjss, line 150: Best practice is to avoid the use of snapshots which require frequent update e.g. when data changes - in this case synonym data. If the intent of the test is to check that the synonym map has been created then this doesn't require the validation of all the synonyms. But you could check a small number are present.

search-setup/src/main.ts
```diff
@@ -15,9 +15,11 @@ import { parseIndicatorData } from './utils/indicatorDocumentHelpers.js';
 import {
   AREA_SEARCH_INDEX_NAME,
   INDICATOR_SEARCH_INDEX_NAME,
+  INDICATOR_SEARCH_SYNONYM_MAP_NAME,
 } from './constants.js';
 import rawAreasData from '../assets/areas.json' with { type: 'json' };
-import { createAndPopulateIndex } from './utils/indexHelper.js';
+import { createAndPopulateIndex, createSynonymMap } from './utils/indexHelper.js';
+import synonymData from '../assets/synonyms.json' with { type: 'json'}
```
HowardMayBjss, line 22: Suggest indicator-synonms.json

search-setup/integration_tests/search.test.ts
```diff
@@ -147,48 +147,78 @@ describe('AI search index creation and data loading', () => {
...
+      it('should do one way synonym mapping for keywords that have the explicit flag set', async () => {
         const wordResponse = await searchIndicatorsRequest('chd');
         const wordResult = await wordResponse.json();
-        const synonymResponse = await searchIndicatorsRequest('disease');
-        const synonymResult = await synonymResponse.json();
-
-        const expectedIndicatorNames: string[] = wordResult.value.map(
-          (indicatorResult: { indicatorName: string }) =>
-            indicatorResult.indicatorName
+        const requestKeywords = ['chd', 'coronary', 'heart', 'disease'];
```
gareth-allan-bjss, line 161: I think it might be clearer to remove `chd` from this list. Otherwise you're looking up the results for `chd` twice.  Also it would allow you to just assert that the length of `expectedIndicatorNamesForExplicitKeyword` equals `indicatorNamesPerSynonym.length`.

search-setup/src/utils/indexHelper.ts
```diff
@@ -42,11 +48,24 @@ export async function createSynonymMap(
...
+
+function generateMappingString(keyword: string, synonyms: string[]) {
+  const mappingString = [keyword, ...synonyms].join(', ');
+  return escapeSpecialCharacters(mappingString);
+}
+
+function escapeSpecialCharacters(mappingString: string) {
+  let resultString = mappingString;
+  SPECIAL_CHARS.forEach((char) => {
+    if (mappingString.includes(char)) {
```
gareth-allan-bjss, line 66: Why do you need to check if the string contains `char`? Doesn't `replaceAll` just do nothing if it doesn't?

search-setup/assets/indicator-synonyms.json
```diff
@@ -0,0 +1,335 @@
...
+    "synonyms": ["family"],
+    "isExplicit": false
+  },
+  {
+    "keyword": "diabetes",
+    "synonyms": ["diabetic"],
+    "isExplicit": false
+  },
+  {
+    "keyword": "cvd",
```
HowardMayBjss, line 168: This one isn't working as expected. The results we get back from cvd equate to the same results you get from 'disease' which is not the intent. I presume that the issue is that the synonmn is two separate words and the search is matching on either word. This requires understanding and some change to be made

HowardMayBjss (approved):

PR #586, opened 2025-05-20
DHSCFT-785: Display postcode in area suggestions

adawari-thomas-bjss:
# Description

Jira ticket: [DHSCFT-785](https://bjss-enterprise.atlassian.net/browse/DHSCFT-785)

## Changes

- Updated area suggestion to display postcode for areas that have their postcode available

## Validation

image
image

### DIFF
--- a/frontend/fingertips-frontend/components/molecules/AreaSuggestionPanel/__snapshots__/AreaSuggestionPanel.test.tsx.snap
+++ b/frontend/fingertips-frontend/components/molecules/AreaSuggestionPanel/__snapshots__/AreaSuggestionPanel.test.tsx.snap
@@ -206,6 +206,9 @@ exports[`AreaSuggestionPanel should render correctly and match snapshot 1`] = `
           <span>
             h
           </span>
+          <span>
+             
+          </span>
         </div>
         <div
           class="c8"
@@ -287,6 +290,9 @@ exports[`AreaSuggestionPanel should render correctly and match snapshot 1`] = `
           </span>
           <span>
             e
+          </span>
+          <span>
+             
           </span>
         </div>
         <div
--- a/frontend/fingertips-frontend/components/molecules/AreaSuggestionPanel/__snapshots__/AreaSuggestionPanel.test.tsx.snap
+++ b/frontend/fingertips-frontend/components/molecules/AreaSuggestionPanel/__snapshots__/AreaSuggestionPanel.test.tsx.snap
@@ -206,9 +206,6 @@ exports[`AreaSuggestionPanel should render correctly and match snapshot 1`] = `
           <span>
             h
           </span>
-          <span>
-             
-          </span>
         </div>
         <div
           class="c8"
@@ -290,9 +287,6 @@ exports[`AreaSuggestionPanel should render correctly and match snapshot 1`] = `
           </span>
           <span>
             e
-          </span>
-          <span>
-             
           </span>
         </div>
         <div
--- a/frontend/fingertips-frontend/lib/areaFilterHelpers/formatAreaName.ts
+++ b/frontend/fingertips-frontend/lib/areaFilterHelpers/formatAreaName.ts
@@ -3,9 +3,10 @@ import { gpsAreaType } from './areaType';
 export function formatAreaName(
   areaCode: string,
   areaName: string,
-  areaTypeKey: string
+  areaTypeKey: string,
+  postcode?: string
 ): string {
   return areaTypeKey.toLowerCase() === gpsAreaType.key
-    ? `${areaCode} - ${areaName}`
+    ? `${areaCode} - ${areaName} ${postcode ?? ''}`
     : areaName;
 }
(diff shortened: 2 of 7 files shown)


frontend/fingertips-frontend/lib/areaFilterHelpers/formatAreaName.ts
```diff
@@ -3,9 +3,10 @@ import { gpsAreaType } from './areaType';
...
   areaCode: string,
   areaName: string,
-  areaTypeKey: string
+  areaTypeKey: string,
+  postcode?: string
 ): string {
   return areaTypeKey.toLowerCase() === gpsAreaType.key
-    ? `${areaCode} - ${areaName}`
+    ? `${areaCode} - ${areaName} ${postcode ?? ''}`
     : areaName;
```
gareth-allan-bjss, line 11: I think we want to display the postcode if it's present for other area types than GPs. PCNs also have a postcode (and will be suggested based on it), so we should be showing the postcode to make it clear to a user why the result is being returned.

frontend/fingertips-frontend/lib/areaFilterHelpers/formatAreaName.ts
```diff
@@ -3,9 +3,10 @@ import { gpsAreaType } from './areaType';
 export function formatAreaName(
   areaCode: string,
   areaName: string,
-  areaTypeKey: string
+  areaTypeKey: string,
+  postcode?: string
```
gareth-allan-bjss, line 7: At this point, I think I'd suggest passing in the entire `AreaDocument` object as a parameter to this function. 

Perhaps it would also be a good opportunity to name the function something more representative of what it's doing as well, given it's not really formatting just the area name any more. Perhaps something like `generateAreaDisplayString`?

frontend/fingertips-frontend/lib/areaFilterHelpers/formatAreaName.ts
```diff
@@ -3,9 +3,10 @@ import { gpsAreaType } from './areaType';
...
 ): string {
   return areaTypeKey.toLowerCase() === gpsAreaType.key
-    ? `${areaCode} - ${areaName}`
+    ? `${areaCode} - ${areaName} ${postcode ?? ''}`
```
gareth-allan-bjss, line 10: If the area type is "GP" but `postcode` isn't specified this is going to result in an area with a trailing string. We should also have a test for this scenario.

frontend/fingertips-frontend/components/molecules/AreaSuggestionPanel/__snapshots__/AreaSuggestionPanel.test.tsx.snap
```diff
@@ -288,6 +291,9 @@ exports[`AreaSuggestionPanel should render correctly and match snapshot 1`] = `
           <span>
             e
           </span>
+          <span>
+
```
gareth-allan-bjss, line 295: Looks like this may be the trailing space cropping up.

frontend/fingertips-frontend/lib/areaFilterHelpers/formatAreaName.ts
adawari-thomas-bjss: This function is also being used by the `AreaSelectedPill` component, which uses an area of type `AreaWithRelations`. So the less complicated solution is to pass in the string that should be formatted, instead of working to figure out whether its an area of type `AreaDocument` or type `AreaWithRelations` within the function. But if you feel strongly towards it, I can make the change.

frontend/fingertips-frontend/lib/areaFilterHelpers/generateAreaDisplayString.test.ts
```diff
@@ -0,0 +1,36 @@
...
+      'A001',
+      'Some name',
+      gpsAreaType.key as AreaTypeKeys,
+      'Some postcode'
+    );
+
+    expect(formattedAreaName).toEqual('A001 - Some name Some postcode');
+  });
+
+  it('should just return the areaName provided when the areaTypeKey is not gps', () => {
```
gareth-allan-bjss, line 16:
```suggestion
  it('should return the areaName and postcode provided when the areaTypeKey is not gps', () => {
```

frontend/fingertips-frontend/lib/areaFilterHelpers/generateAreaDisplayString.test.ts
```diff
@@ -0,0 +1,36 @@
+import { AreaTypeKeys, gpsAreaType, nhsRegionsAreaType } from './areaType';
+import { generateAreaDisplayString } from './generateAreaDisplayString';
+
+describe('generateAreaDisplayString', () => {
+  it('should return the areaCode and areaName provided when the areaTypeKey is gps', () => {
```
gareth-allan-bjss, line 5:
```suggestion
  it('should return the areaCode, areaName and postcode provided when the areaTypeKey is gps', () => {
```

gareth-allan-bjss (approved):
Looks good, just a couple of test name tweaks needed I think.

gareth-allan-bjss (approved):

PR #587, opened 2025-05-20
DHSCFT-738: Update Quartiles API to allow benchmarking against AreaGroup

HowardMayBjss:
# Description

Jira ticket: [DHSCFT-738](https://bjss-enterprise.atlassian.net/browse/DHSCFT-738)

The benchmarking updates require the API to provide Quartile values calculated for an AreaGroup rather than for England as a whole.

NOTE: This PR is based on 714 which is expected to be merged first - so at present this appears to hold the content of both,

## Changes

- Update swagger for Quartiles API to include benchmark-ref-type
- Changed AncestorCode to AreaGroup for consistency with Indicators API
- Updated Service to handle new parameter and pass to repository / SQL
- Updated SQL to allow the benchmark areaGroup to be specified.
- Updated the SQL to select areas only from within the AreaGroup.
- Added http tests to validate new functionality

## Validation

Http scripts updated to validate new functionality


obarojohnson (approved):

database/fingertips-db/StoredProcedures/GetIndicatorQuartileDataForLatestYear.sql
```diff
@@ -3,10 +3,19 @@ CREATE PROCEDURE [dbo].[GetIndicatorQuartileDataForLatestYear]
 @RequestedAreaType varchar(50),
```
obarojohnson, line 3: --- This stored procedure Gets HealthData and performs **Quintile** calculations.   I think this should be Quartiles right?

DataCreator/DataCreator/DataFileReader.cs
```diff
@@ -53,7 +53,7 @@ public static List<HealthMeasureEntity> GetHealthDataForIndicator(int indicatorI
                 if (split.Length != 27)
                     continue; //avoid bad data
 
-                if (split[6].Equals("CCG", StringComparison.CurrentCultureIgnoreCase)) //CCGs and ICBs share the same area code so do this to avoid doubling up data
+                if (split[6].Trim().Equals("CCG", StringComparison.CurrentCultureIgnoreCase)) //CCGs and ICBs share the same area code so do this to avoid doubling up data
```
obarojohnson, line 56: Does that means any area type that is "CCG" will be ignored? cause this will throw away all the data points that're associated with the area type right?

HowardMayBjss: Yes - we don't support the CCG area type.

PR #593, opened 2025-05-22
DHSCFT-790: highlight search matches

adawari-thomas-bjss:
# Description

Jira ticket: [DHSCFT-790](https://bjss-enterprise.atlassian.net/browse/DHSCFT-790)

## Changes

- Updated highlighting logic to use AI search highlighting suggestions

## Validation

Manual validation with actual AI search 
image
image

Manual validation against mock AI search
image


frontend/fingertips-frontend/components/molecules/AreaSuggestionPanel/AreaSuggestionPanel.test.tsx
```diff
@@ -13,10 +13,19 @@ import { SearchStateContext } from '@/context/SearchStateContext';
...
+    document: { areaCode: 'GP01', areaName: 'Greenwich', areaType: 'GPs' },
+  },
+  {
+    text: '',
+    document: { areaCode: 'GP02', areaName: 'Cambridge', areaType: 'GPs' },
+  },
+  {
+    text: '',
+    document: { areaCode: 'CT01', areaName: 'Central London', areaType: 'CT' },
+  },
```
gareth-allan-bjss, line 28: Again, may as well put some actual values in the `text` property here.

frontend/fingertips-frontend/components/atoms/HighlightText/index.tsx
```diff
@@ -1,51 +1,43 @@
...
+        index < highlightStartPos + textToBeHighlighted.length
+      ) {
+        accumulator.push(
+          <StyleHighLightedText key={char + index}>{char}</StyleHighLightedText>
+        );
+      } else {
+        accumulator.push(<span key={char + index}>{char}</span>);
+      }
+      return accumulator;
+    }, []);
```
gareth-allan-bjss, line 34: This does result in every character in the string being wrapped in its own `<span>` element, which I think seems unnecessary. Really, all we'd need is the text to be highlighted to be wrapped in a `StyleHighLightedText` span.

Not necessarily something to spend ages on if it proves complex to change, as it was this way before, but might be worth a quick look.

frontend/fingertips-frontend/components/atoms/HighlightText/HighlightText.test.tsx
```diff
@@ -1,26 +1,36 @@
 import { render } from '@testing-library/react';
 import { HighlightText } from '.';
+import { highlightTag } from '@/lib/search/searchTypes';
 
 describe('HighlightText', () => {
```
gareth-allan-bjss, line 5: I think your implementation would handle having an instance of the highlight tag in the middle of the search hint (which is nice!), so it might be good to have a test that ensures this remains the case.

frontend/fingertips-frontend/lib/search/areaSearchServiceMock.ts
```diff
@@ -35,6 +40,15 @@ export class AreaSearchServiceMock implements IAreaSearchService {
...
         );
       })
+      .map((mockArea) => ({
+        text: `${highlightTag}${partialAreaName}${highlightTag}`,
+        document: {
+          areaCode: mockArea.areaCode,
+          areaName: mockArea.areaName,
+          areaType: mockArea.areaType,
+          postcode: mockArea.postcode,
+        },
```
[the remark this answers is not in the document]
adawari-thomas-bjss: this didn't seem to work initially, but it works now, so I must've missed something

frontend/fingertips-frontend/lib/areaFilterHelpers/generateAreaDisplayString.ts
```diff
@@ -11,4 +11,5 @@ export function generateAreaDisplayString(
     : areaName + formatPostcode(postcode);
 }
 
-const formatPostcode = (postcode?: string) => (postcode ? ' ' + postcode : '');
+const formatPostcode = (postcode?: string) =>
+  postcode ? '\u00A0 ' + postcode : '';
```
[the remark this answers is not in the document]
adawari-thomas-bjss: There's been an additional requirement to add an extra space between the area name and the postcode after the BA review for DHSCFT-785. I've used the non-breaking space to add the extra space, as HTML just ignores an extra space character

frontend/fingertips-frontend/components/atoms/HighlightText/index.tsx
gareth-allan-bjss: This is definitely an improvement, but it's still wrapping each of the highlighted characters in their own `<span>` element. How about this for `highlightChars` instead?

```typescript
const highlightChars = (text: string, searchHint: string) => {
  const highlightStart = searchHint.indexOf(highlightTag);
  const hightlightEnd = searchHint.lastIndexOf(highlightTag);

  const textToBeHighlighted = searchHint.slice(
    highlightStart + highlightTag.length,
    hightlightEnd
  );

  const highlightStartPos = text
    .toLowerCase()
    .indexOf(textToBeHighlighted.toLowerCase());
  const highlightEndPos = highlightStartPos + textToBeHighlighted.length;
  return [
    text.slice(0, highlightStartPos),
    <StyleHighLightedText key="searchMatch">
      {textToBeHighlighted}
    </StyleHighLightedText>,
    text.slice(highlightEndPos),
  ];
};
```

gareth-allan-bjss (approved):
Looks good other than my proposed improvement to the highlighting.

gareth-allan-bjss (approved):

PR #601, opened 2025-05-23
DHSCFT-TECH: tweak to have benchmark select area per view

alibombaye:
# Description

Due to complexity issue decided as an interim solution to have benchmark area selection at view level rather than at chart level

Screenshot 2025-05-23 at 15 16 09

## Validation

Tested One indicator One Area View benchmark selection still works as expected


frontend/fingertips-frontend/playwright/testHelpers.ts
```diff
@@ -87,7 +87,7 @@ export function getScenarioConfig(
       componentLocator: ChartPage.lineChartComponent,
       componentProps: {
         hasConfidenceIntervals: true,
-        hasBenchmark: true,
+        // hasBenchmark: true,
```
alibombaye, line 90: Temporary disable the benchmark e2e testing after moving the benchmark area select to top of the view page.

frontend/fingertips-frontend/lib/searchStateManager.ts
```diff
@@ -16,7 +16,8 @@ export enum SearchParams {
   PopulationAreaSelected = 'pas',
   SearchedOrder = 'so',
   PageNumber = 'pn',
-  LineChartBenchmarkAreaSelected = 'lcbas',
+  OneIndOneAreaViewBenchmarkSelected = 'oioabs',
```
Joe-Hendley, line 19: if we're only ever going to have one view displayed at a time, why do we need seperate parameters for each view?

alibombaye: That's a good catch. Will update 👍

alibombaye: Updated

Joe-Hendley (approved):

PR #629, opened 2025-06-02
DHSCFT-773: add linting to api

adawari-thomas-bjss:
# Description

Jira ticket: [DHSCFT-733](https://bjss-enterprise.atlassian.net/browse/DHSCFT-733)

## Changes

- formatted API code
- made manual changes to fix lint errors 
- added step to check api code format to workflow


api/DHSC.FingertipsNext.Modules.HealthData.Repository/HealthDataRepository.cs
```diff
@@ -156,10 +159,9 @@ public async Task<IEnumerable<HealthMeasureModel>> GetIndicatorDataWithQuintileB
 
         var YearsTable = new DataTable();
```
Mr-Ev-H, line 160: With this one, github is asking for the datatable to be disposed properly, you can solve the github warnings by wrapping the creation of the YearsTable in a using block, 

e.g.

`using (var YearsTable = new DataTable())
{
// Code that uses it here
}`

It's also being warned of in the use of the IndicatorIdTable further below.

api/DHSC.FingertipsNext.Modules.HealthData.Repository/Models/HealthMeasureModel.cs
```diff
@@ -54,6 +54,6 @@ public class HealthMeasureModel
     public bool IsDeprivationAggregatedOrSingle { get; set; } = true;
 
     public bool IsAggregate { get; set; }
-    [NotMapped] 
+    [NotMapped]
     public BenchmarkComparisonModel? BenchmarkComparison { get; set; }
```
Mr-Ev-H, line 58: Don't need the nullability check here - GH is complaining because unless the file or the project is enabled for nullability this doesnt do anything

.github/workflows/fingertips-workflow.yml
```diff
@@ -150,6 +150,10 @@ jobs:
 
       - name: Restore Packages
         run: dotnet restore
+        
+      - name: Check API Code Format
+        run: dotnet format --verify-no-changes
+        continue-on-error: true
```
gareth-allan-bjss, line 156: I know the original proposal was to continue on error, so that we didn't have to fix every error immediately, but do you have a feel for what's left to fix before we can enforce this formatting on build? I think it would be good to get a ticket (or multiple tickets if there's an obvious split) in the backlog to do these fixes, as we should aim to get to a point where formatting failures result in build failures, like for the frontend.

Mr-Ev-H (approved):
API code is looking much better, GJ

HowardMayBjss (approved):

PR #639, opened 2025-06-04
DHSCFT-814: Dynamic title for line chart and table

stuc-bjss:
# Description

Jira ticket: [DHSCFT-814](https://bjss-enterprise.atlassian.net/browse/DHSCFT-814)

Added title

## Changes

- A ChartTitle atom component
- Determine chart title when prepping highcharts options

## Validation

image
image
image
image

### DIFF
--- a/frontend/fingertips-frontend/components/organisms/LineChart/helpers/generateStandardLineChartOptions.ts
+++ b/frontend/fingertips-frontend/components/organisms/LineChart/helpers/generateStandardLineChartOptions.ts
@@ -14,6 +14,7 @@ import { generateXAxis } from './generateXAxis';
 import { generateTooltip } from './generateTooltip';
 import { generateAccessibility } from './generateAccessibility';
 import { generateSeriesData } from './generateSeriesData';
+import Highcharts from 'highcharts';
 
 export enum LineChartVariant {
   Standard = 'standard',
@@ -74,6 +75,7 @@ export function generateStandardLineChartOptions(
   lineChartCI: boolean,
   benchmarkToUse: string,
   optionalParams?: {
+    indicatorName?: string;
     englandData?: HealthDataForArea;
     groupIndicatorData?: HealthDataForArea;
     yAxisTitle?: string;
@@ -122,8 +124,19 @@ export function generateStandardLineChartOptions(
         }
       : sortedGroupData;
 
+  const fromTo = `from ${firstYear} to ${lastYear}`;
+  const titleText = optionalParams?.indicatorName
+    ? `${optionalParams?.indicatorName} ${fromTo}`
+    : fromTo;
+
   return {
     ...lineChartDefaultOptions,
+    title: {
+      text: titleText,
+      style: {
+        display: 'none',
+      },
+    },
     yAxis: generateYAxis(optionalParams?.yAxisTitle),
     xAxis: generateXAxis(
       optionalParams?.xAxisTitle,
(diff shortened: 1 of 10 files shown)


frontend/fingertips-frontend/components/organisms/LineChart/index.tsx
```diff
@@ -33,12 +36,13 @@ export function LineChart({
   const id = `${variant}LineChart-component`;
   return (
     <div data-testid={id}>
-      <ConfidenceIntervalCheckbox
-        chartName={`${variant}LineChart`}
-        showConfidenceIntervalsData={showConfidenceIntervalsData}
-        setShowConfidenceIntervalsData={setShowConfidenceIntervalsData}
-      />
       <div id={id}>
```
Joe-Hendley, line 39: do we need both of these wrappers if we've moved the confidence interval checkbox inside?

frontend/fingertips-frontend/components/organisms/LineChart/index.tsx
```diff
@@ -9,13 +9,16 @@ import { HighChartsWrapper } from '@/components/molecules/HighChartsWrapper/High
 import { useState } from 'react';
 import { ExportCopyright } from '@/components/molecules/Export/ExportCopyright';
 import { ExportOnlyWrapper } from '@/components/molecules/Export/ExportOnlyWrapper';
+import { ChartTitle } from '@/components/atoms/ChartTitle/ChartTitle';
 
 interface LineChartProps {
+  title?: string;
```
Joe-Hendley, line 15: when would we render a line chart without a title?

stuc-bjss: at the moment just on the inequalities line chart but only because it's title is in another ticket

frontend/fingertips-frontend/components/organisms/LineChart/helpers/generateStandardLineChartOptions.ts
```diff
@@ -122,8 +124,19 @@ export function generateStandardLineChartOptions(
...
+    : fromTo;
+
   return {
     ...lineChartDefaultOptions,
+    title: {
+      text: titleText,
+      style: {
+        display: 'none',
+      },
+    },
```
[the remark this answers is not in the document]
stuc-bjss: it's convenient as the generateStandardLineChartOptions already does the determination of from and to periods, also there is another ticket on the way to add the title to the svg export so it will be used then

frontend/fingertips-frontend/components/molecules/Inequalities/InequalitiesTrend/index.tsx
```diff
@@ -228,6 +228,7 @@ export function InequalitiesTrend({
             title: 'Line chart',
             content: (
               <LineChart
+                title={''}
```
stuc-bjss, line 231: there is another ticket to place a title here

frontend/fingertips-frontend/components/organisms/LineChart/helpers/generateStandardLineChartOptions.ts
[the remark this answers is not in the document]
gareth-allan-bjss: OK, if we're going to use it for the SVG generation it makes sense then. I'd have been wary of just using the highcharts object as a way of passing an unrelated title around, but if it needs to be in there it makes sense.

Joe-Hendley (approved):

PR #641, opened 2025-06-04
DHSCFT-875: Date Period Updates

HowardMayBjss:
# Description

Jira ticket: [DHSCFT-875](https://bjss-enterprise.atlassian.net/browse/DHSCFT-875)

This PR proposes changes to the Indicators Service API to support different time period types which have recently been added to the database.

This will require year to be reinstated before merging - I left it out for the purpose of this review. I also don't think it makes sense to merge this until we have the Service implementing it.

## Changes

- Data types for Dates and PeriodType added
- Query fields for to and from and period-type added to Indicators/data API
- Query field for period-type added to Indicators/quartile API
- Year replaced with to and from in results
- Behaviour called out in endpoint description where appropriate

## Validation

No validation has been performed

### DIFF
--- a/api/definition/swagger.yaml
+++ b/api/definition/swagger.yaml
@@ -243,12 +243,23 @@ paths:
         - indicators
       summary: Get health data for an indicator
       description:
-        Get data for a public health indicator. This will return all data for all areas and all years for the indicators. Optionally filter the results by supplying one
-        or more area codes and one or more years in the query string.
+        Get data for a public health indicator. This will return all data for
+        all requested areas and all time periods for the specified indicator and for a single 
+        period-type e.g. Calendar year. Optionally filter the results by 
+        supplying to and from time periods in the query string. If the 
+        Indicator supports data for more than one time period type, the 
+        period_type query parameter can be used to specify which period type to
+        return data for.
+
+        The returned data will include a benchmark comparison against either England or an ancestor area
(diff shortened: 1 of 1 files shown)


api/definition/swagger.yaml
```diff
@@ -529,24 +556,33 @@ components:
...
     HealthDataPoint:
       type: object
       description:
         Represents a health data point for a public health indicator with a
         count, value, upper confidence interval, lower confidence interval,
-        year, age band and sex.
+        date range, age band and sex.
       required:
-        - year
+        - from-date
```
stuc-bjss, line 571: from_date?

api/definition/swagger.yaml
```diff
@@ -529,24 +556,33 @@ components:
...
+        - to-date
```
stuc-bjss, line 572: to_date?

stuc-bjss (commented):
Do we need to say something about wether of not the dates are inclusive, would seem like an obvious yes for the fromDate, but not so obvious for the end date?!

api/definition/swagger.yaml
```diff
@@ -756,17 +809,26 @@ components:
         The server could not understand the request due to invalid syntax.
         The client should modify the request and try again.
   parameters:
-    years:
+    from_date:
```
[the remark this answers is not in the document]
HowardMayBjss: There are places this is used

PR #642, opened 2025-06-04
DHSCFT-854: Keep items focused after selection triggers a reload

stuc-bjss:
# Description

Jira ticket: [DHSCFT-854](https://bjss-enterprise.atlassian.net/browse/DHSCFT-854)

Add a generic hook to use url fragments to maintain element focus when the page reloads due to selection change

## Changes

- Added FocusOnFragment component
- Added id's to checkbox and select elements on results page
- Updated target paths to include a hash fragment

### DIFF
--- a/frontend/fingertips-frontend/components/atoms/FocusOnFragment/FocusOnFragment.tsx
+++ b/frontend/fingertips-frontend/components/atoms/FocusOnFragment/FocusOnFragment.tsx
@@ -1,12 +1,12 @@
 'use client';
 
 import { FC, useEffect } from 'react';
-import { useSearchParams } from 'next/navigation';
 import { useLoadingState } from '@/context/LoaderContext';
+import { useSearchParams } from 'next/navigation';
 
 export const FocusOnFragment: FC = () => {
-  const query = useSearchParams();
   const loadingState = useLoadingState();
+  const searchParams = useSearchParams();
 
   useEffect(() => {
     if (!window) return;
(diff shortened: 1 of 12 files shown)


frontend/fingertips-frontend/components/atoms/FocusOnFragment/FocusOnFragment.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { FC, useEffect } from 'react';
+import { useLoadingState } from '@/context/LoaderContext';
+import { useSearchParams } from 'next/navigation';
+
+export const FocusOnFragment: FC = () => {
+  const loadingState = useLoadingState();
+  const searchParams = useSearchParams();
+
+  useEffect(() => {
+    if (!window) return;
```
HowardMayBjss, line 12: Probably a novice question - but given we have 'use client' set then in what circumstances will window not exist?

Joe-Hendley (approved):
Confident this works in isolation but there's some cases due to lazy loading / page layout where we might need to take a look at how it all meshes together

HowardMayBjss:
I've run this up locally and am clicking on the indicators on the results page. Each time I do the page flashes which I find uncomfortable. Has that been introduced by this change?

Joe-Hendley:
> I've run this up locally and am clicking on the indicators on the results page. Each time I do the page flashes which I find uncomfortable. Has that been introduced by this change?

Looking at it on main, no. I believe that's the loading screen and we should probably raise a separate ticket to look at that behaviour.

PR #646, opened 2025-06-06
DHSCFT-764: Added unit test

HowardMayBjss:
# Description

Jira ticket: [DHSCFT-764](https://bjss-enterprise.atlassian.net/jira/software/projects/DHSCFT/boards/823?selectedIssue=DHSCFT-764)

This adds a unit test for the work associated with this ticket
## Changes

- Unit test for validating that CCG and STP areas are excluded from results

### DIFF
--- a/DataCreator/DataCreator.Tests/DataFileReaderTests.cs
+++ b/DataCreator/DataCreator.Tests/DataFileReaderTests.cs
@@ -1,9 +1,12 @@
 using System.Collections.Generic;
 using System.IO;
-using System.Linq;
 using DataCreator;
 using NUnit.Framework;
 
+namespace DataCreator.Tests;
+
+[TestFixture]
+[TestOf(typeof(DataFileReader))]
 public class DataFileReaderTests
 {
     [Test]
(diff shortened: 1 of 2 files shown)


DataCreator/DataCreator.Tests/DataFileReaderTests.cs
```diff
@@ -0,0 +1,48 @@
...
+        {
+            { "AREA3", "Area 3" }
+        };
+
+        // Act
+        var result = DataFileReader.GetHealthDataForIndicator(indicatorId, areasDict);
+
+        // Assert
+        Assert.That(result, Has.Exactly(1).Items);
+        Assert.That(result[0].AreaCode, Is.EqualTo("AREA3"));
```
williamwillis11, line 41: is it worth adding a couple of assertions that CCG and STP isnt hiding in there?

williamwillis11: i know this ^^^ should be sufficient - but thinking, since we are here anyway - that a check that the file doesnt contain CCG or STP would be good

HowardMayBjss: Actually, the output file doesn't specify the area type it is for so this test is not possible. The content of the Health Measure output is shown here 
image

williamwillis11 (approved):

PR #649, opened 2025-06-09
DHSCFT-751: Update chart width and make cols size dynamic

alibombaye:
# Description

Jira ticket: [DHSCFT-751](https://bjss-enterprise.atlassian.net/browse/DHSCFT-751)

Updates the chart page to take a larger viewable width on larger screens.
Updates the size of area filter column depending on screen size.

NOTE: the header and footer widths will be handled in a seperate PR.

## Validation

Screenshot 2025-06-09 at 10 22 48
^ Large screens sets width to 80vw and sets the areaFilter col size to one-quarter from one-third. Since there we are using more width we don't need the areaFilter to be as large.

Screenshot 2025-06-09 at 10 23 13
^ smaller screens sets width to 90vw and sets the areaFilter col size to one-third.

Screenshot 2025-06-09 at 10 26 54
^ still remain responsive when screen size is tablet/mobile


frontend/fingertips-frontend/components/pages/chartPageWrapper/index.tsx
```diff
@@ -44,10 +44,42 @@ export function ChartPageWrapper({
...
+          width: 90,
+          areaFilterCol: 'one-third',
+          chartCol: 'two-thirds',
+        });
+      }
+    }
+    window.addEventListener('resize', handleResize);
+    handleResize();
+    return () => window.removeEventListener('resize', handleResize);
+  }, []);
```
gareth-allan-bjss, line 72: What's this going to mean for users without JS enabled?

frontend/fingertips-frontend/components/pages/chartPageWrapper/index.tsx
```diff
@@ -44,10 +44,42 @@ export function ChartPageWrapper({
...
+
+  const width = !isHideFilters ? containerStyle.chartCol : 'full';
```
gareth-allan-bjss, line 74: I think now we're dealing with multiple sizes/widths it would be good to give `width` a more specific name. What's it the width of? `pageWidth`? `displayWidth`?

frontend/fingertips-frontend/components/pages/chartPageWrapper/index.tsx
alibombaye: The page doesn't work for non-JS anyway. We get stuck in the loading state. But what I can see is that the resizing after page load won't fully work.

alibombaye: Yes the responsiveness doesn't fully work in this situation.

I did think of using the CSS and the `@media` rule but I'm not quite sure how I would dynamically adjust the widths of the areaCol and the chartCol to be one-third/two-thirds or one-quarter/three-quarters depending on the screen size. 

Happy to keep digging into this. This seemed like it was the simplest and readable way of implementing this behaviour without too much CSS changes.

alibombaye: OK I think I have a CSS alternative solution.

gareth-allan-bjss (approved):

PR #659, opened 2025-06-10
DHSCFT-920: Create indicator management DB schema 

adawari-thomas-bjss:
# Description

Jira ticket: [DHSCFT-920](https://bjss-enterprise.atlassian.net/browse/DHSCFT-920)

## Changes

- Created db schema for indicator manager
- Created batch table script
- Moved existing sql scripts to appropriate module folders

## Validation

image


database/fingertips-db/Security/UploadManager.sql
```diff
@@ -0,0 +1 @@
+CREATE SCHEMA [UploadManager]
```
gareth-allan-bjss, line 1: I think we want to think about the name `UploadManager`, as it's not really a term we've used for this functionality so far. We should align with what @WolfieKnee is calling the API module as well, as I'd like the DB schema to align with the module name.

database/fingertips-db/Tables/Area/AreaDimension.sql
gareth-allan-bjss, line 1: This is actually part of the `IndicatorData` tables (you can see it's creating `dbo.AreaDimension` rather than `Areas.AreaDimension`.

database/fingertips-db/Tables/Area/AreaRelationships.sql
gareth-allan-bjss, line 1: The DB schema is called `Areas`, so I think we should have these files in a directory called `Areas` (rather than the singular `Area` you have at the moment).

database/fingertips-db/Tables/IndicatorData/AgeDimension.sql
gareth-allan-bjss, line 1: We call the API module `HealthData` so I think these files should probably live in a folder with the same name.

database/fingertips-db/Tables/UploadManager/Batch.sql
```diff
@@ -0,0 +1,15 @@
+--This stores a batch of health data points belonging to a single indicator that have been uploaded together.
```
gareth-allan-bjss, line 1: This table doesn't actually store health data points, so I think this comment is misleading. I think it's really storing the details (or metadata) of the batch of health data points that were uploaded.

database/fingertips-db/Tables/UploadManager/Batch.sql
```diff
@@ -0,0 +1,15 @@
...
+CREATE TABLE [UploadManager].[Batch]
+(
+    [BatchId] [int] IDENTITY(1,1) NOT NULL,  -- Surrogate key for the batch
```
gareth-allan-bjss, line 4: The format of the `BatchId` column needs to align with what @ehsh1 is doing for his changes to the HealthMeasure table. I believe we'd agreed on using a `varchar` (so we could include the indicator ID and a timestamp, to make things more readable).

database/fingertips-db/Tables/UploadManager/Batch.sql
```diff
@@ -0,0 +1,15 @@
...
+    [IndicatorKey] [smallint] NOT NULL,  -- Foreign key to the indicator this batch belongs to
```
gareth-allan-bjss, line 5: As this table is in a different schema, I don't think we should reference the key of the indicator in the `IndicatorDimension` table. The idea with these API modules is that they could be split out into separate applications if needed, so this would make that harder.

I do think we should store the indicator ID a batch is for in this table though. Perhaps just as an `IndicatorId` column matching the one in the `IndicatorDimension` though.

database/fingertips-db/Tables/UploadManager/Batch.sql
```diff
@@ -0,0 +1,15 @@
...
+    [Created] [datetime2](7) NOT NULL DEFAULT GETDATE(),  -- The date and time the batch was created
+    [Published] [bit] NOT NULL DEFAULT 0,  -- Whether the batch has been published
```
gareth-allan-bjss, line 7: I'm not sure we need a published flag in this batch table. I don't know when we would change it, as the idea is we rely on a publication date for each health data point to determine whether data should be published or not, rather than any process that would set a batch as published.

database/fingertips-db/Security/UploadManager.sql
andyevans-bjss: Agreed, not sure UploadManager is the correct name

database/fingertips-db/Tables/IndicatorData/AgeDimension.sql
andyevans-bjss: do we need to change these at all for this PR?

database/fingertips-db/Tables/UploadManager/Batch.sql
andyevans-bjss: Agreed, we shouldn't have referential integrity between our schemas as Gareth says. 
Having different schemas in the same database is a halfway house, you get the benefit of different modules owning their own independent data store but without the hassle of creating a dedicated database for each module which could incur extra hosting costs and management effort. You can then scale out in the future if needed and move the schema to a dedicated database.

andyevans-bjss (changes requested):

database/fingertips-db/Tables/IndicatorData/AgeDimension.sql
gareth-allan-bjss: It's not a must-have, but I'd suggested that it might be nice to introduce a bit of organisation into the DB project, again just to make it easier to split up in future if required.

PR #665, opened 2025-06-12
Feature/DHSCFT-904 e2e export testing

williamwillis11:
# Description

Jira ticket: [DHSCFT-904](https://bjss-enterprise.atlassian.net/browse/DHSCFT-904)

add e2e test coverage for the Export functionality added in Sprint 9

BE AWARE - as i changed some data test ids i deleted all the screenshots locally before regenerating them in the CI job (so that we didnt get left with a load of old ones that we no longer needed) - apologies 

## Changes

- made ids consistent across all charts + updated their snapshot tests to reflect this change
- above was so that we can select the export button easily in all the tests
- updated e2e playwright journey config to cover export testing for 3 of the journeys
- testing all three of the types of export (PNG, SVG and CSV) in those 3 journeys across most of the charts
- testing the previews for the 3 types of exports
- refactored selected area filters and areas to make it clearer
- refactored the e2e playwright journey config to remove area filters to select for journeys that don't need it i.e. SearchMode.BOTH_SUBJECT_AND_AREA or SearchMode.ONLY_AREA 
- removed logScenarioInfo as it wasnt providing value and created a lot of noise

## Validation
E2E tests passing locally and in CI


frontend/fingertips-frontend/.gitignore
```diff
@@ -12,6 +12,9 @@
 
 # testing
 /coverage
+.test/spec/exports/png/*/*/**.png
```
Mr-Ev-H, line 15: is it possible to ignore everything under the exports folder here? rather than doing it per file type

frontend/fingertips-frontend/playwright.config.ts
```diff
@@ -15,8 +15,8 @@ const config: PlaywrightTestConfig = {
   fullyParallel: true,
   forbidOnly: isCI, // fails the build on CI if you accidentally left test.only in the source code
   retries: isCI ? 1 : 0,
-  workers: isCI ? 2 : '50%', // 50% of the available CPUs
-  timeout: 60_000,
+  workers: isCI ? 2 : '25%', // 50% of the available CPUs
```
Joe-Hendley, line 18: is the comment correct?

williamwillis11: good spot will update

PR #672, opened 2025-06-16
DHSCFT-980: Add an Azurite container to local Docker stack

gareth-allan-bjss:
# Description

Jira ticket: [DHSCFT-980](https://bjss-enterprise.atlassian.net/browse/DHSCFT-980)
Subtask: [DHSCFT-1009](https://bjss-enterprise.atlassian.net/browse/DHSCFT-1009)

Add an Azurite container to the `compose.yaml` file, so that Azurite is started alongside the application. Also adds a single-run container that uses the Azure CLI to create a storage container in Azurite for the application to use.

## Changes

- Added a `storage` container that runs Azurite
  - This container only runs the blob storage service (as opposed to the table and queue services, which we won't need)
  - This is run as part of the `all`, `api` and (new) `storage` profiles
- Added a `storage-setup` container that uses the Azure CLI to create a storage container for the application to use
- Added a `storage` profile that runs only the `storage` and `storage-setup` containers
- Added a dependency on the `storage-setup` container being started to the `api` container
  - This is a `service_started` dependency, rather than a `service_healthy` check (which uses a healthcheck to make sure the service has properly started up, as the container is relatively quick to start, and we won't actually make requests to the storage account until a request is made to the relevant API endpoint
- Set a `STORAGE_CONNECTION_STRING` and `STORAGE_CONTAINER_NAME ` environments variable for the `api` container, to provide storage account/container connection details
- Added default values for the storage account connection string and container to `.env.template`
  - The connection string uses the `storage` hostname for the storage container on the internal Docker network
  -  The connection string also uses the [publicly available](https://learn.microsoft.com/en-us/azure/storage/common/storage-use-azurite?tabs=npm%2Cblob-storage#well-known-storage-account-and-key) Azurite well-known storage account and key. 

## Validation

Ran `docker compose --profile all up --build --remove-orphans -d` to start up the application:
Screenshot 2025-06-16 at 16 24 09

Verified that I could connect to the container using [Azure Storage Explorer](https://azure.microsoft.com/en-us/products/storage/storage-explorer) and see the `fingertips-upload-container` had been created:
Screenshot 2025-06-16 at 16 23 11

Verified that the API container has been provided with the storage container's connection string and container name as expected:
Screenshot 2025-06-16 at 16 25 47


gareth-allan-bjss:
This isn't currently creating a container for us to store data to, so marking this as a draft until I can get it to.

.env.template
```diff
@@ -3,3 +3,5 @@ DB_USER=sa
 # DB_PASSWORD needs to be at least 8 characters including uppercase letters,
 # lowercase letters, base-10 digits and/or non-alphanumeric symbols.
 DB_PASSWORD=
+
+UPLOAD_STORAGE_CONTAINER_NAME=fingertips-upload-container
```
adawari-thomas-bjss, line 7: should `STORAGE_CONNECTION_STRING` be included here without a value? Just so people are aware of it

williamwillis11 (approved):

.env.template
gareth-allan-bjss: Hmm, good point. I'm not sure my solution would actually allow the user to override the connection string. Let me take a look and see if I can improve it.

adawari-thomas-bjss: great work!

adawari-thomas-bjss (approved):

PR #674, opened 2025-06-17
feature/DHSCFT-900_POST_upload

WolfieKnee:
# Description

Jira ticket: [DHSCFT-900](https://bjss-enterprise.atlassian.net/browse/DHSCFT-900)

## Changes

- Etc.

## Validation


api/DHSC.FingertipsNext.Modules.DataManagement/Controllers/V1/DataManagementController.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace DHSC.FingertipsNext.Modules.DataManagement.Controllers.V1;
+
+[ApiController]
+[Route("indicators/{indicatorId:int}/data")]
+[ProducesResponseType(StatusCodes.Status200OK)]
+[ProducesResponseType(typeof(JSType.Error), StatusCodes.Status400BadRequest)]
+public class DataManagementController() : ControllerBase
+{
+    [HttpPost]
+    public IActionResult AcceptFile([FromForm] IFormFile? file)
```
Mr-Ev-H, line 14: If this was not nullable, would you be able to remove the null check further down?

api/definition/indicators/paths.yaml
```diff
@@ -80,14 +81,26 @@ indicators_by_indicator_id_data:
...
+          schema:
+            type: object
+            properties:
+              file: 
+                type: string
+                format: binary
     responses:
-      "501":
-        description: Not Implemented
+      "200":
```
gareth-allan-bjss, line 95: Should we use `202` instead? I think it's well suited to this purpose.
```suggestion
      "202":
```

api/definition/indicators/paths.yaml
```diff
@@ -80,14 +81,26 @@ indicators_by_indicator_id_data:
       - indicators
     summary: add a batch of new data for an indicator
     description: Not Yet Implemented - Creates new data for the indicator. The data is always created in an unpublished and unapproved state.
+    requestBody:
+      required: true
+      content:
+        multipart/form-data:
+          schema:
+            type: object
+            properties:
```
gareth-allan-bjss, line 90: I think we also need to accept the data publication date as a property here, as [described in the OpenAPI docs](https://swagger.io/docs/specification/v3_0/describing-request-body/file-upload/).

api/definition/indicators/paths.yaml
gareth-allan-bjss, line 1: I think we should be regenerating the API clients as part of this change, so that we can prove the spec produces workable client code.

api/DHSC.FingertipsNext.Modules.DataManagement.UnitTests/Controllers/V1/DataManagementControllerTests.cs
```diff
@@ -0,0 +1,68 @@
...
+{
+    private readonly IDataManagementRepository _dataManagementRepository;
+    private readonly IDataManagementService _dataManagementService;
+    private readonly DataManagementController _controller;
+
+    public DataManagementControllerTests()
+    {
+        _dataManagementRepository = new DataManagementRepository();
+        _dataManagementService = new DataManagementService(_dataManagementRepository);
+        // _controller = new DataManagementController(_dataManagementService);
```
gareth-allan-bjss, line 20: I think it's worth either keeping the `DataManagementService` as a dependency of your new controller, or removing these bits that set up a repository/service from here until we need them again.

api/DHSC.FingertipsNext.Modules.DataManagement/Controllers/V1/DataManagementController.cs
gareth-allan-bjss, line 14: `AcceptFile` is a pretty generic name. I'd suggest something more specific to what it is going to do, like `UploadHealthData`.

api/DHSC.FingertipsNext.Modules.DataManagement/Controllers/V1/DataManagementController.cs
```diff
@@ -0,0 +1,20 @@
...
+    {
+        if (file == null || file.Length == 0) return BadRequest();
+
+        return Ok($"File {file.FileName} has been accepted.");
```
gareth-allan-bjss, line 18: I appreciate this is a simple version, but we should start as we mean to go on. See this warning from the [ASP.NET Core docs on uploading files](https://learn.microsoft.com/en-us/aspnet/core/mvc/models/file-uploads?view=aspnetcore-9.0#file-upload-scenarios) about not trusting the filename the user has submitted:
Screenshot 2025-06-17 at 17 33 46

We should look to [HTML encode](https://learn.microsoft.com/en-us/dotnet/api/system.web.httputility.htmlencode?view=net-9.0) this value before returning it in the response.

api/definition/indicators/paths.yaml
WolfieKnee: I did think we'd swap to that later, but happy to make the change now.

api/DHSC.FingertipsNext.Modules.DataManagement/Controllers/V1/DataManagementController.cs
WolfieKnee: ✅ done

api/definition/indicators/paths.yaml
gareth-allan-bjss: OK, fair enough.

gareth-allan-bjss (approved):

api/definition/indicators/paths.yaml
WolfieKnee: ✅ - regenerated and updated handlers.ts

api/DHSC.FingertipsNext.Modules.DataManagement.UnitTests/Controllers/V1/DataManagementControllerTests.cs
```diff
@@ -55,10 +56,31 @@ public void EmptyFileReturns400()
...
+        var bytes = System.Text.Encoding.UTF8.GetBytes(csv);
+        var stream = new MemoryStream(bytes);
+        var formFile = new FormFile(stream, 0, bytes.Length, "file", stubFileName);
+        var expectedEncoded = HttpUtility.HtmlEncode(stubFileName);
+
+        // Act
+        var response = _controller.UploadHealthData(formFile, stubIndicatorId) as OkObjectResult;
+
+        // Assert
+        response?.StatusCode.ShouldBe(200);
```
gareth-allan-bjss, line 83: Shouldn't the response's status code be 202 now? As per the Swagger spec.

PR #675, opened 2025-06-17
DHSCFT-1001-Remove searchState prop drilling for chart

alibombaye:
# Description

Jira ticket: [DHSCFT-1001](https://bjss-enterprise.atlassian.net/browse/DHSCFT-1001)

Removes the prop drilling for chart two indicator views.

## Validation

All tests passing. Tested manually.


stuc-bjss:
`benchmarkToUse={benchmarkToUse}` also looks like something that a component that is dependent on the `useSearchStateParams` can easily work out from that without needing it to be passed around?

PR #707, opened 2025-06-30
Sex Segmentation - Basic version

HowardMayBjss:
# Description

Jira ticket: [DHSCFT-1060](https://bjss-enterprise.atlassian.net/browse/DHSCFT-1060)

This introduces Segmentation of Indicators by Sex Dimension. It works for Basic (non-quartile) indicators and provides correct benchmarking for the Aggregate and new Segmented data against National and Sub-National areas.

It also takes care to ensure that benchmarking for Deprivation inequalities continues to work and that these are benchmarked against the fully aggregate data.

## Changes

- Backwards compatible with existing code
- Fetches data for segments from DB
- Includes IndicatorSegments in response
- Supports Benchmarking
- Deprecates old fields in Schema

## Validation

Existing tests pass.
New HTTP tests for Indictator Segments
New unit tests for Indicator Segments


api/definition/fingertipsAPI.yaml
```diff
@@ -689,6 +689,22 @@ components:
...
+        - isAggregate
+      properties:
+        value:
+          type: string
+          example: Female
+          description: Sex which the data are for.
+        isAggregate:
+          type: boolean
+          example: false
+          description: Indicates if the datapoint is an aggregated value for the sex dimension.
```
stuc-bjss, line 707: Is this actually specific to the Sex segment, is it not a segment dimension or segment meta data info.
Such that the IndicatorSegment then has props 
```
sex: SegmentInfo, 
age: SegmentInfo
frequency: SegmentInfo
```

api/definition/fingertipsAPI.yaml
```diff
@@ -879,6 +880,23 @@ components:
...
           $ref: '#/components/schemas/DeprivationData'
+    IndicatorSegment:
+      title: IndicatorSegment
+      type: object
+      description: Represents a segment of an Indicator health data point for a public health indicator .
+      required:
+        - sex
+      properties:
+        sex:
+          $ref: '#/components/schemas/SexData'
```
stuc-bjss, line 891: as mentioned previously could this just be 
```suggestion
          $ref: '#/components/schemas/SegmentInfo'
```

andyevans-bjss (approved):

api/DHSC.FingertipsNext.Modules.HealthData.Repository/HealthDataRepository.cs
```diff
@@ -96,6 +93,7 @@ public async Task<IEnumerable<HealthMeasureModel>> GetIndicatorDataAsync(int ind
             .Include(healthMeasure => healthMeasure.IndicatorDimension)
             .Include(healthMeasure => healthMeasure.DeprivationDimension)
             .Include(healthMeasure => healthMeasure.TrendDimension)
+            .OrderBy(healthMeasure => healthMeasure.FromDateDimension.Date)
```
Mr-Ev-H, line 96: I dont think the order of the data matters when coming from the database does it? It feels like a presentation (service) rather than a repository.

api/DHSC.FingertipsNext.Modules.HealthData.Schemas/HealthDataForArea.cs
```diff
@@ -20,6 +20,12 @@ public class HealthDataForArea
     [JsonPropertyName("areaName")]
     public string AreaName { get; init; } = string.Empty;
 
+    /// <summary>
+    ///     Segments of the indicator
+    /// </summary>
+    [JsonPropertyName("indicatorSegments")]
+    public IEnumerable<IndicatorSegment> IndicatorSegments { get; set; } = [];
```
Mr-Ev-H, line 27:
```suggestion
    public IEnumerable<IndicatorSegment> IndicatorSegments { get; init; } = [];
```

api/DHSC.FingertipsNext.Modules.HealthData.Schemas/IndicatorSegment.cs
```diff
@@ -0,0 +1,28 @@
...
+    ///     Sex which the data are for.
+    /// </summary>
+    [JsonPropertyName("sex")]
+    public required Sex Sex { get; init; }
+
+    /// <summary>
+    ///     Is the segment the aggregate for the indicator
+    /// </summary>
+    [JsonPropertyName("isAggregate")]
+    public bool IsAggregate { get; set; }
```
Mr-Ev-H, line 21:
```suggestion
    public bool IsAggregate { get; init; }
```

api/DHSC.FingertipsNext.Modules.HealthData.UnitTests/Repository/HealthDataRepositoryTests.cs
```diff
@@ -400,38 +401,40 @@ public async Task GetIndicatorDataAsyncShouldFilterResultsByAllFiltersWhenProvid
     }
 
     [Fact]
-    public async Task GetIndicatorDataAsyncShouldOnlyIncludeResultsWithoutASexDimensionValue()
+    public async Task GetIndicatorDataAsyncShouldIncludeResultsWithSexDimensionValue()
```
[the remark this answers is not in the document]
HowardMayBjss: Indictor data in the database has entries with and without Sex Values of "Male", "Female" and "Persons". This was only passed back if the Indicator-data request called out for sex inequalities data. Now it is always passed back, where present for the indicator / request. So request to the repository has now been updated to return data from the DB for all Sex values. If the DB only has data for one sex value then only that would be returned.

api/DHSC.FingertipsNext.Modules.HealthData.Repository/HealthDataRepository.cs
HowardMayBjss: As I see it there is a trade off here. Sorting data in the database is more efficient while the frontend (service) is, as you say, the real arbiter of how it should be displayed. So what I have here is the DB doing some simple ordering on date and leaving the rest to the frontend. Once the ordering on data is done then hopefully the rest is a bit easier.

Mr-Ev-H: Not every consumer of the repository is going to care about the order of the results though. You're ordering the results here, but you dont know whether consumers of GetIndicatorDataAsync are always going to require the results to be ordered. 

I dont think we should worry about whether it's more efficient to do the ordering in memory vs in database, we should worry about reducing the complexity of the code needed to get the data out in the first place.

api/DHSC.FingertipsNext.Modules.HealthData.UnitTests/Services/IndicatorServiceTests.cs
```diff
@@ -93,9 +95,11 @@ public async Task GetIndicatorDataShouldReturnExpectedResult()
             [],
             []
         );
-        result.Content.AreaHealthData.ShouldNotBeEmpty();
-        result.Content.AreaHealthData.Count().ShouldBe(1);
-        result.Content.AreaHealthData.ElementAt(0).ShouldBeEquivalentTo(expected);
+        var areaHealthData = result.Content.AreaHealthData.ToList();
+        areaHealthData[0].IndicatorSegments = new List<IndicatorSegment>();
+        areaHealthData.ShouldNotBeEmpty();
```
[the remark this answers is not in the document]
Mr-Ev-H: I think you'd get an exception further up when you call .ToList() on AreaHealthData or addressing the object at index zero if that was the case.
You can also use .ShouldHaveSingleItem()

api/DHSC.FingertipsNext.Modules.HealthData.Repository/HealthDataRepository.cs
stuc-bjss: a consistent and predictable order of data from the API is a good though!?

api/DHSC.FingertipsNext.Modules.HealthData.UnitTests/Services/IndicatorServiceTests.cs
```diff
@@ -93,9 +95,11 @@ public async Task GetIndicatorDataShouldReturnExpectedResult()
             [],
             []
         );
-        result.Content.AreaHealthData.ShouldNotBeEmpty();
-        result.Content.AreaHealthData.Count().ShouldBe(1);
-        result.Content.AreaHealthData.ElementAt(0).ShouldBeEquivalentTo(expected);
+        var areaHealthData = result.Content.AreaHealthData.ToList();
+        areaHealthData[0].IndicatorSegments = new List<IndicatorSegment>();
```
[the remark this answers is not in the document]
HowardMayBjss: This was done to make the existing test continue to validate the existing healthDataPoints without failing because of the addition of the IndicatorSegments. It's not the best long term strategy so have refactored to change this.

api/DHSC.FingertipsNext.Modules.HealthData.UnitTests/Services/IndicatorServiceTests.cs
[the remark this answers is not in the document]
HowardMayBjss: I think ShouldNotBeEmpty also checks whether areaHealthData is null and prevents the next line from causing an exception.

api/definition/fingertipsAPI.yaml
HowardMayBjss: I think we are misscommunicating. SexData and AgeData are existing types in the API and refactoring them is not in scope of this ticket. And they have distict enum values which wouldn't combine into the same type.

PR #708, opened 2025-06-30
Feature/dhscft 1094 unify project references

Mr-Ev-H:
# Description

Jira ticket: [DHSCFT-1094](https://bjss-enterprise.atlassian.net/browse/DHSCFT-1094)

Unifying all nuget package versions and performing upgrade to latest versions.

## Changes

- Moving all package versions into single root Packages.props file
- Fixing crash on startup for openapi to allow viewing of spec in windows

## Validation

Unit tests and integration tests continue to pass.


api/DHSC.FingertipsNext.Api/Program.cs
```diff
@@ -54,6 +57,14 @@ public static void Main(string[] args)
 
         builder.Services.AddControllers().AddControllersAsServices();
 
+
+        // This is included to resolve a runtime crash on windows caused by https://github.com/dotnet/aspnetcore/issues/58805
```
williamwillis11, line 61: worth us logging a tech debt ticket to remove this if its likely to be fixed in a future .Net version or something?

Mr-Ev-H: Yeah good shout, worth revisiting in the future

HowardMayBjss (approved):

PR #718, opened 2025-07-04
DHSCFT-1024: Delete unpublished data from healthdata module

adawari-thomas-bjss:
# Description

Jira ticket: [DHSCFT-1024](https://bjss-enterprise.atlassian.net/browse/DHSCFT-1024)

## Changes

- Added new endpoint to delete unpublished batch data from health measures table
- Added new db context to the health data module to handle delete operation
- Added integration tests for health data module to verify changes

## Validation
 
- Integration tests passing


api/DHSC.FingertipsNext.Modules.HealthData.Repository/HealthDataRepository.cs
```diff
@@ -125,6 +125,28 @@ public async Task<IEnumerable<AreaDimensionModel>> GetAreasAsync(string[] areaCo
...
+        var batchToDelete = batchHealthDataDbContext.HealthMeasure
+            .Where(healthMeasure => healthMeasure.IndicatorDimension.IndicatorId == indicatorId)
+            .Where(healthMeasure => healthMeasure.BatchId == batchId);
+
+        var batchToDeleteList = await batchToDelete.ToListAsync();
+
+        if (batchToDeleteList.Count == 0)
+        {
+            return false;
+        }
```
gareth-allan-bjss, line 139: [`ExecuteDeleteAsync`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.relationalqueryableextensions.executedeleteasync?view=efcore-8.0) should return:
> The total number of rows deleted in the database.

So no need to count the rows first, we can just return true if > 0 rows were deleted, else false.

api/DHSC.FingertipsNext.Modules.HealthData.Repository/HealthDataRepository.cs
```diff
@@ -125,6 +125,28 @@ public async Task<IEnumerable<AreaDimensionModel>> GetAreasAsync(string[] areaCo
...
+
+        if (batchToDeleteList.Any(healthMeasure => healthMeasure.PublishedAt <= DateTime.UtcNow))
+        {
+            throw new InvalidOperationException("Error attempting to delete published batch.");
+        }
```
gareth-allan-bjss, line 144: It would be good to see the SQL being generated when deleting (it should be logged to the console by the API when you make a delete request).

I'm a little concerned that we may be selecting all rows to be deleted, then deleting them, which could be inefficient.

api/DHSC.FingertipsNext.Modules.HealthData.Repository/HealthDataRepository.cs
```diff
@@ -123,6 +125,24 @@ public async Task<IEnumerable<AreaDimensionModel>> GetAreasAsync(string[] areaCo
...
+    public async Task<bool> DeleteAllHealthMeasureByBatchIdAsync(int indicatorId, string batchId)
+    {
+        var batchToDelete = _batchDbContext.HealthMeasure
+            .Where(healthMeasure => healthMeasure.IndicatorDimension.IndicatorId == indicatorId)
+            .Where(healthMeasure => healthMeasure.BatchId == batchId);
+
+        var batchToDeleteFirstEntry = await batchToDelete.FirstOrDefaultAsync();
+        if (batchToDeleteFirstEntry == null) return false;
+
+        if (batchToDeleteFirstEntry.PublishedAt <= DateTime.UtcNow)
```
Mr-Ev-H, line 137: @andyevans-bjss has introduced a helper function to return this as a db query from the db context which it would be good to make use of here 

https://github.com/dhsc-govuk/FingertipsNext/blob/33afed6d8285b4ddb6fb4cfd3248546e2b1becc0/api/DHSC.FingertipsNext.Modules.HealthData.Repository/HealthDataDbContext.cs#L29

adawari-thomas-bjss: The helper function is using a different db context from the delete operation. Also, it seems the helper function is returning all the health measures while this just checks a single health measure to verify if it's published. I don't think the helper applies here.

api/DHSC.FingertipsNext.Api.IntegrationTests/setup.sql
ehsh1, line 1: Is this better placed under /HealthData?

api/DHSC.FingertipsNext.Modules.HealthData.UnitTests/Repository/HealthDataRepositoryDeleteTests.cs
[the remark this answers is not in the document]
gareth-allan-bjss: The repository tests were getting pretty big anyway!

api/DHSC.FingertipsNext.Modules.HealthData.UnitTests/Repository/HealthDataRepositoryDeleteTests.cs
```diff
@@ -0,0 +1,149 @@
...
+        if (disposing)
+        {
+            _dbContext.Database.EnsureDeleted();
+            _dbContext.Dispose();
+            _connection.Close();
+            _connection.Dispose();
+        }
+    }
+
+    #region DeleteUnpublishedData
```
Mr-Ev-H, line 48:
```suggestion

```

PR #722, opened 2025-07-04
Add an E2E test for data upload

gareth-allan:
# Description

Jira ticket: [DHSCFT-918](https://bjss-enterprise.atlassian.net/browse/DHSCFT-918)
Subtask: [DHSCFT-1123](https://bjss-enterprise.atlassian.net/browse/DHSCFT-1123)

This PR adds a Playwright test that navigates to the upload page, fills in the form and uploads a file and then verifies a success message is shown. **It is currently set to do this both in CI and CD (i.e. against the deployed site)**. I think that running against the deployed site has value, as it'll prove we've wired together the components needed for an upload, and the side-effect of uploading files/adding unpublished data is (currently) negligable.

## Changes

- Added an `UploadPage` page object to represent the indicator data upload page
- Added a new test file that navigates to the upload page, fills in the upload form, clicks the submit button and then checks the expected success message is displayed

## Validation
Verified that the new test is run as part of the pipeline:
Screenshot 2025-07-04 at 17 14 57

I ran the test in UI mode locally and verified that it resulted in the expected message:
Screenshot 2025-07-04 at 17 12 51

I also checked the storage container running in Docker and verified that a file had been uploaded as expected:
Screenshot 2025-07-04 at 17 13 05

Finally, I ran the tests against the deployed site using the `npm run test-e2e-cd` command and verified that all tests (except the two know failures we currently have) passed:
Screenshot 2025-07-04 at 17 28 48

I also confirmed that the new test passed against the deployed site:
Screenshot 2025-07-04 at 17 29 05

### DIFF
--- a/frontend/fingertips-frontend/playwright/page-objects/pages/uploadPage.ts
+++ b/frontend/fingertips-frontend/playwright/page-objects/pages/uploadPage.ts
@@ -1,4 +1,3 @@
-import path from 'path';
 import BasePage from '../basePage';
 import { expect } from '../pageFactory';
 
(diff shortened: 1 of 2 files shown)


frontend/fingertips-frontend/playwright/page-objects/pages/uploadPage.ts
```diff
@@ -0,0 +1,80 @@
...
+    __dirname,
+    '..',
+    '..',
+    'resources',
+    this.csvFileName
+  );
+
+  private readonly indicatorId = '41101';
+
+  async fillInUploadForm() {
```
williamwillis11, line 28: can we have the date in the test spec passed to this function?

makes it a lot easier when debugging a test failure to see whats gone on/wrong

frontend/fingertips-frontend/playwright/page-objects/pages/uploadPage.ts
```diff
@@ -0,0 +1,80 @@
...
+    await this.navigateTo('/upload');
+  }
+
+  async checkOnUploadPage() {
+    await expect(
+      this.page.getByRole('heading', { name: this.pageHeadingText }).first()
+    ).toBeVisible();
+  }
+
+  async checkApiResponsePanelContainsASuccessMessage() {
```
williamwillis11, line 69: can this be

checkApiResponsePanelContainsMessage

and it take a string param of the message its going to check for?

that way we can have the message were checking for, success in this instance - in the test spec passed through, and use this same method for negative test cases right?

williamwillis11: maybe pass two things through for lines 73 and 74?

statusCode
message

PR #723, opened 2025-07-04
Sex segmentation Quintile

HowardMayBjss:
# Description

Jira ticket: [DHSCFT-1062](https://bjss-enterprise.atlassian.net/jira/software/projects/DHSCFT/boards/823?selectedIssue=DHSCFT-1062)

Supports Sex Segmentation for Quintiles. Data for each Sex Segmentation is now returned for Quintile data. At present there doesn't seem to be any data segmented based on Sex and so the principal change here is to present the data back in the new IndicatorSegment structure. When we have data which is quintiles but different reportingPeriods then some of these structures may be used a bit more.

## Changes

- Updates to Quintile to support Sex Segments
- Presentation of Quintile data in Segments
- Updated tests to validate segmented structure
- Etc.

## Validation

HTTP tests updated and added to validate Segmented data


api/DHSC.FingertipsNext.Modules.HealthData/HttpClientScripts/test-indicators-benchmark-quintiles-response.json
```diff
@@ -19,9 +477,9 @@
             "isAggregate": true
           },
           "deprivation": {
-            "sequence": 10,
```
HowardMayBjss, line 22: There appears to be a change in ordering here which should be looked at and resolved

andyevans-bjss (approved):

PR #742, opened 2025-07-11
DHSCFT-1103: Exclude unpublished data. Update CSV validation logic

ehsh1:
# Description

Jira ticket: https://bjss-enterprise.atlassian.net/browse/DHSCFT-1103

## Changes

- Exclude unpublished data from trend calculation
- Update CSV validation logic
- Update HTTP tests

## Validation
Unit and HTTP tests pass

Ran e2e tests locally against docker and got passing after updating the config as expected. Also ran the e2e tests against CD and see that they fail as expected as the trend cal app will only run on deployment to update the value - so we should see it go green after merging in.


api/DHSC.FingertipsNext.Modules.DataManagement.Service/Validation/UploadedCsvValidator.cs
```diff
@@ -77,6 +77,11 @@ private static bool ValidateHeaders(CsvReader csv, out List<string> unexpectedHe
         if (!headers.SequenceEqual(validHeaderList))
         {
             unexpectedHeaders = headers.Except(validHeaderList).ToList();
+
+            // If there aren't any unexpected headers, then the header order is incorrect
```
williamwillis11, line 81: wait does this make sense?

if unexpectedHeaders.Count === 0 then we are good as we have no unexpected headers, only expected headers right?

ehsh1: Updated comment to make it clearer

Mr-Ev-H (approved):

gareth-allan-bjss (approved):

PR #746, opened 2025-07-15
DHSCFT-1063: Sex Segmentation - Quartiles end point

HowardMayBjss:
# Description

Jira ticket: [DHSCFT-1063](https://bjss-enterprise.atlassian.net/browse/DHSCFT-1063)

Adds Sex Segmentation to Quartiles end point

## Changes

- The Quartiles SQL updated to include data segmented by Sex inequality.
- SQL returned data now includes Sex dimension
- C# Service marshalls additional data to API as JSON
- Http tests updated with new format and additional test added
- QuartileDataModel made keyless because C# doesn't support a compound key (indicator_id, sexName) a component of which can be null (sexName)
- UI filters out non-aggregate responses for now for backwards compatibility

## Validation

HTTP tests updated and extended
Unit tests updated


andyevans-bjss (approved):

api/DHSC.FingertipsNext.Modules.HealthData.Repository/Models/QuartileDataModel.cs
```diff
@@ -1,13 +1,16 @@
...
 
+[Keyless]
 public class QuartileDataModel
 {
-    [Key]
     public required int IndicatorId { get; set; }
     public required string? Polarity { get; set; }
     public required short? Year { get; set; }
+    public required string? SexName { get; set; }
+    public required bool? IsSexAggregatedOrSingle { get; set; }
```
Nayira-BJSS, line 13: Given this is a boolean, wouldn't it be better to have this as just "isSexAggregated" ? 
Similarly for isAgeAggreagatedOrSingle and isDeprivationAggregatedOrSingle, although I am suspecting this is existing code so will leave it to your discretion to update them all if you agree

HowardMayBjss: I don't disagree but this is mapping through to the database tables and how the fields are named in there. While we could improve the name throughout that would be quite disruptive.

Nayira-BJSS (approved):

PR #738, opened 2025-07-10
Feature/dhscft 942 integrate auth

Joe-Hendley:
draft pr please ignore

PR #758, opened 2025-07-17
DHSCFT-1039_View_unpublished_data_LineCharts_AreasTable_no_RBAC_2

WolfieKnee:
# Description

Jira ticket: [DHSCFT-1039](https://bjss-enterprise.atlassian.net/browse/DHSCFT-1039)

NOTE: dependant on [PR738](https://github.com/dhsc-govuk/FingertipsNext/pull/738)

View unpublished data (without RBAC) for Line chart & Compare areas table. 
* Server session is acquired from auth.js and used to seed ReactQuery and initialise `SessionProvider`. Session from `useSession()` in `useApiGetHealthDataForAnIndicator` hook is used to determine which end point is used to get the data. `useEffect` is added to render of SignIn in order to ensure `SessionProvider` is updated on sgnOut.

## Changes

- add necessary env variables for testing to gh workflows. 
- `app/chart/page`: use session to determine endpoint to seed ReactQuery
- `layout`: initialise `SessionProvider`
- `AuthButton` use render of `SignIn` to refresh `SessionProvider`
- `components/chart`: 
  - add `HealthDataForAnIndicatorIncludingUnpublished` to `queryKey` Endpoints
  - `useApiGetHealthDataForAnIndicator`: use session to determine which endpoint to use to get data
- `mocks\utils`: 
  - add `mockIndicatorsApi`
  - add optional `QueryClient` parameter to `testRenderWrapper`

## Validation

Indicator 41101 with unpublished data not shown when not signed in
image of indicator without unpublished data when not signed in

Indicator 41101 with unpublished data in line chart and compare areas table when signed in
image of indicator with unpublished data when not signed in


.github/workflows/fingertips-workflow.yml
```diff
@@ -334,6 +334,7 @@ jobs:
     environment: development
     runs-on: dev
     if: >
+      false &&
```
pete-bjss, line 337: Do we want to remove this now, or will that be another PR?

PR #761, opened 2025-07-17
Bugfix/dhscft tech remove token from session

Joe-Hendley:
Previously we were attaching the access token to the user session, but this leaks it to the client when it doesn't need to be through the session callback

this keeps it inside the server and reads the req cookie directly to access it

added some helper functions

manually tested
unit tests WIP - we won't be able to unit test the root getJWT function unless I crack header/cookies context in unit tests

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -10,18 +10,23 @@ DB_PASSWORD=
 #UPLOAD_STORAGE_ACCOUNT_CONNECTION_STRING=""
 UPLOAD_STORAGE_CONTAINER_NAME=fingertips-upload-container
 
+### Authentication ###
+
+# bare minimum for the compose to run is: 
+# API authentication env vars populated - placeholder values below are fine
+# and AUTH_SECRET to be populated
+
+# api authentication integration - values provided by OIDC provider
+AZUREAD__INSTANCE="https://foobar.example.com/"
+AZUREAD__TENANTID="tenant-id"
+AZUREAD__CLIENTID="client-id"
+
 # local authentication secret for the frontend - mandatory
 # can be populated locally with npm exec auth secret
 AUTH_SECRET=""
 
-
 # frontend authentication integration - values provided by OIDC provider
-AUTH_CLIENT_ID=""
-AUTH_CLIENT_SECRET=""
-AUTH_ISSUER=""
-AUTH_WELLKNOWN=""
-
-# api authentication integration - values provided by OIDC provider
-AZUREAD__INSTANCE=""
-AZUREAD__TENANTID=""
-AZUREAD__CLIENTID=""
+AUTH_CLIENT_ID="client-id"
+AUTH_CLIENT_SECRET="client-secret"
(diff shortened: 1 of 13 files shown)


frontend/fingertips-frontend/lib/auth/accessToken.test.ts
```diff
@@ -0,0 +1,55 @@
...
+  it('should return the access token on a JWT if a session exists and a JWT with a valid access token is found', async () => {
+    mockAuth.mockResolvedValue(validSession);
+    mockGetJWT.mockResolvedValue(validJWT);
+
+    const result = await getAccessToken();
+
+    expect(result).toBe(validJWT.accessToken);
+  });
+
+  it("should return no access token a session doesn't exist", async () => {
```
williamwillis11, line 29: should return no access token if a session doesn't exist

frontend/fingertips-frontend/lib/auth/accessToken.test.ts
```diff
@@ -0,0 +1,55 @@
...
+import { Mock } from 'vitest';
+
+vi.mock('@/lib/auth/getJWT', () => {
+  return { getJWT: vi.fn() };
+});
+
+const mockGetJWT = getJWT as Mock;
+
+const validJWT: JWT = { accessToken: 'hunter2' };
+const validSession: Session = { expires: 'some time' };
```
williamwillis11, line 17: i think i know your answer but wanted to double check, is now/here not the time to add an expired session test?

Joe-Hendley: the auth library manages that for us

Joe-Hendley: I can take a closer look at that next week but again, if the session has expired then session just returns null

williamwillis11 (approved):

frontend/fingertips-frontend/mock/utils/mockAuth.ts
```diff
@@ -0,0 +1,11 @@
+export const mockAuth = vi.fn();
```
stuc-bjss, line 1: nice

stuc-bjss (approved):

frontend/fingertips-frontend/lib/auth/accessToken.test.ts
```diff
@@ -0,0 +1,55 @@
...
+  });
+
+  it('should return no access token if access token not on JWT', async () => {
+    mockAuth.mockResolvedValue(validSession);
+    mockGetJWT.mockResolvedValue(null);
+
+    const result = await getAccessToken();
+
+    expect(result).toBeUndefined();
+  });
```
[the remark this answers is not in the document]
Joe-Hendley: good spot

gareth-allan-bjss (approved):

WolfieKnee (approved):
Validated behaviour as expected when viewing unpublished data against mock password.

PR #772, opened 2025-07-21
DHSCFT-928: Add Authorisation to List Batches Endpoint

gareth-allan-bjss:
# Description

Jira ticket: [DHSCFT-928](https://bjss-enterprise.atlassian.net/browse/DHSCFT-928)

This PR applies authentication to the "list batches" endpoint, so that it only shows a user batches which belong to indicators they have permission to modify. Administrators are shown all batches and user's without any permissions receive an HTTP 403 (forbidden) response.

## Changes

- Updated the `ListBatches` function in `DataManagementBatchController` to use the user's roles to obtain a list of indicators they have permissions to modify (or whether they have the admin role)
  - This code is based on the approach in the `UserHasIndicatorManagementRolesRequirementHandler` class
- Updated the API spec to specify that authentication is required for the list batches endpoint and regenerated the API client
- Updated the upload page in the frontend to pass an access token with "list batches" requests
- Refactored the `DataManagementIntegrationTests` so that it is now a base class providing common functionality and we now have `FileUploadIntegrationTests` for the existing "file upload" tests and `ListBatchesIntegrationTests` for the existing "list batches" test, plus new ones to verify auth is applied as expected

## Validation

Uploaded files for indicators `41101`, `12345`, `54321`.

Logged in as an administrator and used the upload page to list batches:
Screenshot 2025-07-21 at 15 06 39

Logged in as a user with permissions for `41101` only (as the others aren't real indicators) and used the upload page to list batches:
Screenshot 2025-07-21 at 15 07 33

Logged in as a user with a valid account but no indicators assigned and navigated to the upload page:
Screenshot 2025-07-21 at 15 08 40

Navigated to the upload page without logging in:
Screenshot 2025-07-21 at 15 09 01


api/DHSC.FingertipsNext.Modules.DataManagement.UnitTests/Controllers/V1/DataManagementBatchControllerTests.cs
gareth-allan-bjss, line 1: I've removed this test because the only logic in `DataManagementBatchController` is for the authorisation of the list batches endpoint, which I felt was already (better) covered by the integration tests. Mocking out all of the auth requirements here didn't seem particularly valuable.

I'm happy to discuss this approach though!

api/DHSC.FingertipsNext.Modules.DataManagement.UnitTests/Controllers/V1/DataManagementBatchControllerTests.cs
```diff
@@ -20,29 +20,13 @@ public class DataManagementBatchControllerTests
 
     public DataManagementBatchControllerTests()
     {
-        _dataManagementService = Substitute.For<IDataManagementService>();
-        _controller = new DataManagementBatchController(_dataManagementService);
-    }
-
-    [Fact]
-    public async Task ListBatchesReturnsTheExpectedResult()
```
gareth-allan-bjss, line 28: I've removed this test because the only logic in `DataManagementBatchController` is for the authorisation of the list batches endpoint, which I felt was already (better) covered by the integration tests. Mocking out all of the auth requirements here didn't seem particularly valuable.

I'm happy to discuss this approach though!

api/DHSC.FingertipsNext.Api.IntegrationTests/DataManagement/DeleteBatchIntegrationTests.cs
```diff
@@ -0,0 +1,113 @@
...
+using DHSC.FingertipsNext.Modules.DataManagement.Repository.Models;
+using DHSC.FingertipsNext.Modules.DataManagement.Schemas;
+using DHSC.FingertipsNext.Modules.HealthData.Repository;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.Extensions.DependencyInjection;
+using Shouldly;
+
+namespace DHSC.FingertipsNext.Api.IntegrationTests.DataManagement;
+
+public sealed class DeleteBatchIntegrationTests : DataManagementIntegrationTests
```
williamwillis11, line 14: think we need a test for a request to delete a batch that you dont have permissions to delete 

so i know we enforce and test for that via the UI in the list batches code and tests but since this delete batch api can be called without the UI - i think we need to assert that if a user with role for indicator batch 12345 tries to delete 54321 indicator batch (that isnt already deleted or published) then they get a forbidden or something 

sorry if this test exists and i missed it

api/DHSC.FingertipsNext.Api.IntegrationTests/DataManagement/FileUploadIntegrationTests.cs
```diff
@@ -0,0 +1,278 @@
...
+    {
+        // Arrange
+        var apiClient = Factory.CreateClient();
+
+        using var content = new MultipartFormDataContent();
+
+        // Act
+        var response = await apiClient.PostAsync(new Uri($"/indicators/{IndicatorId}/data", UriKind.Relative), content);
+
+        //Assert
```
williamwillis11, line 102: // Assert

there are a couple other instances of this critical code issue!!! ;)

api/DHSC.FingertipsNext.Api.IntegrationTests/DataManagement/DeleteBatchIntegrationTests.cs
williamwillis11: could an expired test similar to

'ExpiredRequestToDataManagementEndpointShouldBeRejected'

be added at this point though?

api/DHSC.FingertipsNext.Modules.DataManagement/Controllers/V1/DataManagementBatchController.cs
```diff
@@ -1,30 +1,75 @@
...
     [ProducesResponseType(typeof(Batch[]), StatusCodes.Status200OK)]
+    [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+    [ProducesResponseType(StatusCodes.Status403Forbidden)]
+    [ProducesResponseType(StatusCodes.Status500InternalServerError)]
     public async Task<IActionResult> ListBatches()
     {
-        // The indicator IDs a user has access to will be determined based on their authorisation roles in a subsequent change.
-        int[] indicatorIds = [];
+        var userRoles = GetRoles(User);
+        var userIndicatorPermissions = await GetIndicatorIdsFromRoles(userRoles);
```
Mr-Ev-H, line 65: I think calling this simply `roleIds` is clearer and would aid with readability.

api/DHSC.FingertipsNext.Api.IntegrationTests/DataManagement/ListBatchesIntegrationTests.cs
```diff
@@ -0,0 +1,169 @@
...
+        // Assert
+        // Cannot assert that this is the entire list of batches, as other tests may be inserting data.
+        response.ShouldNotBeNull();
+        response.ShouldContain(BatchFor41101);
+        response.ShouldContain(BatchFor383);
+        response.ShouldContain(BatchFor22401);
+    }
+
+    [Fact]
+    public async Task ListBatchesEndpointShouldListAllBatchesForAdministratorsWithAdditionalRoles()
```
williamwillis11, line 73: i didnt know this was a valid scenario, or are we just covering it incase of a miss configuration of a user in EntraID?

api/DHSC.FingertipsNext.Api.IntegrationTests/DataManagement/ListBatchesIntegrationTests.cs
```diff
@@ -0,0 +1,169 @@
...
+        var expectedResult = new[] { BatchFor41101 };
+
+        var apiClient = Factory.CreateClient();
+        apiClient.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("Bearer",
+            Factory.GenerateTestToken([Indicator41101GroupRoleId]));
+
+        // Act
+        var response = await apiClient.GetFromJsonAsync<Batch[]>(new Uri("/batches", UriKind.Relative));
+
+        // Assert
```
williamwillis11, line 104: do we need something explicit here saying 

response.ShouldNotContain(BatchFor383);
response.ShouldNotContain(BatchFor22401);

api/DHSC.FingertipsNext.Api.IntegrationTests/DataManagement/ListBatchesIntegrationTests.cs
```diff
@@ -0,0 +1,169 @@
...
+        var expectedResult = new[] { BatchFor41101, BatchFor383 };
+
+        var apiClient = Factory.CreateClient();
+        apiClient.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("Bearer",
+            Factory.GenerateTestToken([Indicator41101GroupRoleId, Indicator383GroupRoleId]));
+
+        // Act
+        var response = await apiClient.GetFromJsonAsync<Batch[]>(new Uri("/batches", UriKind.Relative));
+
+        // Assert
```
williamwillis11, line 121: and maybe:

response.ShouldNotContain(BatchFor22401);

?

api/DHSC.FingertipsNext.Api.IntegrationTests/DataManagement/ListBatchesIntegrationTests.cs
```diff
@@ -0,0 +1,169 @@
...
+        response.ShouldBeEquivalentTo(expectedResult);
+    }
+
+    [Fact]
+    public async Task ListBatchesEndpointShouldReturnForbiddenWhenTheUserHasNoPermissions()
```
williamwillis11, line 126: ListBatchesEndpointShouldReturnForbiddenWhenUnauthorisedUserHasNoPermissions

would this be clearer ^^^?

api/DHSC.FingertipsNext.Api.IntegrationTests/DataManagement/FileUploadIntegrationTests.cs
```diff
@@ -0,0 +1,278 @@
...
+    [Theory]
+    [InlineData(AdminRoleGuid)]
+    [InlineData(Indicator41203GroupRoleId)]
+    [InlineData(Indicator41203GroupRoleId, Indicator383GroupRoleId)]
+    public async Task AuthorisedRequestToDataManagementEndpointShouldUploadAFile(params string[] userRoleIds)
+    {
+        // Arrange
+        var apiClient = Factory.CreateClient();
+
+        var publishedAt = DateTime.UtcNow.AddMonths(1);
```
Mr-Ev-H, line 56: Because we're providing a mocked time, this should be a fixed value rather than dynamic to reduce the risk of test instability.

api/DHSC.FingertipsNext.Api.IntegrationTests/DataManagement/FileUploadIntegrationTests.cs
```diff
@@ -0,0 +1,278 @@
...
+
+    [Theory]
+    [InlineData(AdminRoleGuid)]
+    [InlineData(Indicator41203GroupRoleId)]
+    public async Task UploadFailuresShouldReturn500Response(string userRoleId)
+    {
+        // Arrange
+        var apiClient = Factory.CreateClient();
+
+        var publishedAt = DateTime.UtcNow.AddMonths(1);
```
Mr-Ev-H, line 148: See above, should be a const rather than dynamic

api/DHSC.FingertipsNext.Api.IntegrationTests/DataManagement/FileUploadIntegrationTests.cs
```diff
@@ -0,0 +1,278 @@
...
+
+    [Theory]
+    [InlineData(AdminRoleGuid)]
+    [InlineData(Indicator41203GroupRoleId)]
+    public async Task UploadingInvalidFileShouldReturn400Response(string userRoleId)
+    {
+        // Arrange
+        var apiClient = Factory.CreateClient();
+
+        var publishedAt = DateTime.UtcNow.AddMonths(1);
```
Mr-Ev-H, line 181: See above.

api/DHSC.FingertipsNext.Api.IntegrationTests/DataManagement/FileUploadIntegrationTests.cs
```diff
@@ -0,0 +1,278 @@
...
+
+    [Theory]
+    [InlineData(AdminRoleGuid)]
+    [InlineData(Indicator41203GroupRoleId)]
+    public async Task UploadToBlobStorageShouldFailIfContainerDoesNotExist(string userRoleId)
+    {
+        // Arrange
+        var apiClient = Factory.WithWebHostBuilder(config => config.UseSetting("UPLOAD_STORAGE_CONTAINER_NAME", "invalid-container-name")).CreateClient();
+
+        var publishedAt = DateTime.UtcNow.AddMonths(1);
```
Mr-Ev-H, line 255: See above

api/DHSC.FingertipsNext.Api.IntegrationTests/DataManagement/FileUploadIntegrationTests.cs
```diff
@@ -0,0 +1,278 @@
...
+    {
+        var cleanupPath = Path.Combine(AppContext.BaseDirectory, SqlScriptDirectory, "file-upload-cleanup.sql");
+        RunSqlScript(cleanupPath, Connection);
+
+        _azureStorageBlobClient.DeleteBlob(_blobName);
+
+        base.Dispose(disposing);
+    }
+
+    [Theory]
```
williamwillis11, line 47: do we need tests along the lines of:

users that are still authorised, and at one point had the admin role, but now only have one specific indicator role - making sure they can only upload data for that indicator 

same question for list and delete apis as well

or would this be testing EntraId? 

this would need to be at a higher test level though right?

api/DHSC.FingertipsNext.Api.IntegrationTests/DataManagement/ListBatchesIntegrationTests.cs
gareth-allan-bjss: Yeah, I'm not sure how valid it would be, but I don't think there's anything that would stop an admin having other roles assigned to them and I can imagine a scenario where someone gets "promoted" to admin and their previous roles aren't removed.

I mainly added this test because I realised that my initial version of the logic would ignore the admin role in this scenario.

api/DHSC.FingertipsNext.Api.IntegrationTests/DataManagement/ListBatchesIntegrationTests.cs
gareth-allan-bjss: I think `ShouldBeEquivalentTo` would fail if extra elements are present (it fails if I set the expected result to an empty array). I realise though that I could use `response.ShouldBe(expectedResult)` here, which might be a little clearer?

api/DHSC.FingertipsNext.Modules.DataManagement/Controllers/V1/DataManagementBatchController.cs
Mr-Ev-H: ** indicatorIds sorry

api/DHSC.FingertipsNext.Api.IntegrationTests/DataManagement/FileUploadIntegrationTests.cs
gareth-allan-bjss: How would we even represent a user that had an admin role at one point, but no longer has it? As you say, this seems like it's testing Entra to make sure it reports the roles a user has correctly, rather than anything we've implemented. Unless @Mr-Ev-H has a different view?

Joe-Hendley (approved):

api/DHSC.FingertipsNext.Api.IntegrationTests/DataManagement/FileUploadIntegrationTests.cs
Mr-Ev-H: No, don't think this is a valid scenario. All we're capable of doing is testing the roles at a snapshot in time.,

PR #788, opened 2025-07-23
DHSCFT-930: Store User IDs When Creating and Deleting Batches

gareth-allan-bjss:
# Description

Jira ticket: [DHSCFT-930](https://bjss-enterprise.atlassian.net/browse/DHSCFT-930)

This PR updates the upload file and delete batch endpoints to record the IDs of the users performing the actions to the Batch table in the database.

## Changes

- Made the upload file endpoint capture the user's ID from the access token's `sub` claim and stored it in the batch's `UserId` field
- Made the delete batch endpoint capture the user's ID from the access token's `sub` claim and stored it in the batch's `DeletedUserId` field

## Validation

Added the following code to the upload page, to print the access token for a user (**Note:** this token is secret, so this code should never be checked in!):
```typescript
const accessToken = await getAccessToken();
console.log(`Access token: ${accessToken}`);
```

Using this, I logged in as the following users and  put the token into jwt.io to obtain their `sub` claims:

automationtestuser_admin@fingertipsdevapp.onmicrosoft.com (Admin user): `DsZbYClNPP7PaOfnzQKQZzC-BSF7KovQSqRBxL2WmPw`
automationtestuser_multipleindicators@fingertipsdevapp.onmicrosoft.com ("normal" user with access to indicator 41101 amongst others): `ZONknEk-P2EKYeURsgTjnwBYU_kkDwEPaVDZ-mnxWgo`

I then created a batch using `automationtestuser_admin` and deleted it with `automationtestuser_multipleindicators` and verified that the batch showed the correct values for the "created by" and "deleted by" user fields:
Screenshot 2025-07-24 at 09 04 21


api/DHSC.FingertipsNext.Api.IntegrationTests/DataManagement/FileUploadIntegrationTests.cs
```diff
@@ -76,11 +76,18 @@ public async Task AuthorisedRequestToDataManagementEndpointShouldUploadAFile(par
...
 
         var model = await response.Content.ReadFromJsonAsync<Batch>();
-        model.IndicatorId.ShouldBe(IndicatorId);
-        model.Status.ShouldBe(BatchStatus.Received);
-        model.OriginalFileName.ShouldBe(IntegrationTestFileName);
-        model.UserId.ShouldBe(Guid.Empty.ToString());
-        model.PublishedAt.ShouldBe(_publishedAt);
+        model.ShouldBe(new Batch
+        {
+            BatchId = $"{IndicatorId}_{MockTime:yyyy-MM-ddTHH:mm:ss.fff}",
```
williamwillis11, line 81: just wondering if 

yyyy-MM-ddTHH:mm:ss.fff

should go in as a sharable constant/var/etc somewhere so that it can be managed in one place? this time format string is in a few places

api/DHSC.FingertipsNext.Api.IntegrationTests/DataManagement/DataManagementWebApplicationFactory.cs
```diff
@@ -15,6 +15,7 @@ public class DataManagementWebApplicationFactory<T> : WebApplicationFactory<T> w
 {
     private const string JwtStubIssuer = "TestIssuer";
     private const string JwtStubAudience = "TestAudience";
+    public string SubClaim => "test-user";
```
Mr-Ev-H, line 18: Why isn't this also a const?

api/DHSC.FingertipsNext.Api.IntegrationTests/DataManagement/FileUploadIntegrationTests.cs
gareth-allan-bjss: I've extracted as many uses as I think make sense.

ehsh1 (approved):

williamwillis11 (approved):

PR #794, opened 2025-07-24
add logging to generated clients

Joe-Hendley:
add logging to generated clients to investigate auth failures


Joe-Hendley:
RIP `[${uuid}] JH...` logs, you served us well

PR #806, opened 2025-07-29
go home after signout

Joe-Hendley:
trying to preserve full state (ie through querystring) doesn't seem to work for logout so we're now just sending the user straight back to the homepage

### DIFF
--- a/frontend/fingertips-frontend/app/auth/signout/route.test.ts
+++ b/frontend/fingertips-frontend/app/auth/signout/route.test.ts
@@ -8,7 +8,10 @@ import { mockRedirect } from '@/mock/utils/mockNextNavigation';
 import { signOutHandler } from '@/lib/auth/handlers';
 import { GET } from '@/app/auth/signout/route';
 import { NextRequest } from 'next/server';
-import { buildLogoutURLWithRedirect } from '@/lib/auth/providers/fingertipsAuthProvider';
+import {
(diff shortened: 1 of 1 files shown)


frontend/fingertips-frontend/lib/auth/providers/fingertipsAuthProvider.test.ts
```diff
@@ -82,4 +84,18 @@ describe('get auth logout endpoint', () => {
...
+    vi.stubEnv('AUTH_CLIENT_ID', 'id');
+    vi.stubEnv('AUTH_CLIENT_SECRET', 'secret');
+    vi.stubEnv('AUTH_ISSUER', 'issuer');
+    vi.stubEnv('AUTH_LOGOUT', baseLogoutURL);
+
+    const redirect = 'https://some-url.place.app.cloudprovider.com';
+
+    const expectedURLString = `${baseLogoutURL}?${FTA_SIGNOUT_REDIRECT_PARAM}=${redirect}`;
+
+    expect(buildLogoutURLWithRedirect(redirect)).toEqual(expectedURLString);
```
williamwillis11, line 99: given the name of the test i was kind of expecting a negative assertion  .not.contain(redirect) 

might be me missunderstanding the test

gareth-allan-bjss (approved):

PR #807, opened 2025-07-29
Bugfix/dhscft tech sign out callback url

Joe-Hendley:
adds fingertips_frontend_url env var in to enable signout callback in deployed env


frontend/fingertips-frontend/lib/urlHelper.test.ts
```diff
@@ -0,0 +1,18 @@
+import { getFingertipsFrontendURL } from '@/lib/urlHelper';
+
+describe('get fingertips frontend url', () => {
+  it('should return FINGERTIPS_FRONTEND_URL if set', () => {
```
williamwillis11, line 4: is set

williamwillis11 (approved):

PR #812, opened 2025-07-30
DHSCFT-1078: heatmap reportingPeriods

alibombaye:
# Description

Jira ticket: [DHSCFT-1078](https://bjss-enterprise.atlassian.net/browse/DHSCFT-1078)

Using the reporting period for remaining charts. (Line Chart was previously done).

Uses a function called isSmallestReportingPeriod to determine if the label to show should render as normal. This is when the reportingPeriod is equal to frequency and represents a single range. When this is not the case it will render the labels and titles to indicate the range it represents. As shown in the screenshots below

Heatmap
Screenshot 2025-07-31 at 15 37 52

Basic Table
Screenshot 2025-07-31 at 15 38 28

Thematic Map
Screenshot 2025-07-31 at 15 38 49

Spine Chart
Screenshot 2025-07-31 at 15 41 25

Compare Area Table
Screenshot 2025-07-31 at 15 44 11

## Validation
Test manually

### DIFF
--- a/frontend/fingertips-frontend/components/charts/ThematicMap/ThematicMap.test.tsx
+++ b/frontend/fingertips-frontend/components/charts/ThematicMap/ThematicMap.test.tsx
@@ -60,6 +60,7 @@ const testRender = (name?: string) => {
         periodType={PeriodType.Calendar}
         frequency={Frequency.Annually}
         latestDataPeriod={mockDatePeriod}
+        isSmallestReportingPeriod={true}
       />
     </QueryClientProvider>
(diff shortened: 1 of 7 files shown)


alibombaye:
Regarding the sonar comments about the parameter size exceeding the maximum size of 7. This has been acknowledged and as a team we have agreed that the approach to take here is to revisit these functions at a later date. Either break it down into smaller functions or look at how we can logically group some of the params into an object and therefore reduce the size of the params passed.

